PR #5, opened 2015-05-14
Filter draft published

xybrnet:
@longhotsummer

### DIFF
--- a/indigo_app/static/javascript/indigo/views/library.js
+++ b/indigo_app/static/javascript/indigo/views/library.js
@@ -110,18 +110,6 @@
       this.trigger('change');
     },
 
-    filterBySearch: function(e) {
-      var needle = this.$el.find('.filter-search').val().trim();
-      if (needle != this.filters.search) {
-        this.filters.search = needle;
-        this.trigger('change');
-      }
-    },
-
-    resetSearch: function(e) {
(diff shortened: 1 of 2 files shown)


indigo_app/templates/library.html
```diff
@@ -127,6 +127,26 @@ <h3 class="panel-title">Tags</h3>
...
+
+<div class="panel panel-default">
+  <div class="panel-heading">
+    <h3 class="panel-title">Status</h3>
+  </div>
+
+  <div class="panel-body">
+    <div class="btn-group" data-toggle="buttons">
+      <label class="btn btn-default active">
+        <input type="radio" name="options" autocomplete="off" class="filter-status" data-status="all" checked>All
```
longhotsummer: You can store the "all" in the `value` attribute, which is the "right" way to do it for HTML forms. Then you don't need to use `data("status")` in the code, and can just use `val()` instead.

indigo_app/static/javascript/indigo/views/library.js
```diff
@@ -108,6 +110,29 @@
       this.trigger('change');
     },
 
+    filterByStatus: function(e) {
+      var $link = $(e.currentTarget);
```
longhotsummer: Perhaps rename to `$input`? It's not a link after all.

longhotsummer:
:+1:

PR #100, opened 2018-06-14
Amended setup instructions to include bit about postgresql authentica…

nickyspag:
Something that stumped me as a postgresql newb the first time I tried to set up Indigo


longhotsummer:
Thanks @nickyspag !

PR #107, opened 2018-07-10
add descriptor for Parent work

goose-life:
please also review the content of the description!


indigo_app/templates/work/edit.html
```diff
@@ -73,6 +73,7 @@ <h5 class="card-header">Work details</h5>
         <div class="form-group col-md-6">
           <button class="btn btn-outline-primary choose-parent pull-right">Choose parent work</button>
           <label>Parent work</label>
+          <p class="form-text text-muted"><br>For example where a notice puts regulations in terms of a primary Act into place</p>
```
longhotsummer: The `<br>` seems suspicious. Does it look weird if it's directly below the label?

How about `For example, a government notice or regulation's parent work is the primary act.`

PR #126, opened 2018-07-25
add 'woordomskrywing/s' to section heads to look for

goose-life:


indigo_za/terms.py
```diff
@@ -40,4 +40,4 @@ class TermsFinderAFR(TermsFinderENG):
     # country, language, locality
     locale = (None, 'afr', None)
 
-    heading_re = re.compile(r'woordbepaling', re.IGNORECASE)
+    heading_re = re.compile(r'woordbepaling|woordomskrywing|woordomskrywings', re.IGNORECASE)
```
longhotsummer: Looks good. This doesn't match an entire word, so you could just include the singular and it'll also match the plural, but this is fine.

goose-life: good to know thanks! it didn't seem to work on the plural but admittedly I was in a hurry when trying it out

PR #128, opened 2018-07-25
[DO NOT MERGE] Batch creation of new works, from spreadsheet

longhotsummer:


indigo_app/views.py
```diff
@@ -261,6 +269,142 @@ def post(self, request, frbr_uri, version_id):
...
+                frbr_uri = get_frbr_uri(row)
+
+                # TODO: fix get_uri to return uri as false if validation error?
+
+                if frbr_uri:
+
+                    try:
+                        work = Work.objects.get(frbr_uri=frbr_uri)
+                        info['work'] = work
+                        info['status'] = 2
```
longhotsummer: Magic numbers like this can be confusing because it's not immediately apparent what they mean. In some languages, one could define a constant like `EXISTS = 2` and then use `EXISTS` rather than 2, which makes it more legible. In python (and in this case), you could also use a string, such as `info['status'] = 'exists'`.

indigo_app/views.py
```diff
@@ -261,6 +269,142 @@ def post(self, request, frbr_uri, version_id):
...
+                    except Work.DoesNotExist:
+
+                        work = Work()
+
+                        work.frbr_uri = frbr_uri
+                        work.title = row[8]
+                        work.country = row[9]
+                        work.publication_name = row[10]
+                        work.publication_number = row[11]
+                        work.publication_date = row[12]
```
longhotsummer: These will need to be parsed from strings into dates. Python has some support for this. Let's see what format we get from google sheets, first.

openbylaws metadata batch-310fffa29429.json
```diff
@@ -0,0 +1,12 @@
+{
```
longhotsummer: This file can be removed.

requirements.txt
```diff
@@ -30,6 +30,7 @@ futures==3.0.3
 gevent==1.2.2
 git+https://github.com/aerkalov/ebooklib.git@a87ef643690a89d8a72389c65073913bde58d1df#egg=EbookLib-master
 greenlet==0.4.13
+gspread==3.0.1
```
longhotsummer: Can be nuked.

requirements.txt
```diff
@@ -40,13 +41,16 @@ mammoth==1.4.4
 mock==1.3.0
 newrelic==2.40.0.34
 nose==1.3.7
+oauth2client==4.1.2
```
longhotsummer: Can be nuked.

requirements.txt
```diff
@@ -40,13 +41,16 @@ mammoth==1.4.4
...
 pathlib==1.0.1
 psycopg2==2.7.3.2
+PyOpenSSL==18.0.0
```
longhotsummer: Can be nuked.

longhotsummer:
Fixes #142

PR #129, opened 2018-07-25
Process form for batch works

longhotsummer:
Here's some logic to show and process a form for capturing the spreadsheet URL. Django provides a fair amount of support for processing forms, such as validating entry.

### DIFF
--- a/indigo_app/views.py
+++ b/indigo_app/views.py
@@ -1,5 +1,3 @@
-
-
 import json
 
 from django.contrib.auth.mixins import PermissionRequiredMixin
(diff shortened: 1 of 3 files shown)


indigo_app/views.py
```diff
@@ -269,17 +268,15 @@ def post(self, request, frbr_uri, version_id):
...
-class BatchAddWorkView(AbstractAuthedIndigoView, TemplateView):
+class BatchAddWorkView(AbstractAuthedIndigoView, FormView):
     template_name = 'work/new_batch.html'
     # permissions
     permission_required = ('indigo_api.add_work',)
+    form_class = BatchCreateWorkForm
 
-    def get_context_data(self, **kwargs):
-        context = super(BatchAddWorkView, self).get_context_data(**kwargs)
+    def form_valid(self, form):
```
longhotsummer: We're only going to work with the form if the basic validation in place succeeds.

PR #163, opened 2018-09-04
recognise Constitution as an external reference

### LINKED ISSUE #114: Recognise constitution as an external reference

longhotsummer:


indigo_za/refs.py
```diff
@@ -10,16 +10,23 @@ class RefsFinderENG(BaseRefsFinder):
...
+        Constitution [of [the Republic of] South Africa] [Act][,] [1996]
+
     """
 
     # country, language, locality
     locale = (None, 'eng', None)
 
-    act_re = re.compile(r'Act,?\s+(no\.?\s*)?(\d+)+\s+of\s+(\d{4})', re.I)
-    candidate_xpath = ".//text()[contains(., 'Act') and not(ancestor::a:ref)]"
+    act_re = re.compile(r'Act,?\s+([nN]o\.?\s*)?(\d+)+\s+of\s+(\d{4})|\bConstitution\b(\s+of(\s+the\s+Republic\s+of)?\s+South\s+Africa)?((\s+Act)?,?\s+1996)?')
```
longhotsummer: Is the second `\b` in  `\bConstitution\b` necessary?
Maybe we should use `\b` for `r'\bAct`, too.

indigo_za/refs.py
```diff
@@ -28,13 +35,19 @@ class RefsFinderAFR(BaseRefsFinder):
...
+        Grondwet [van [die Republiek van] Suid-Afrika] [Wet][,] [1996]
+
     """
 
     # country, language, locality
     locale = (None, 'afr', None)
 
-    act_re = re.compile(r'Wet,?\s+(no\.?\s*)?(\d+)+\s+van\s+(\d{4})', re.I)
-    candidate_xpath = ".//text()[contains(., 'Wet') and not(ancestor::a:ref)]"
+    act_re = re.compile(r'Wet,?\s+(no\.?\s*)?(\d+)+\s+van\s+(\d{4})|\bGrondwet\b(\s+van(\s+die\s+Republiek\s+van)?\s+Suid-Afrika)?((\s+Wet)?,?\s+1996)?')
```
longhotsummer: Should `no` also be `[nN]o`?

indigo_za/refs.py
goose-life: I think for Act adding a `\b` at the start would prevent it from picking up 'fAct 20 of 2015', but what are the odds?

longhotsummer: I think the first is necessary (and Act should have it too, that's an oversight). I understand the second one now -- the alternative is to ensure it has whitespace afterwards, but this is better.

indigo_za/refs.py
goose-life: it absolutely should thank you!

Issue #173, opened 2018-09-07
If a document's expression date isn't one of the formal work amendment dates, it's never shown in the points in time page


longhotsummer:
If the doc's date doesn't fit into the recorded timeline it exists but doesn't appear on the timeline, effectively disappearing from the Work's history

solutions I can think of off the top of my head:
- force-create a PiT for each doc's date if one doesn't exist yet – but if initial pub date is fixed later on we'd rather have the doc's date move with it
- when user changes a date give them the option to delete/mark for deletion the docs at that date, or change their date to match the new PiT
- maybe to start with, just warn a user when changing a recorded date that any documents saved at that date will have to have their dates manually changed to match if they want to be retained?

(via @goose-life)

longhotsummer:
Two major cases:

* when the work's initial publication date is changed
* when an amendment's date changes

longhotsummer:
For updating the initial expressions: https://github.com/OpenUpSA/indigo/blob/master/indigo_api/serializers.py#L657

longhotsummer:
Fixed by #191.

PR #191, opened 2018-09-26
Orphan docs

goose-life:
deals with the two major cases in #173 

there is still the question of existing orphan docs, but those will be dealt with individually?

### DIFF
--- a/indigo_api/serializers.py
+++ b/indigo_api/serializers.py
@@ -4,7 +4,6 @@ from collections import OrderedDict
 from lxml.etree import LxmlError
 from itertools import groupby
 
-from django.db.models import Manager
 from django.contrib.auth.models import User
 from rest_framework import serializers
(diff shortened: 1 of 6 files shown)


indigo_api/serializers.py
```diff
@@ -653,11 +653,26 @@ def update(self, work, validated_data):
...
         # save as a revision
         with reversion.revisions.create_revision():
             reversion.revisions.set_user(user)
             work = super(WorkSerializer, self).update(work, validated_data)
 
+        ## get new date
+        new_date = work.publication_date
+
+        ## change and save docs once work has successfully updated
+        for doc in docs:
```
longhotsummer: I think we should only do this if the date doesn't match -- saving a document isn't "cheap".

longhotsummer:
Fixes #173 .

PR #198, opened 2018-10-02
Uname

goose-life:


indigo_social/views.py
```diff
@@ -24,21 +24,23 @@ class ContributorsView(ListView):
 
 class UserProfileView(DetailView):
     model = UserProfile
```
longhotsummer: I think this view's model should be the `User` object, not `UserProfile`. After all, the view encompasses the whole user (eg. first name, username, etc.).

That would mean you don't need to special-case `get_object`.

In the template, the profile is available as `user.userprofile`.

indigo_social/views.py
```diff
@@ -96,6 +99,10 @@ def form_invalid(self, form):
         self.form = form
         return redirect(self.get_success_url())
 
+    def get_object(self, queryset=None):
+        self.kwargs['pk'] = UserProfile.objects.get(user=User.objects.get(username=self.kwargs['username'])).pk
```
longhotsummer: You can simply return the userprofile object here and not call super. Otherwise, you're going to the db to get the user, then the userprofile, then getting the pk, then Django goes back to the db to get the userprofile object again.

indigo_social/templates/indigo_social/user_profile.html
```diff
@@ -103,7 +103,7 @@ <h5 class="mt-0 mb-0"><a href="{% url 'indigo_social:badge_detail' slug=badge.sl
                 {% endfor %}
               </select>
 
-              <button tyep="submit" class="btn btn-primary ml-3" data-confirm="Are you sure you want to award this badget to this user?">Award badge</button>
+              <button type="submit" class="btn btn-primary ml-3" data-confirm="Are you sure you want to award this badge to this user?">Award badge</button>
```
longhotsummer: LOL.

indigo_social/views.py
goose-life: for this I just also changed the view to use the User model, so if I've understood correctly the above doesn't apply?

indigo_social/views.py
goose-life: I was lazy to change the template but I see the very good reason for doing that ( :

PR #201, opened 2018-10-04
Pubdate

goose-life:
batch: raises a ValueError in the form, work is not saved
individual:
- create: disables Create button until model required fields plus publication_date have been provided
- edit: keeps Save button as only disabled until model required fields have been provided


indigo_app/static/javascript/indigo/views/work.js
```diff
@@ -149,11 +149,12 @@
     },
 
     canSave: function() {
-      if (this.model.isNew()) {
-        this.$('.btn.save').attr('disabled', !this.dirty || !this.model.isValid() || !this.model.get('publication_date'));
-      } else {
-        this.$('.btn.save').attr('disabled', !this.dirty || !this.model.isValid());
-      }
+      this.$('.btn.save').attr('disabled', !this.dirty || !this.model.isValid());
```
longhotsummer: The form submission essentially handles this now. Which is better because rather than disabling the button, it provides an error message to the user so they can see what's wrong.

PR #206, opened 2018-10-16
Indigo social

goose-life:
8e4fe6e is the most important commit here: I noticed that when viewing a profile other than the one I was signed into 'my' name in the navbar would change to their name

### DIFF
--- a/indigo_social/badges.py
+++ b/indigo_social/badges.py
@@ -140,7 +140,7 @@ class ContributorBadge(PermissionBadge):
     slug = 'contributor'
     name = 'Contributor'
     description = 'Can view work details'
-    permissions = ('indigo_api.view_work',)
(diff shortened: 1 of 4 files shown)


indigo_social/badges.py
```diff
@@ -140,7 +140,7 @@ class ContributorBadge(PermissionBadge):
     slug = 'contributor'
     name = 'Contributor'
     description = 'Can view work details'
-    permissions = ('indigo_api.view_work',)
+    permissions = ('indigo_api.view_work', 'indigo_workflow.view_bulklistimportworksprocess')
```
longhotsummer: Eish, this is starting to blur the lines between indigo_social and indigo_workflow.

goose-life: would it be better to change the permissions for seeing the tasks?

Issue #210, opened 2018-10-23
In indigo_social the username field updates unnecessarily


goose-life:
Eg if the surname is changed the username changes because the existing username already exists – this should only happen if the username field has been edited

longhotsummer:
This has been fixed.

PR #214, opened 2018-10-25
Username

goose-life:


indigo_social/forms.py
```diff
@@ -44,7 +44,8 @@ def save(self, commit=True):
         super(UserProfileForm, self).save()
         self.instance.user.first_name = self.cleaned_data['first_name']
         self.instance.user.last_name = self.cleaned_data['last_name']
-        self.instance.user.username = self.cleaned_data['username']
+        if 'username' in self.changed_data:
```
longhotsummer: What if we changed `clean_username` such that it creates a unique username only if the username isn't specified at all? If the given username isn't unique, then it simply shows an error, rather than trying to create one.

In most cases, users will have a username generated automatically when they sign up.

longhotsummer:
Fixes #210

indigo_social/forms.py
```diff
@@ -37,8 +43,10 @@ def clean_twitter_username(self):
             return twitter_username
 
     def clean_username(self):
-        if self.cleaned_data['username']:
-            return generate_unique_username([self.cleaned_data['username']])
+        username = self.cleaned_data['username']
+        if username in [u.username for u in User.objects.all()]:
```
longhotsummer: I think this will raise an error even if it's the user's current username. How about:

```python
if User.objects.filter(username=username).exclude(pk=self.request.user.pk).exists():
```

I'm still suspicious that we even have to do this.

indigo_social/forms.py
```diff
@@ -37,8 +43,10 @@ def clean_twitter_username(self):
             return twitter_username
 
     def clean_username(self):
-        if self.cleaned_data['username']:
-            return generate_unique_username([self.cleaned_data['username']])
+        username = self.cleaned_data['username']
+        if User.objects.filter(username=username).exclude(pk=self.instance.user.pk).exists():
+            raise forms.ValidationError("already taken")
```
longhotsummer: "This username is already taken."

indigo_social/forms.py
```diff
@@ -1,17 +1,23 @@
...
+from indigo_api.models import Country, User
 from indigo_social.models import UserProfile
 from indigo_social.badges import badges
 
 
 class UserProfileForm(forms.ModelForm):
 
+    validate_username = RegexValidator(
+        _lazy_re_compile(r'^[-a-z0-9_]+\Z'),
+        "no spaces, punctuation or capital letters",
```
longhotsummer: "You username cannot include spaces, punctuation or capital letters."

indigo_social/forms.py
```diff
@@ -1,17 +1,23 @@
 # -*- coding: utf-8 -*-
 from django import forms
-from allauth.utils import generate_unique_username
+from django.core.validators import validate_slug, _lazy_re_compile, RegexValidator
```
longhotsummer: You can remove the validate_slug import.

Issue #223, opened 2018-11-06
Don't match Constitution on non-za locales


longhotsummer:
Perhaps

1. make the current refs finder ZA-specific
2. make a copy of the English finder and put it under indigo/analysis/refs/global.py that isn't country specific, and just handles 'Act'.

PR #225, opened 2018-11-06
don't match Constitution on non-za locales

goose-life:
not sure how to test for non-SA locales but PRing in case something jumps out in the meantime


longhotsummer:
@goose-life I've made a tweak to work around the python import issues (turns out you can't call a file `global.py`). See if it works now?

longhotsummer:
Fixes #223

PR #230, opened 2018-11-08
Refs

goose-life:
For interest's sake, this worked perfectly: if I had indigo_za installed, it didn't apply the SA Constitution to a work from Angola, but did for one from SA; and if I didn't have indigo_za installed, it defaulted to just finding Acts for both. so it is possible. I do think the global refs finder should have its own name though, so that if I have indigo_za installed I can still find refs for countried outside of SA?


longhotsummer:
Cool. Fixed by #225.

Issue #185, opened 2018-09-15
Use list view rather than table for library


longhotsummer:
This would allow us to show more information without being limited to rows and columns.

https://www.patternfly.org/pattern-library/content-views/list-view/

* Expandable rows, shows documents (as table?)
* Would allow for subtype, parent work 
* Put sorting into toolbar.
* Can have multi row toolbar.

longhotsummer:
Fixed by #235

PR #235, opened 2018-11-09
List view in library

goose-life:
css doesn't work yet and we probably want to think some more about what goes into the description part – for now it's just a list of points in time if they exist, and the layout of the card doesn't use the space very well imo


longhotsummer:
Fixes #185

PR #245, opened 2018-11-14
catch refs to Acts with year included before number

goose-life:
but whole regex links which doesn't look great
would be cool if the link could start at match.group(3) instead of match.start()

### DIFF
--- a/indigo/analysis/refs/base.py
+++ b/indigo/analysis/refs/base.py
@@ -13,8 +13,7 @@ class BaseRefsFinder(LocaleBasedMatcher):
 
     act_re = None
     """ This must be defined by a subclass. It should be a compiled regular
-    expression. The first numbered group in the regex is the text that will be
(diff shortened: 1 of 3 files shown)


longhotsummer:
This winds up producing a ref like this, which I think is weird:

`<p>Something to do with Income Tax <ref href="/za/act/1962/58">Act, 1962 (No 58 of 1962</ref>).</p>`

Wouldn't this be better:

`<p>Something to do with Income Tax Act, 1962 (<ref href="/za/act/1962/58">No 58 of 1962</ref>).</p>`

goose-life:
agreed; I was stuck on getting the ref text to be something other than the regex match (my first attempt just deleted the text I wasn't linking). as mentioned above I was trying to get the link to start at match.group(3) instead of match.start(). 
I agree that the linking needs to be resolved before the regex change can be merged – just not sure if it should happen now?

Issue #72, opened 2018-04-30
As an editor I want to be able to easily see the source document for a work


longhotsummer:
Cohesive mechanism for linking work and document attachments.

longhotsummer:
This describes a mechanism to attach a single original publication document to a work, usually a PDF file.

# User needs
Editors need to be able to refer back to the original publication when reviewing work and document details such as titles, publication information and commencement dates. 

Public consumer need to be able to refer back to original gazettes when reading legislation. In particular, API consumers should be able to pull the original gazette directly from Indigo to enable access on their platforms.

Publication documents apply equally to primary works and amending works which may not have any expressions in Indigo (ie. stubs).

# How attachments work currently
Attachments are currently only supported for documents, not works. They are freeform and unlimited. This makes it confusing for API consumers to determine what is an original attachment for the initial expression of a work, versus for subsequent expressions. Often the attachment to a document is the source for the import, which could be a Word file, not the original gazette. Additionally, document attachments aren’t support for amending works that don’t have a document. Attachments may be duplicated between different expressions of a work.

# Suggested changes

Editors should be able to attach a single attachment to a work that is the best available official source for that work -- usually the gazette. This may be the gazette in full, or only the applicable pages. It is strongly modelled as the publication attachment (ie. it is not one of many attachments).

## Editor Changes

1. The work overview page will include a link to the publication attachment nearby to the publication document and date information.
2. The edit work page will allow a user to drag-and-drop (or upload) a file as the publication attachment. The file can be any supported attachment file. ie. PDF is not enforced.
3. When editing or creating a work and Indigo is able to find a gazette based on publication information (eg. via opengazettes.org.za), the page will offer to fetch and attach the gazette as the original publication.

## Published Document API Changes

The work API will include a `publication_attachment` alongside the `publication_date` and other publication fields. This will include information similar to other attachments, including a download URL.

```json
{
  "url": "host/api/za/act/1980/1/media/publication/filename.pdf",
  "filename": "filename.pdf",
  "mime_type": "application/pdf",
  "size": 123456,
}
```

## Database and Storage Changes

Attachments will be stored in S3 using the same mechanism that is currently used for document attachments. Work attachments will be stored under the S3 path “work-attachments” in the same S3 bucket as document attachments (which are stored under the path “attachments”.)

longhotsummer:
[...]

PR #286, opened 2018-11-28
publication attachment

goose-life:
See #72 

* [x] database model
* [x] published document API
* [x] use form to save work changes
* [x] UI to handle attaching a document
* [x] link to document in work overview view
* [x] handle new work using new form
* [x] ensure revisions, `updated_by` and `created_by` when saving/creating a work
* [x] remove writeable work API
* [x] enforce filename
* [x] handle uploading a file (either new, or to replace an existing one)
* [x] handle removing the attachment
* [ ] UI to auto-link to gazette


longhotsummer:
Ref #72

indigo_app/forms.py
```diff
@@ -21,29 +20,31 @@ class Meta:
...
     def save_publication_document(self):
         pub_doc_file = self.cleaned_data['publication_document_file']
         if pub_doc_file:
-
-            pub_doc = PublicationDocument(work=self.instance, file=pub_doc_file,
-                                          size=pub_doc_file.size, filename=pub_doc_file.name,
-                                          mime_type="application/pdf")
-            try:
+            # if one already exists for this work, update it
+            if hasattr(self.instance, 'publication_document'):
```
longhotsummer: You want to check `if self.instance.publication_document`. This is because `hasattr(x, 'foo')` will be True even if `x.foo` is None.

longhotsummer: `hasattr(x, 'foo')` is usually used before calling `x.foo` if you're unsure if `x` actually has an attribute `foo`. If it doesn't, then `x.foo` would raise an exception

indigo_app/forms.py
```diff
@@ -21,29 +20,31 @@ class Meta:
...
+                PublicationDocument.objects.filter(work=self.instance).update(
```
longhotsummer: I don't think this will do the right thing on the actual `file` object, because this goes directly to the database. Instead, you'll need to set the attributes individually.

```
self.instance.publication_document.file = pub_doc_file
self.instance.publication_document.size = pub_doc_file.size,

...
```

PR #304, opened 2018-12-07
library view changes

goose-life:


longhotsummer:
As discussed, we're going to stay with n_docs because it corresponds with the number of items that the user will see when they expand. It's also useful to see when a work has 0 documents linked to it.

PR #307, opened 2018-12-10
link usernames

goose-life:
not sure if there's a better way? in other places we use eg `{% user_profile task.created_by_user %}` but I'm not sure whether this is possible inside handlebars?

### DIFF
--- a/indigo_app/static/javascript/indigo/views/library.js
+++ b/indigo_app/static/javascript/indigo/views/library.js
@@ -426,7 +426,7 @@
 
       this.$el.find('[title]').tooltip({
         container: 'body',
-        placement: 'auto top'
+        placement: 'auto'
       });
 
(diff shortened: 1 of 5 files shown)


indigo_app/static/javascript/indigo/views/library.js
```diff
@@ -369,13 +369,39 @@
         work.n_docs = work_docs.length;
         work.n_docs_singular = work_docs.length == 1;
 
+        if (work.drafts_v_published.n_published) {
+          work.n_published = work.drafts_v_published.n_published
```
longhotsummer: This is simpler and works when `work.drafts_v_published.n_published` is false, null or 0.

```js
work.n_published = work.drafts_v_published.n_published || 0;
```

Issue #281, opened 2018-11-27
Need a mechanism to store notes of issues that should be raised/queried with the publishing authority.


longhotsummer:
These could be about wording in a document, or about a work in general (eg. an unclear commencement).

This should cooperate with annotations well.

longhotsummer:
Task labels.

PR #311, opened 2018-12-13
Task labels

longhotsummer:
Models and UI for applying labels to tasks.


longhotsummer:
Partially allows #281, even better when we can create tasks from annotations.

PR #312, opened 2018-12-13
Tasks for a work WIP

goose-life:


goose-life:
so far there's a button on the work overview that goes to a filtered task view

indigo_app/views/tasks.py
```diff
@@ -32,7 +33,25 @@ class TaskListView(TaskViewBase, ListView):
     paginate_orphans = 4
 
     def get_queryset(self):
-        return Task.objects.filter(country=self.country, locality=self.locality).order_by('-created_at')
+        try:
+            frbr_uri = self.request.GET['frbr_uri']
```
longhotsummer: Much like with a `dict`, you can use `self.request.GET.get('frbr_uri')` which returns None if the key doesn't exist in the dict, rather than raise an exception.

indigo_app/views/tasks.py
```diff
@@ -32,7 +33,25 @@ class TaskListView(TaskViewBase, ListView):
...
+        except MultiValueDictKeyError:
+            frbr_uri = None
+        if frbr_uri:
+            return Task.objects.filter(work__frbr_uri=frbr_uri).order_by('-created_at')
```
longhotsummer: We should first filter by country and locality. Otherwise, I can go to `/places/za-cpt/tasks?frbr_uri=/na/act/2014/1` and see tasks for a Namibian work on the Cape Town task list.

So filtering by the `frbr_uri` should happen *in addition* to the place filtering.

indigo_app/views/tasks.py
```diff
@@ -32,7 +33,19 @@ class TaskListView(TaskViewBase, ListView):
...
+        if frbr_uri:
+            tasks = tasks.filter(work__frbr_uri=frbr_uri).order_by('-created_at')
+        return tasks
+
+    def get_context_data(self, **kwargs):
+        context = super(TaskListView, self).get_context_data(**kwargs)
+        frbr_uri = self.request.GET.get('frbr_uri')
+        if frbr_uri:
+            context['frbr_uri'] = frbr_uri
+            context['title'] = Work.objects.get(frbr_uri=frbr_uri).title
```
longhotsummer: Remember, a work with the given frbr_uri may not exist.

indigo_app/templates/indigo_api/task_list.html
```diff
@@ -9,50 +9,14 @@
...
-                  by {% user_profile task.created_by_user %}
-                  {% endif %}
-                </small>
-              </div>
-            </div>
-          </div>
-        {% endfor %}
-      </div>
-    </div>
+  {% if frbr_uri %}
```
longhotsummer: I don't think you need to do this -- the list of tasks is already passed into the template as context. eg. for the task listing page, it'll be paginated. You can rely on the `tasks` already in the context.

We did this for the old workflows because we weren't able to override the view logic to add tasks to the context.

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -168,7 +168,6 @@ <h4>Active tasks</h4>
     </div>
   </div>
 
-  {% task_list place=place work=work active=True as active_tasks %}
   {% include 'indigo_api/_task_list.html' with tasks=active_tasks %}
```
longhotsummer: You may need to wrap this in `<div class="card">...</div>`? I removed the surrounding card from _task_list.html because I needed to put the filter form in the card header.

goose-life: done, thanks!

Issue #300, opened 2018-12-07
I want to be able to see the history of task state changes


longhotsummer:

goose-life:
we can keep track of which user _last_ put a task into a state by adding e.g. `last_cancelled_by_user` and `last_cancelled_at` to the model and saving these details whenever the state changes.
but that won't give us a full history.
could we also have something like `previously_cancelled_at {<time>: by <user>}`, which we add to each time we record a new cancellation event on a task?

goose-life:
resolved by #322

PR #322, opened 2019-01-28
Activity streams

goose-life:
start using activity streams for keeping track of and viewing task state changes – see #300


indigo_app/views/tasks.py
```diff
@@ -192,32 +185,21 @@ class TaskChangeStateView(TaskViewBase, View, SingleObjectMixin):
...
+            'submit': 'submitted',
+            'cancel': 'cancelled',
+            'reopen': 'reopened',
+            'unsubmit': 'unsubmitted',
+            'close': 'closed',
+        }
+
+        for change_made, verb in potential_changes.items():
+            if self.change == change_made:
+                if not has_transition_perm(task.__getattribute__(change_made), self):
```
longhotsummer: Nice :)

`getattr` is preferred to `task.__getattribute__`, calling underscore methods is generally discouraged. I'd also store the result of that in a variable since you call it twice.

indigo_app/views/tasks.py
```diff
@@ -192,32 +185,21 @@ class TaskChangeStateView(TaskViewBase, View, SingleObjectMixin):
...
+                    raise PermissionDenied
+                task.__getattribute__(change_made)(user)
+                action.send(user, verb=verb, action_object=task)
```
longhotsummer: I think it'd be better to put this line into the transition method on the object. The idea being that the object itself is then responsible for firing all the things that need to happen when it changes state, rather than the view (which simply says "change state").

indigo_api/models.py
```diff
@@ -945,21 +945,6 @@ class Meta:
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
-    last_submitted_by_user = models.ForeignKey(User, related_name='+', null=True, on_delete=models.SET_NULL)
```
longhotsummer: Nice. We may need these details later for metrics, but we can add them back if necessary.

indigo_api/models.py
```diff
@@ -227,6 +229,10 @@ def repeal(self):
                 self._repeal = RepealEvent(self.repealed_date, self.repealed_by.title, self.repealed_by.frbr_uri)
         return self._repeal
 
+    @property
+    def place(self):
+        return self.locality if self.locality else self.country
```
longhotsummer: Nice. You can also do `return self.locality or self.country`, but no need to change it.

indigo_api/models.py
```diff
@@ -342,6 +349,24 @@ def __unicode__(self):
         return '%s (%s)' % (self.frbr_uri, self.title)
 
 
+@receiver(signals.post_save, sender=Work)
+def post_save_work(sender, instance, **kwargs):
+    """ Cascade (soft) deletes to linked documents
```
longhotsummer: Outdated comment.

indigo_api/models.py
```diff
@@ -945,23 +974,12 @@ class Meta:
...
-
     labels = models.ManyToManyField('TaskLabel', related_name='+')
 
+    @property
+    def place(self):
+        return self.locality if self.locality else self.country
+
     @property
     def place_code(self):
         return self.country.code + '-' + self.locality.code if self.locality else self.country.code
```
longhotsummer: This can now change to `self.place.place_code` which hides this weirdness inside the country/locality class, where it belongs.

indigo_social/templates/indigo_social/user_profile.html
```diff
@@ -108,6 +110,61 @@ <h5 class="mt-0 mb-0"><a href="{% url 'indigo_social:badge_detail' slug=badge.sl
           {% endif %}
         </div>
       </div>
+
```
longhotsummer: Note to myself to look at this.

indigo_api/models.py
goose-life: can `place_code` on Task actually just go away altogether?

longhotsummer: Yes, that makes sense to me.

PR #331, opened 2019-01-31
Make / redirect to /places/, which lists places and works.

### LINKED ISSUE #323: When I land on the homepage I want to see a list of countries to choose from
Perhaps for each country show a header image, a count of works and a count of open tasks.

longhotsummer:
Fixes #323


indigo_app/views/places.py
```diff
@@ -92,7 +103,14 @@ def get_context_data(self, **kwargs):
             document_tasks[doc_id]['n_tasks'] = sum(states.itervalues())
         context['document_tasks_json'] = json.dumps(document_tasks)
 
-        work_n_amendments = {x.id: {'n_amendments': x.amendments.count()} for x in works}
-        context['work_n_amendments_json'] = json.dumps(work_n_amendments)
+        # summarise amendments per work
```
longhotsummer: This does one query, rather than the old method's query-per-work.

goose-life: nice

PR #340, opened 2019-02-05
Merge work changes with activities.

### LINKED ISSUE #333: As a user looking at the work page, I want to see recent activity
Help the user differentiate or understand the difference between recent changes, and recent activity.

Can they be merged into one timeline? Can we link changes (diffs) to activity entries?

Probably show recent things, then click through to a full paginated listing.

longhotsummer:
Closes #333

changes cpt muni planning laws africa 2019-02-05 17-17-02

### DIFF
--- a/indigo_app/templates/indigo_api/_work_diff_card.html
+++ b/indigo_app/templates/indigo_api/_work_diff_card.html
@@ -1,45 +1,44 @@
 {% load indigo_app %}
 
-<div class="card mb-3">
+<div>
+  {% user_profile version.revision.user %} made
+  {{ version.changes|length }} change{{ version.changes|length|pluralize}} 
+  (<a href="#changes-{{ version.id }}" data-toggle="collapse">show changes</a>)
+  <span class="time-ago" data-timestamp="{{ version.revision.date_created|date:'c' }}">{{ version.revision.date_created|date:"Y-m-d H:i" }}</span>
+  {% if version.revision.comment %}
+    <span class="text-muted">· {{ version.revision.comment }}</span>
+  {% endif %}
+</div>
+
+<div class="card mt-3 collapse" id="changes-{{ version.id }}">
   <div class="card-body">
-    <a href="#changes-{{ version.id }}" data-toggle="collapse" class="float-right">{{ version.changes|length }} change{{ version.changes|length|pluralize}}</a>
-    <div>
-      <span class="time-ago" data-timestamp="{{ version.revision.date_created|date:'c' }}">{{ version.revision.date_created|date:"Y-m-d H:i" }}</span>
-      by {% user_profile version.revision.user %}
-      {% if version.revision.comment %}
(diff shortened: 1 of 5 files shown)


goose-life:
1. if I've understood right, this is just going to double up every time: 'Person updated Work X time ago'; AND '(same) Person made N changes (almost same) X time ago' – what's the benefit in also including the activity stream sentences?
2. 'Restore this version' seems to be giving the more up-to-date version as the option to restore?

I wonder if it would be better just to leave the activity stream out of `/revisions` and link to it from the place's activity tab whenever a work has been updated?

longhotsummer:
I thought it might be useful to see what *changed* when it says "Greg update this work...", which is why I tried to merge the views.

What do you think about automatically combining "updated" activity and the "X changes" items when they're, say, within 5 seconds of each other (ie. the one caused the other)?

Another option is to keep them totally separate, and then provide a link from "Greg updated" to the (probably) appropriate version change in the changes view? I do see that maybe the revision history and reverting changes functionality deserves a specific view.

goose-life:
I agree that it would be good to know what changed, but then the fact that the work was updated becomes redundant I think, so we can just display the X changes? especially in the `/revisions` view, but I don't see why it shouldn't also be in `/activities` – it's autocollapsed so won't take up any more space than it currently does. 
but maybe we should keep the 'revert' stuff out of `/activity`?

longhotsummer:
I think I'm leaning towards a combined `/revisions` and `/activity` view, and skipping "updated" activities in lieu of the changes themselves. I'm worried that separate views would be confusing -- when should the user look at one rather than the other? They're both showing historical activity. The bulk of the activities will be changes.

goose-life:
> I think I'm leaning towards a combined `/revisions` and `/activity` view, and skipping "updated" activities in favour of the changes themselves.

I think we're on the same page here.

Do you mean by 'separate views' that it would look different on the Work's Activity page vs the Place's? If yes then I agree there too, there's no reason to have it be different. It would just be weird for someone to want to revert to a previous version of a work from what's essentially the Place overview I think, so I would like to make a case for that only being included in the Work's Activity view.

longhotsummer:
> Do you mean by 'separate views' that it would look different on the Work's Activity page vs the Place's? If yes then I agree there too, there's no reason to have it be different. It would just be weird for someone to want to revert to a previous version of a work from what's essentially the Place overview I think, so I would like to make a case for that only being included in the Work's Activity view.

No, I mean that there's a work `/activity` page and a work `/revisions` page. I prefer a combined page (which we're in agreement with).

I'm not suggesting we include work revisions anywhere else in activity views (eg. the place activity page).

In think we're in agreement. I'll go ahead and make the tweaks to essentially remove the "updated" activity.

goose-life:
totally on the same page

longhotsummer:
changes example with another change indigo platform 2019-02-06 13-52-52

goose-life:
Nice!

PR #341, opened 2019-02-05
Don't use a fixed-top header

longhotsummer:
This means that the header scrolls out the way. Advantages:

* simpler to make other components sticky to the top of the page (eg. list column headers when scrolling through long lists)
* can remove the drop shadow, so cleaner
* #anchors in URLs position correctly
* more page space for long pages

tester9 laws africa 2019-02-05 17-37-42
tester9 laws africa 2019-02-05 17-37-26


longhotsummer:
I'm surprised at how simple this was. CSS flexbox is awesome.

PR #342, opened 2019-02-05
Place activities

goose-life:


goose-life:
screenshot from 2019-02-05 18-05-01

goose-life:
should the `place` come out? is it redundant given we're in that place's activities tab?

longhotsummer:
I'm okay if we include the place. It means we can use the same template everywhere. An alternative is to have a flag to the template to tell it not to render the place. Would be slightly easier to read for the user, I guess. Up to you.

indigo_app/templates/place/tabbed_layout.html
```diff
@@ -35,6 +35,7 @@
       <a class="nav-item nav-link {% if view.tab == 'tasks' %}active{% endif %}" href="{% url 'tasks' place=place.place_code %}">
         Tasks <span class="badge badge-light">{{ place.tasks.unclosed.count }}</span>
       </a>
+      <a class="nav-item nav-link {% if view.tab == 'activities' %}active{% endif %}" href="{% url 'place_activities' place=place.place_code %}">Activities</a>
```
longhotsummer: I think the title "Activity" makes more sense.

indigo_app/urls.py
```diff
@@ -18,6 +18,7 @@
 
     url(r'^places/$', places.PlaceListView.as_view(), name='places'),
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/$', places.PlaceDetailView.as_view(), name='place'),
+    url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/activities$', places.PlaceActivityView.as_view(), name='place_activities'),
```
longhotsummer: I prefer "/activity" which matches what we're doing for the user profile.

indigo_app/views/places.py
```diff
@@ -114,3 +117,31 @@ def get_context_data(self, **kwargs):
...
+
+    object_list = None
+    page_size = 20
+    js_view = ''
+
+    def get_context_data(self, **kwargs):
+        context = super(PlaceActivityView, self).get_context_data(**kwargs)
+
+        all_activities = Action.objects.all()
+        activities = [act for act in all_activities if act.data and act.data['place_code'] == self.place.code]
```
longhotsummer: We need to get the database filter working, this will rapidly become unsustainable.

indigo_app/templates/place/tabbed_layout.html
goose-life: agreed

indigo_app/views/places.py
```diff
@@ -114,3 +117,30 @@ def get_context_data(self, **kwargs):
...
+    tab = 'activity'
+
+    object_list = None
+    page_size = 20
+    js_view = ''
+
+    def get_context_data(self, **kwargs):
+        context = super(PlaceActivityView, self).get_context_data(**kwargs)
+
+        activity = Action.objects.filter(data__place_code=self.place.place_code)
```
longhotsummer: yay.

PR #346, opened 2019-02-06
Workflows

goose-life:
didn't do a lot on displaying the tasks as that'll get pulled in from #345; 
also still a bit to be done ito linking and listing workflows etc

* [x] choose tasks in edit and create views
* [x] show workflow in task listing view
* [x] show workflow in task detail view
* [x] finalise workflow listing view
* [x] finalise workflow detail view
* [x] activity view should show workflow activities correctly
* [x] calculate and show completeness information
* [x] support closing a workflow
* [x] support deleting a workflow
* [x] due date


indigo_app/views/tasks.py
```diff
@@ -202,3 +202,68 @@ def post(self, request, *args, **kwargs):
...
+
+    js_view = ''
+
+    context_object_name = 'workflow'
+    form_class = WorkflowForm
+    model = Workflow
+
+    def get_context_data(self, *args, **kwargs):
+        context = super(WorkflowCreateView, self).get_context_data(**kwargs)
+        if self.locality:
```
longhotsummer: Is it possible to combine into `tasks = self.place.tasks`?

Also, do you want to restrict to non-closed tasks? I think I added a method for that on `Task.objects`

goose-life:
workflow_detail_closed
workflow_detail
task_detail
user_activity
place_activity

indigo_app/views/places.py
```diff
@@ -144,3 +146,61 @@ def get_context_data(self, **kwargs):
...
+                if getattr(action, 'verb', None) == 'added':
+                    # if yes, was the previous action also an addition?
+                    prev = stream[i - 1]
+                    if getattr(prev, 'verb', None) == 'added':
+                        # if yes, did the two actions happen close together and was it on the same workflow?
+                        if prev.timestamp - action.timestamp < self.threshold \
+                                and action.target_object_id == prev.target_object_id:
+                            # if yes, the previous action was added to the stash and
+                            # this action should also be added to the stash
+                            added_stash.append(action)
```
longhotsummer: The logic here is fine. It's worth noting that all of these blocks end with `added_stash.append(action)`. That means you could probably remove a bunch of `else` statements and just have `added_stash.append(action)` at the end of the if block that starts on line 169.

PR #358, opened 2019-02-12
wrap up workflows

goose-life:


indigo_app/views/tasks.py
```diff
@@ -107,6 +107,10 @@ def form_valid(self, form):
         response_object = super(TaskCreateView, self).form_valid(form)
         task = self.object
         task.workflows = form.cleaned_data.get('workflows')
+        if task.workflows.all():
```
longhotsummer: Each `.all()` will go to the database. So you can simply drop the `if` and rely on the fact that if there aren't workflows, the `for` loop does nothing.

indigo_app/views/tasks.py
```diff
@@ -164,13 +168,33 @@ class TaskEditView(TaskViewBase, UpdateView):
...
-        # TODO: send action signal saying 'User added|removed Task to|from Workflow,
-        #  and don't send  generic 'updated' signal if this was the only change
+        # action signals
+        # first, was something changed other than workflows?
+        if form.changed_data:
+            action.send(self.request.user, verb='updated', action_object=task,
+                        place_code=task.place.place_code)
+        # then, was the task added to / removed from any workflows?
+        new_workflows = [wf.id for wf in task.workflows.all()]
+        if set(old_workflows) != set(new_workflows):
```
longhotsummer: You can also do set operations: `set(old_workflows) - set(new_workflows)` is all the ones that have been removed. `new - old` are the ones that are new. `old & new` is the intersection, those that haven't changed.

See https://docs.python.org/2/library/sets.html

indigo_app/views/tasks.py
```diff
@@ -164,13 +168,33 @@ class TaskEditView(TaskViewBase, UpdateView):
...
         task.updated_by_user = self.request.user
-        action.send(self.request.user, verb='updated', action_object=task,
-                    place_code=task.place.place_code)
         task.workflows = form.cleaned_data.get('workflows')
 
-        # TODO: send action signal saying 'User added|removed Task to|from Workflow,
-        #  and don't send  generic 'updated' signal if this was the only change
+        # action signals
+        # first, was something changed other than workflows?
+        if form.changed_data:
```
longhotsummer: I think `form.changed_data` would also include updated workflows?

goose-life: if only the workflows are changed `changed_data` is empty, I know because I initially looked there for changes to workflows

indigo_app/views/tasks.py
goose-life: this is amazing thank you!

indigo_app/views/tasks.py
longhotsummer: Cool.

PR #363, opened 2019-02-12
Render document's coverpage in the editor view

### LINKED ISSUE #269: Not always clear what editorial metadata to include at the start of the document, and what Indigo will put there itself.
Best solution is to asynchronously render the header in the background and show it when the whole document is rendered. That will make the preview redundant.

longhotsummer:
Fixes #269. Makes it much easier to understand what Indigo does for you, and what should/shouldn't be included at the start of a document.

This also moves AKN templates into a better namespace and allows for a pluggable coverpage template.

example 2010-12-05 2019-02-12 18-26-43


longhotsummer:
We should now be able to remove the Preview functionality, too.

PR #392, opened 2019-02-18
redirect to work_overview after successfully editing work

goose-life:


goose-life:
should fix #384

longhotsummer:
I think we should add a flash message saying work updated. We should show more of these. eg.

`messages.success(self.request, u"Work updated.")`

goose-life:
closes #384

PR #394, opened 2019-02-18
Customise publication doc download filename. Fixes #385

### LINKED ISSUE #385: as a reviewer downloading the publication documents for reference, I want them not to all have the same name
### Suggestion
Fall back to `publication-document.pdf` if no metadata is available, but if any is available use the publication number and/or date instead

(Request from MBB)

longhotsummer:
Eg. `za-act-2003-51-publication-document.pdf`


longhotsummer:
We currently force the name to `publication-document.pdf` when the file is attached. Are you suggesting we force the name at that point?

The issue there is we have to keep it up to date if the work's FRBR URI changes. This way, we do it at the last minute.

goose-life:
my only issue with not doing it from he start is that it isn't obvious to a person downloading the pdf before they download it that it'll get renamed – so if they look in their downloads folder they'll be looking for `publication-document`, and they won't realise that they can download a bunch without ending up with `publication-document (1)`, etc

longhotsummer:
@goose-life reworked this to follow your suggestion, please take a look.

Issue #367, opened 2019-02-13
as a user I want clarity on the difference between publication and commencement dates


goose-life:
### Suggestion
A richer timeline that gives the commencement date as well as the current info (initial publication, amendments)

### Considerations
- Plenty of works commence upon publication; we probably don't want two entries on the timeline for these, implying that they're two different versions when they're not.
- Most works that commence after publication aren't changed between publication and commencement; we probably only want to display the document once for these, but give the publication and commencement dates.
- Some works are amended with effect from whatever their commencement date ends up being; for these we may want to combine the commencement and amendment events as they happen on the same date and the version of that work on that date will be as amended.

### More specifically
We'll have to think about how to lay out the dates and versions so as to be more informative rather than confusing; here are some initial ideas (bear in mind the dates are on the left-hand side of the 'timeline' (`|`); everything else is in a card on the right-hand side):

#### Scenario 1: commences on publication
- 1 May 2000 
`|` Amended by XYZ 
`|` Document title • Language

- 1 Jan 2000 
`|` Initial publication `(this work commenced on publication)` 
`|` Document title • Language

#### Scenario 2: commences after publication
- 1 May 2000 
`|` Amended by XYZ 
`|` Document title • Language

- 1 Mar 2000 
`|` Commencement
1 Jan 2000 
`|` Initial publication 
`|` Document title • Language

#### Scenario 3: commences after publication and is amended from commencement
- 1 May 2000 
`|` Amended by XYZ 
`|` Document title • Language

- 1 Mar 2000 
`|` Commencement 
`|` Amended by ABC
`|` Document title • Language

- 1 Jan 2000 
`|` Initial publication 
`|` Document title • Language

PR #395, opened 2019-02-18
Timeline

goose-life:


goose-life:
see #367

indigo_app/views/works.py
```diff
@@ -56,6 +56,29 @@ def determine_place(self):
...
+        other_dates = [
+            ('assent_date', self.work.assent_date),
+            ('commencement_date', self.work.commencement_date),
+            ('repealed_date', self.work.repealed_date)
+        ]
+        # add to existing dates (e.g. if publication and commencement dates are the same)
+        for date in work_timeline:
+            for name, other_date in other_dates:
+                if date['date'] == other_date:
+                    date['{other_date}'.format(other_date=other_date)] = True
```
longhotsummer: This seems weird. What are you trying to do?

`str(other_date)` may be simpler than a format string, if that's what you want. Why do you want a date string to be True?

indigo_app/views/works.py
```diff
@@ -56,6 +56,29 @@ def determine_place(self):
...
+
+        # add other dates to timeline
+        work_timeline = self.work.points_in_time()
+        other_dates = [
+            ('assent_date', self.work.assent_date),
+            ('commencement_date', self.work.commencement_date),
+            ('repealed_date', self.work.repealed_date)
+        ]
+        # add to existing dates (e.g. if publication and commencement dates are the same)
+        for date in work_timeline:
```
longhotsummer: Perhaps call this `entry` since it's not just a date.

indigo_app/views/works.py
longhotsummer: I think you want `date[name] = True`L

indigo_app/views/works.py
```diff
@@ -56,6 +56,29 @@ def determine_place(self):
...
+            for name, other_date in other_dates:
+                if date['date'] == other_date:
+                    date['{other_date}'.format(other_date=other_date)] = True
+        # add new dates (e.g. if assent is before any of the other dates)
+        existing_dates = [date['date'] for date in work_timeline]
+        for name, other_date in other_dates:
+            if other_date and other_date not in existing_dates:
+                work_timeline.append({
+                    'date': other_date,
+                    '{name}'.format(name=name): True,
```
longhotsummer: You can simply do `name: True`

indigo_app/views/works.py
```diff
@@ -56,6 +56,49 @@ def determine_place(self):
...
+        other_dates = [
+            ('assent_date', self.work.assent_date),
+            ('commencement_date', self.work.commencement_date),
+            ('repealed_date', self.work.repealed_date)
+        ]
+        # add to existing dates (e.g. if publication and commencement dates are the same)
+        for entry in work_timeline:
+            for name, date in other_dates:
+                if entry['date'] == date:
+                    if name == 'commencement_date' and self.work.commencing_work:
```
longhotsummer: Minor style comment (totally optional):

You're effectively always setting `entry[name] = True`, so you could do that before line 71 and simplify the `if`s. Additionally, it's not going to hurt to have `event['commencing_work']` be None, so you can remove that check too.

indigo_app/views/works.py
```diff
@@ -56,6 +56,49 @@ def determine_place(self):
...
+                    elif name == 'repealed_date' and self.work.repealed_by:
+                        entry['repealed_date'] = True
+                        entry['repealed_by'] = self.work.repealed_by
+                    else:
+                        entry[name] = True
+        # add new dates (e.g. if assent is before any of the other dates)
+        existing_dates = [entry['date'] for entry in work_timeline]
+        for name, date in other_dates:
+            if date and date not in existing_dates:
+                if name == 'commencement_date' and self.work.commencing_work:
```
longhotsummer: Similar comment from above applies here (but it's stylistic). You could start by doing everything in the final `else`, and then handle the exceptions.

goose-life: this way I don't have to then find the index of the dict in the list to update, so I was being a little lazy :grimacing:

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -16,16 +16,46 @@
...
+        {% if event.assent_date %}
+          <h6>Assented to</h6>
+        {% endif %}
+
+        {% if event.initial %}
+          <h6>
+            Initial publication
+            {% if event.commencement_date %}
+              <span class="text-muted"> – Commenced on publication
+                {% if event.commencing_work %}
```
goose-life: actually, I don't think we need to pass this into the context separately – we can just use `if work.commencing_work`?

PR #396, opened 2019-02-19
include work's frbr_uri in its publication document from upload

goose-life:


longhotsummer:
Thanks, I've borrowed this.

PR #404, opened 2019-02-21
move cancellation of task out of detail view and into edit

goose-life:


indigo_app/templates/indigo_api/task_form.html
```diff
@@ -116,10 +116,20 @@
         </div>
 
         <div class="card-footer text-right">
+          {% if task.state == 'open' or task.state == 'pending_review' %}
+            <form method="POST">
+              {% csrf_token %}
+                {% if cancel_task_permission %}
```
longhotsummer: These 3 lines are too far indented

Issue #291, opened 2018-12-04
UI to help auto-link discovered gazette as a work publication


longhotsummer:
When we auto-discover a gazette, make it easy to link it as a publication. We should be able to do this for new works, even before they've been saved. See #286 for reference.

longhotsummer:
Fixed by #407.

Issue #111, opened 2018-07-13
Gazette link should show gazette if possible, rather than download, be clearer what it does


longhotsummer:

longhotsummer:
Done.

PR #407, opened 2019-02-22
Make it easy to attach Gazette Machine gazettes as original publication documents

longhotsummer:
edit remarks import tester laws africa 2019-02-22 15-53-12

This does two major things

1. Allows a work's publication document to be a URL to a trusted gazette source, rather than a directly attached PDF. This means that for trusted sources (eg.  Gazette Machine or opengazettes.org.za), we don't duplicate the PDF and store it twice in S3. Correspondingly, the user is redirected to this trusted URL when they try to download the publication document.
2. Improves the interface to help the user attach a recognise gazette easily. Instead of having to upload a PDF, it uses the existing publication search API to try to find matching gazette(s) and then shows them to the user with the option of attaching one of them as the original publication document. This means that for many cases where we have the gazette, attaching it is a one-click operation.


longhotsummer:
Closes #291.

longhotsummer:
Fixes #111.

Issue #388, opened 2019-02-18
As a reviewer I want to get a quick idea of key details of a work by hovering over a link to that work


longhotsummer:
For instance, when reviewing that amending information is correct, I don't want to have to click through to the amending work's page. Showing a popover similar to have wikipedia does it would be very helpful.

We could use JS to bind to A tags that link to a work, and use BS4's popover feature to shower a popover box on hover. We can call `.../work/popover` to generate it on the server side.

longhotsummer:
Here's the skeleton of my idea.

1. On page load, wrap a BS4 popover to links to "/works/foo/bar" (or maybe we need to use a `data-popup` to decorate just the ones we want, we'll have to see, otherwise it may decorate "edit this work" buttons.
2. When the popover loads, get the html from `/works/foo/bar/popover`

Something like this (untested):
```js
$('body').on('mouseover', 'a[href^="/works/"]:not(.btn)', function(e) {
  var el = this;
  $.ajax.get(this.href + '/popup').then(function(html) {
    // we probably want to cache html somewhere, for this work on this page
    $(el).popover({content: html}).popover('show');
  });
});
```

The popup could look like a simplified version of this:

<img width="763" alt="nature conservation amendment act 1986 laws africa 2019-02-21 14-41-42" src="https://user-images.githubusercontent.com/4178542/53169742-c4156f00-35e6-11e9-954f-795b964212e9.png">

ie, basic HTML:

* title (bold)
* frbr_uri (muted)
* Promulgation date (publication document isn't crucial)
* assent, commencement date + work
* if repealed, repeal info

AliciadW:
https://github.com/laws-africa/indigo/pull/411

PR #411, opened 2019-02-26
388 key details on hover

AliciadW:
screenshot 2019-02-26 at 10 59 50

Details of work is shown in popover on hover, user can click through to linked works from popover


longhotsummer:
I think this is a great proof of concept. To finalise:

1. @goose-life please can you add a work-based url such as `/works/.../popup` which renders the chunk of HTML to show in the popover. Don't use `{% extends ... %}` so that it doesn't inherit all the placeholder html.

2. @AliciadW can we set it up so that the content is loaded dynamically from the above endpoint on mouseover, rather than injected on the server? See eg my comment in https://github.com/laws-africa/indigo/issues/388 -- we're going to want this to popup on a number of links on the page and don't want to have to duplicate it.

3. @AliciadW the JS code can live in https://github.com/laws-africa/indigo/blob/master/indigo_app/static/javascript/indigo/utils.js

longhotsummer:
This is a bit outdated but is very similar to what we need: https://stackoverflow.com/questions/8130069/load-a-bootstrap-popover-content-with-ajax-is-this-possible

indigo_app/static/javascript/indigo/utils.js
```diff
@@ -15,3 +15,21 @@ $(function() {
 
   $('body').on('click', 'a[data-confirm], button[data-confirm]', handleConfirm);
 });
+
+// Show popover when hovering on selected links
+$('body').on('mouseenter', 'a[data-popup-url]', function() {
+  var _this = this;
+  $.get(this.href + 'popup').then(function(html) {
```
longhotsummer: This URL should be the contents of the `data-popup-url` attribute.

indigo_app/static/javascript/indigo/utils.js
```diff
@@ -15,3 +15,21 @@ $(function() {
...
+    $(_this).popover({content: html, html: true}).popover('show');
+  });
+  $('.popover').mouseleave(function () {
+    $(_this).popover('hide');
+  });
+}).on('mouseleave','a[href^="/works/"]:not(.btn)', function(){
```
longhotsummer: The scoping on this event should be `a[data-popup-url]` I think?

indigo_app/static/javascript/indigo/utils.js
```diff
@@ -15,3 +15,21 @@ $(function() {
 
   $('body').on('click', 'a[data-confirm], button[data-confirm]', handleConfirm);
 });
+
+// Show popover when hovering on selected links
+$('body').on('mouseenter', 'a[data-popup-url]', function() {
```
longhotsummer: Will this allow us to click on links in the popup?

indigo_app/static/javascript/indigo/utils.js
AliciadW: Yes, it should. I forgot to update this.

indigo_app/static/javascript/indigo/utils.js
AliciadW: Yes, you can still click on a link in the popup

indigo_app/static/javascript/indigo/utils.js
AliciadW: Updated

indigo_app/static/javascript/indigo/utils.js
longhotsummer: Great.

indigo_app/templates/indigo_api/_work_related_table.html
```diff
@@ -14,7 +14,7 @@ <h5 class="card-header">{{ title }}</h5>
     {% for item in items %}
       <tr>
         <td><strong>{{ item.rel|capfirst }}</strong></td>
-        <td><a href="{% url 'work' item.work.frbr_uri %}">{{ item.work.title }}</a></td>
+        <td id="testing"><a href="{% url 'work' frbr_uri=item.work.frbr_uri %}" data-popup-url="{% url 'work_popup' frbr_uri=item.work.frbr_uri %}">{{ item.work.title }}</a></td>
```
longhotsummer: Can drop the id now :)

longhotsummer:
How about some box-shadow on the popover element: `box-shadow: 4px 4px 8px 0px rgba(0, 0, 0, 0.25);`

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -45,7 +45,7 @@ <h4 class="text-muted">{{ work.frbr_uri }}</h4>
               </div>
               {% if work.publication_document %}
               <div>
-                <i class="fas fa-file text-muted"></i> <a href="{% url 'work_publication_document' frbr_uri=work.frbr_uri filename=work.publication_document.filename %}" target="_blank" rel="noopener">{{ work.publication_document.filename }}</a> <span class="text-muted">({{ work.publication_document.size|filesizeformat }})</span>
+                <i class="fas fa-file text-muted"></i> <a href="{% url 'work_publication_document' frbr_uri=work.frbr_uri filename=work.publication_document.filename %}" target="_blank" rel="noopener" data-popup-url="{% url 'work_popup' frbr_uri=work.frbr_uri %}">{{ work.publication_document.filename }}</a> <span class="text-muted">({{ work.publication_document.size|filesizeformat }})</span>
```
goose-life:
```suggestion
                <i class="fas fa-file text-muted"></i> <a href="{% url 'work_publication_document' frbr_uri=work.frbr_uri filename=work.publication_document.filename %}" target="_blank" rel="noopener">{{ work.publication_document.filename }}</a> <span class="text-muted">({{ work.publication_document.size|filesizeformat }})</span>
```

PR #413, opened 2019-02-26
Added assigned_to to list and column views for tasks

AliciadW:
screenshot 2019-02-26 at 12 23 19
screenshot 2019-02-26 at 12 23 27


longhotsummer:
Looks good to me. @goose-life what do you think about having "this task has not been assigned" vs just not showing anything at all?

goose-life:
My inclination would be to have nothing rather – it's already 'open'?

goose-life:
Maybe we eventually want an 'Assign' drop-down in that space?

longhotsummer:
Good idea. @AliciadW if there's no assignee, let's just put nothing for now.

AliciadW:
This has been updated

PR #418, opened 2019-03-05
Stub works

### LINKED ISSUE #262: As a user I want to be able to filter out works that are "stubs" and only simply proclamations of commencements.
For example, difference between primary works (and regulations) and secondary (?) works.

longhotsummer:
Move concept of a stub from a document onto a work.

Fixes #262 

example act with changes laws africa 2019-03-05 16-39-49
edit example act with changes laws africa 2019-03-05 16-39-31
points in time example act with changes laws africa 2019-03-05 16-39-15
south africa laws africa 2019-03-05 16-39-01
south africa laws africa 2019-03-05 16-38-46


indigo_app/views/works.py
```diff
@@ -107,7 +107,6 @@ class EditWorkView(WorkViewBase, UpdateView):
     js_view = 'WorkDetailView'
     form_class = WorkForm
     prefix = 'work'
-    template_name_suffix = '_detail'
```
longhotsummer: I've renamed the file to the `_form` default so we don't need this.

indigo_app/static/javascript/indigo/views/library.js
```diff
@@ -165,11 +179,14 @@
...
+      if (filters.status) {
+        var draft = _.contains(filters.status, 'draft'),
+            published = _.contains(filters.status, 'published');
+
         // filter documents by status
         works = filterWorksByDocs(function(doc) {
-          return (filters.status === "draft" && doc.get('draft') ||
-                  filters.status === "published" && !doc.get('draft'));
+          return (draft && doc.get('draft') ||
+                  published && !doc.get('draft'));
```
goose-life: @longhotsummer This block is taking `works` from 2 stubs down to 0 when `Only stubs` is selected, which stands to reason because the stubs don't have any docs to filter by (if I've understood correctly). When `Exclude stubs` is selected it takes `works` from 11 down to 4, because I have 7 works that aren't explicitly marked as stubs but don't have documents.

longhotsummer: Oh. 🤕

longhotsummer: Fixed.

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -85,6 +85,16 @@ <h5 class="card-header">Work details</h5>
...
+        <div class="form-row">
+          <div class="form-group col-md-6 offset-md-6">
+            <div class="form-check">
+              <input type="checkbox" class="form-check-input" id="{{ form.stub.id_for_label }}" name="{{ form.stub.html_name }}" value="on" {% if form.stub.value %}checked{% endif %}>
+              <label class="mb-0" for="{{ form.stub.id_for_label }}">Stub - this work will have no content</label>
+              <p class="form-text text-muted">Check this option for commencing or amending works that are not captured in detail.</p>
+            </div>
+          </div>
+        </div>
+
```
goose-life: just a note that we may want to be more general here in the future depending on what we encounter. for now we can assume that repealing works aren't generally stubs and that we won't be creating works for notices or similar if we aren't also importing them

indigo_app/templates/indigo_api/work_amendments.html
```diff
@@ -6,6 +6,13 @@
 {% block work-content %}
   <h4>Points in time</h4>
 
+  {% if work.stub %}
+    <p class="alert alert-warning">
+      <i class="fas fa-exclamation-triangle text-warning"></i> Stub works should not have points in time.
+      {% if work.amendments.exists %}Either remove these points in time or don't mark this work as a stub.{% endif %}
+    </p>
+  {% endif %}
+
```
goose-life: should a similar warning also appear on the work overview if there are docs?

indigo_app/templates/indigo_api/work_form.html
longhotsummer: Can you suggest different wording that captures these?

We can indeed change later if necessary.

goose-life: "Check this option for works that are not captured in detail, e.g. commencing or amending works."?

indigo_app/templates/indigo_api/work_amendments.html
longhotsummer: Added.

indigo_app/templates/indigo_api/work_amendments.html
```diff
@@ -6,6 +6,13 @@
 {% block work-content %}
   <h4>Points in time</h4>
 
+  {% if work.stub %}
+    <p class="alert alert-warning">
+      <i class="fas fa-exclamation-triangle text-warning"></i> Stub works should not have points in time.
+      {% if work.amendments.exists %}Either remove these points in time or don't mark this work as a stub.{% endif %}
```
longhotsummer:
```suggestion
      {% if work.document_set.exists %}Either remove these points in time or don't mark this work as a stub.{% endif %}
```

PR #419, opened 2019-03-06
Batch create tasks

### LINKED ISSUE #360: as a contributor I'd like to be spared creating the same task/s on a batch of new works
### Suggestion
Bulk creation of tasks

### More specifically
- [ ] When bulk creating works, give the option to create task/s on each work, eg:
- Upload publication document
- Link commencement / amendments / repeal
- Import content
- Double-check content PiT
- [ ] Have the same option for works that already exist 
- [ ] In both cases, give the option to add the bulk-created tasks to an existing workflow
- [ ] or to create a new one

goose-life:
resolves #360 once ready


indigo_app/views/works.py
```diff
@@ -561,6 +562,23 @@ def get_works(self, table):
...
         return works
 
+    def get_tasks(self, works, form):
+        tasks = []
+        if form.cleaned_data.get('tasks'):
+            for work in works:
+                task = Task()
+                task.country = self.country
+                task.created_by_user = self.request.user
+                task.save()
```
longhotsummer: This seems unnecessary... isn't the one lower down before the append enough?

goose-life: we can't assign task.work if task doesn't have an id yet

longhotsummer: Oh. Django is weird.

indigo_app/views/works.py
```diff
@@ -534,6 +540,8 @@ def get_works(self, table):
                     work.publication_number = row.get('publication_number')
                     work.created_by_user = self.request.user
                     work.updated_by_user = self.request.user
+                    if not row.get('primary'):
```
longhotsummer: Prefer `work.stub = not row.get('primary')`

indigo_app/views/works.py
```diff
@@ -559,6 +567,47 @@ def get_works(self, table):
...
 
+    def get_tasks(self, works, form):
+        tasks = []
+
+        def make_task():
+            task = Task()
+            task.country = self.country
+            task.locality = self.locality
+            task.created_by_user = self.request.user
+            task.title = chosen_task.capitalize()
```
longhotsummer: shouldn't this be take from possible_task['label']?

longhotsummer: where does chosen_task come from?

indigo_app/views/works.py
```diff
@@ -559,6 +567,47 @@ def get_works(self, table):
...
+            task.work = work.get('work')
+            task.save()
+            tasks.append(task)
+
+        # bulk create tasks on primary works
+        if form.cleaned_data.get('primary_tasks'):
+            for work in works:
+                if work['status'] == 'success' and not work['work'].stub:
+                    for chosen_task in form.cleaned_data.get('primary_tasks'):
+                        make_task()
```
longhotsummer: Pass chosen_task in here.

indigo_app/views/works.py
```diff
@@ -559,6 +567,47 @@ def get_works(self, table):
...
+        if form.cleaned_data.get('all_tasks'):
+            for work in works:
+                if work['status'] == 'success':
+                    for chosen_task in form.cleaned_data.get('all_tasks'):
+                        make_task()
+
+        # add all the tasks to the workflows selected
+        if form.cleaned_data.get('workflows'):
+            for workflow in form.cleaned_data.get('workflows'):
+                workflow.tasks = tasks
```
longhotsummer: This will make these tasks the ONLY tasks in this workflow. You probably want to do `workflow.tasks.add(*tasks)`

indigo_app/views/works.py
goose-life: that is much better thank you!

indigo_app/views/works.py
goose-life: yeah this makes way more sense thanks – lmk what you think of my change

indigo_app/views/works.py
goose-life: lmk if my change is what you had in mind here?

indigo_app/views/works.py
```diff
@@ -559,6 +570,42 @@ def get_works(self, table):
...
+            task.created_by_user = self.request.user
+            for possible_task in form.possible_tasks:
+                if chosen_task == possible_task['key']:
+                    task.title = possible_task['label']
+                    task.description = possible_task['description']
+
+            # need to save before assigning work, workflow/s because of M2M relations
+            task.save()
+            task.work = work.get('work')
+            task.workflows = form.cleaned_data.get('workflows').all()
```
longhotsummer: I would have thought this was an array, and `.all()` wouldn't work?

indigo_app/views/works.py
longhotsummer: Yup.

indigo_app/views/works.py
longhotsummer: Looks good. Just check `.all()` works when assigning the workflows to the tasks.

PR #428, opened 2019-03-11
don't show warning about a PiT on a stub if document/s 'deleted'

goose-life:


goose-life:
Screenshot from 2019-03-11 16-26-42
:crossed_fingers: this'll fix the above

PR #433, opened 2019-03-19
docx bulk upload

goose-life:


indigo_app/management/commands/bulk_import_docx.py
```diff
@@ -0,0 +1,114 @@
...
+
+                        try:
+                            importer.create_from_docx(docx_file, document)
+                        except ValueError as e:
+                            print("Error during import: %s" % e.message)
+                            raise ValidationError(e.message or "error during import")
+
+                        # TODO: get rid of `updated_by_user` as this isn't what's happening here
+                        #  (will have to happen on works.ImportDocumentView as well)
+                        document.updated_by_user = user
```
longhotsummer: I don't have a problem with this. created == updated when first created.

indigo_app/management/commands/bulk_import_docx.py
```diff
@@ -0,0 +1,114 @@
...
+        except IOError as e:
+            print('\nFile error: ' + str(e))
+
+    def handle(self, *args, **options):
+        for user in User.objects.all().order_by('id'):
+            print('{}: {} {}'.format(user.id, user.first_name, user.last_name))
+        user = self.get_user()
+        csv_file_name = str(options.get('csv_file'))
+        path = options.get('path')
+        with open(csv_file_name) as csv_file:
```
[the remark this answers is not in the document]
goose-life: super handy, thanks!

indigo_app/management/commands/bulk_import_docx.py
```diff
@@ -0,0 +1,114 @@
...
+            content = csv_file.readlines()
+            content = [r.strip() for r in content]
+            headers = [h for h in content[0].split(',')]
+            rows = [
+                {header: row[i] for i, header in enumerate(headers) if header}
+                for row in [r.split(',') for r in content[1:]]
+            ]
+            for row in rows:
+                with transaction.atomic():
```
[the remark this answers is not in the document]
goose-life: lmk what you think of my changes!

indigo_app/management/commands/bulk_import_docx.py
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            return open(path_to_filename)
+        except IOError as e:
+            print('\nERROR at row {}:'.format(i + 2))
+            print('File error: ' + str(e) + '\n')
+
+    def import_rows(self, user, csv_file, path):
+        content = csv.DictReader(csv_file)
+        for i, row in enumerate(content):
+            with transaction.atomic():
```
longhotsummer: This means that if row 1 succeeds but row 2 fails, row 1 is committed. Is that what you'd like? Another approach is all-or-nothing: wrap the entire batch in a transaction so if any one fails, everything is rolled back.

indigo_app/management/commands/bulk_import_docx.py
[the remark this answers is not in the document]
longhotsummer: Looks good.

indigo_app/management/commands/bulk_import_docx.py
goose-life: it is what I had in mind but I'm open to being persuaded otherwise?

longhotsummer: I tend to prefer all-or-nothing. Let's go with what you have and change if necessary.

indigo_app/management/commands/bulk_import_docx.py
```diff
@@ -89,25 +92,18 @@ def import_docx_file(self, user, work, date, language, docx_file):
 
         # hard-coded for Namibian docxes
         importer.section_number_position = 'after-title'
+        upload = UploadedFile(file=docx_file,
+                              content_type='application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+                              size=filesize)
```
longhotsummer: nice :)

Issue #435, opened 2019-03-20
link amendments, commencements when doing bulk work creation


goose-life:
Go over works created in a batch import:

- [ ] Check for `commenced_by` & look for the commencing work
- [ ] Make a task if more than one match for the title (One day we'll link the spreadsheet to have the frbr_uri written to it first; for now we'll check for a matching title)
- [ ] Check for `amends` – if one title only (this will involve some guesswork), link amendments
- [ ] If more than one amendment or no match, create a task
- [ ] Check for `repealed_by` and link if match
- [ ] If no match, create a task

This will be a good improvement on the blanket `link amendments / commencements / repeals` task, which should be removed once this is implemented.

goose-life:
closed by #438

PR #438, opened 2019-03-20
Batch link related works

goose-life:


goose-life:
closes #435

indigo_app/templates/indigo_api/work_new_batch.html
```diff
@@ -110,7 +110,8 @@ <h5>Tasks for all works</h5>
 
         </div>
 
-        <label for="{{ form.workflows.id_for_label }}" class="h5">Add the tasks to the following workflows</label>
+        <label for="{{ form.workflows.id_for_label }}" class="h5">Workflows</label>
```
longhotsummer: If this is "less optional" than the tasks, perhaps move it above the tasks?

indigo_app/views/works.py
```diff
@@ -559,6 +560,16 @@ def get_works(self, table):
                         info['status'] = 'success'
                         info['work'] = work
 
+                        # TODO: neaten this up
+                        if not row.get('commenced_by') == u"":
```
longhotsummer: how about `if row.get('commenced_by'):` because non-empty strings are thruthy, empty strings are falsey.

longhotsummer:
@goose-life when looking up the linked works, I think we should filter by country. There's a good chance of works with matching titles in other countries.

indigo_app/views/works.py
goose-life: of course, thanks – I initially just had it as `info['commenced_by'] = row.get('commenced_by')` and relied on this to be `None` or similar lower down, but that didn't work, so I went overboard in the other direction

PR #439, opened 2019-03-25
Display parent work for subsidiary legislation

### LINKED ISSUE #268: Regulations with a parent work should be more obviously displayed; both in the listing and on a detail page for the child work; also in the final HTML/whatever.

goose-life:
closes #268


goose-life:
Issue specified it should be 
> both in the listing and on a detail page for the child work; also in the final HTML/whatever

So far it's only done on the listing page; the parent work is already listed on the work overview.
Re the final output: should it be added to the coverpage? 
On the end product side, can they work with `parent_work` to display a work and its children together?

indigo_app/templates/place/detail.html
```diff
@@ -86,6 +86,11 @@
               {{/if}}
               <br>
               <span class="text-muted">{{ frbr_uri }}</span>
+              {{#if parent_work}}
+              <p>
```
longhotsummer: `p` has bottom padding, you probably just want a `div`.

Can you include a screenshot?

longhotsummer:
On the product side they can use `parent work` to do whatever is needed.

I'm not yet sure how to add it to the coverpage, we'll need an example such as subsidiary legislation from KL. Let's keep #268 open for now with just that issue (or create a new issue), so we don't block all these changes with deciding what the external view looks like.

goose-life:
Screenshot from 2019-03-25 11-28-09
Screenshot from 2019-03-25 11-29-14

goose-life:
Screenshot from 2019-03-25 11-32-12
this is with `<p>` as I had it initially

goose-life:
new issue #445 opened

PR #440, opened 2019-03-25
Strip unwanted characters when doing a batch import

### LINKED ISSUE #368: as a user I don't want to worry about accidentally pulling in odd characters
extra-character

goose-life:
closes #368


indigo_app/views/works.py
```diff
@@ -702,6 +702,10 @@ def make_date(self, string, field):
                 raise ValidationError('Check the format of %s; it should be e.g. "2012-12-31"' % field)
         return date
 
+    def strip_title_string(self, title_string):
+        no_line_separator = re.sub(u'\u2028', ' ', title_string)
+        return re.sub('  ', ' ', no_line_separator)
```
longhotsummer: If the string has three spaces, this will result in two spaces. Another way of doing this which will handle multiple spaces is:

`re.sub(u'[\u2028 ]+', ' ', title_string)`

ie. replace one or more of either a space or `\u2028` with a single space.

PR #442, opened 2019-03-25
Render part heads correctly

### LINKED ISSUE #430: When rendering a part heading, don't include a hyphen if there is no heading

goose-life:
closes #430


indigo_za/toc.py
```diff
@@ -24,7 +25,7 @@ class TOCBuilderZA(TOCBuilderBase):
     toc_non_unique_components = ['chapter', 'part']
 
     titles = {
-        'chapter': lambda t: _('Chapter') + ' %s' % t.num + (' - %s' % t.heading if t.heading else ''),
-        'part': lambda t: _('Part') + ' %s' % t.num + (' - %s' % t.heading if t.heading else ''),
+        'chapter': lambda t: _('Chapter') + ' %s' % t.num + (' – %s' % t.heading if t.heading else '').decode('utf-8'),
```
longhotsummer: Use of this decode is weird, it's probably their because otherwise it complains about ascii encodings? That'll be because it's trying to combine ascii and unicode strings. You can either make all of these string unicode (with `u'foo'`) or put `from __future__ import unicode_literals` at the top of the file. In Python 3, all literals are unicode.

longhotsummer:
Nice :)

PR #452, opened 2019-03-27
Link publication documents

### LINKED ISSUE #444: When creating a new work (or bulk import) automatically link to publication document
In most cases this can be done automatically.

* [ ] on bulk import, create a task if fails
* [ ] when editing or creating a work, when the publication docs are loaded and one isn't assigned yet AND there's only one match, automatically assign it.

goose-life:
will close #444 

so far it's only done on bulk creation – still to do on individual works


goose-life:
only doing on bulk creation – already very easy to attach on work form

PR #478, opened 2019-04-04
publication documents backfill task tweaks

goose-life:


indigo_app/management/commands/backfill_publication_docs.py
```diff
@@ -46,53 +47,50 @@ def get_params(self, work):
...
         return params
 
-    def pub_doc_task(self, work, user, task_type):
-        task = Task()
+    def pub_doc_task(self, work, user):
+        task_title = 'Link publication document'
 
-        self.stdout.write(self.style.NOTICE("Creating {} task for {}".format(task_type, work)))
+        try:
+            Task.objects.get(title=task_title, work=work)
```
longhotsummer: What if it's closed/done?

indigo_app/management/commands/backfill_publication_docs.py
```diff
@@ -46,53 +47,50 @@ def get_params(self, work):
...
-Otherwise, find it and upload it manually.'''
+        except Task.DoesNotExist:
+            task = Task()
 
-        elif task_type == 'check':
-            task.title = 'Check publication document'
-            task.description = '''This work's publication document was linked automatically.
-Double-check that it's the right one.'''
-            task.state = 'pending_review'
+            self.stdout.write(self.style.NOTICE("Creating ‘Link publication document’ task for {}".format(work)))
```
longhotsummer: I must be honest that I'm skeptical of using unicode chars in log messages. In this case it's highly unlikely to impact anything, but in general if I'm using grep or similar to search through log messages (common), I don't want to have to deal with trying to find the correct quotation mark.

goose-life: fair enough! going to revert them all seeing as we're also not using those quotes anywhere else :)

PR #484, opened 2019-04-09
View tasks on work improved

### LINKED ISSUE #449: when viewing tasks on a work I want to see closed and cancelled tasks too
There's a list of open tasks at the bottom of a work overview, and a `View all tasks` button, which directs to eg `https://edit.laws.africa/places/na/tasks/?frbr_uri=/na/act/2000/27`.
However,
1. The default is to see only open / assigned / pending_review tasks, which I think should be overridden for this particular view (it does promise `all tasks`)
2. `&frbr_uri=/na/act/2000/27` isn't passed along when changing the status filter

goose-life:
closes #449

### DIFF
--- a/indigo_app/urls.py
+++ b/indigo_app/urls.py
@@ -23,6 +23,7 @@ urlpatterns = [
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/activity$', places.PlaceActivityView.as_view(), name='place_activity'),
 
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/$', tasks.TaskListView.as_view(), name='tasks'),
+    url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks(?P<frbr_uri>/\S+?)/$', tasks.TaskListView.as_view(), name='tasks_on_work'),
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/new$', tasks.TaskCreateView.as_view(), name='create_task'),
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/$', tasks.TaskDetailView.as_view(), name='task_detail'),
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/edit$', tasks.TaskEditView.as_view(), name='task_edit'),
--- a/indigo_app/views/tasks.py
+++ b/indigo_app/views/tasks.py
@@ -55,13 +55,15 @@ class TaskListView(TaskViewBase, ListView):
 
     def get_queryset(self):
         tasks = Task.objects.filter(country=self.country, locality=self.locality).order_by('-updated_at')
-        return self.form.filter_queryset(tasks, frbr_uri=self.request.GET.get('frbr_uri'))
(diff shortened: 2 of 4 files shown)


indigo_app/urls.py
```diff
@@ -23,6 +23,7 @@
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/activity$', places.PlaceActivityView.as_view(), name='place_activity'),
 
     url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/$', tasks.TaskListView.as_view(), name='tasks'),
+    url(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks(?P<frbr_uri>/\S+?)/$', tasks.TaskListView.as_view(), name='tasks_on_work'),
```
longhotsummer: This is going to get confusing -- which URL do we submit the filter form to? The one above or this one?

Let's use the normal task listing url, but include the `frbr_uri` as a query string field, like any other filter field.

Otherwise, we're essentially adding a new view entirely.

indigo_app/views/tasks.py
```diff
@@ -38,7 +38,10 @@ def get(self, request, *args, **kwargs):
         params.update(request.GET)
 
         # initial state
-        if not params.get('state'):
+        if self.kwargs.get('frbr_uri') and not params.get('state'):
```
longhotsummer: Why would showing a work now show done and cancelled tasks? That seems odd to me. Couldn't it just be a new filter field? The user can always opt into the other task states by adjusting the filters.

indigo_app/views/tasks.py
```diff
@@ -52,13 +55,15 @@ def get(self, request, *args, **kwargs):
 
     def get_queryset(self):
         tasks = Task.objects.filter(country=self.country, locality=self.locality).order_by('-updated_at')
-        return self.form.filter_queryset(tasks, frbr_uri=self.request.GET.get('frbr_uri'))
+        return self.form.filter_queryset(tasks, frbr_uri=self.kwargs.get('frbr_uri'))
```
longhotsummer: As mentioned above with the FRBR URI, I think it's simpler to make the work just another filter. This would allow selecting multiple works, too.

indigo_app/urls.py
goose-life: we would be adding a new (template) view, but we'd be using the same (view) controller – and if our only aim was to view all tasks on one work I think it would work well?
I'm not sure I get the use case for filtering on multiple works, and thought this would be the simplest way of making sure that we 
a) display visually that we're looking at tasks on a particular work and 
b) retain the FRBR URI of that work in the context without having to pass it along unnecessarily for every other task listing view

indigo_app/views/tasks.py
goose-life: My understanding was that we'd use workflows to group tasks that belonged to certain works together, but that we'd sometimes want to see all the tasks on a work to see what had already been resolved / still needed to be dealt with, especially before publishing.
I also think that if we had to have all the works for a place as a multi select dropdown in the task listing view that could get pretty hairy pretty quickly.

Let's chat about this before I make any other changes to make sure we're on the same page?

indigo_app/urls.py
goose-life: I think I misunderstood the concern here – are you talking about a filter form that doesn't exist yet?

goose-life:
closing this in favour of #488

PR #485, opened 2019-04-09
Task workflows UI

longhotsummer:
Improved way of changing workflows in task view. I created a new URL for this, because it's not possible to submit parts of a form (eg. workflows only) to the 'edit' endpoint.

Also:

* permissions
* layouts

Tester task – Kenya Law 2019-04-09 16-58-56

Tester task – Kenya Law 2019-04-09 16-59-07


indigo_app/views/tasks.py
```diff
@@ -323,3 +325,33 @@ def get_redirect_url(self):
...
+        ids = self.request.POST.getlist('workflows')
+
+        if ids:
+            workflows = Workflow.objects.filter(country=task.country, locality=task.locality, id__in=ids).all()
+        else:
+            workflows = []
+
+        task.workflows.set(workflows)
+
+        # TODO: timeline?
```
longhotsummer: @goose-life I should be doing "added to" and "removed from" here, right?

goose-life:
> This is a bit cheeky - it doesn't refresh the list of workflows, it only updates the selected items in the checkbox. it's quick and easy, and I hope it suffices. Another option is to immediately refresh the page when a workflow is chosen. This makes it a bit awkward to choose more than one, or deselect + select, but refreshes the page (this is the behaviour we have in the dropdowns on the task filtering pages). The latter is even easier to implement than what we've done here.
> 
> Which do you think is better?

I personally prefer it not refereshing every time, it's confusing as a user when there are multiple choices but it refreshes every time a single change is made?

indigo_app/views/tasks.py
goose-life: https://github.com/laws-africa/indigo/blob/9750dd3b4508e38c8ec1920821db8490d2711411/indigo_app/views/tasks.py#L177-L193

indigo_app/templates/indigo_api/task_detail.html
```diff
@@ -141,18 +141,32 @@ <h6>Labels</h6>
...
+          {% endfor %}
+        </select>
+      </form>
+      <ul class="list-unstyled mb-2">
         {% with task.workflows.all as workflows %}
         {% if workflows %}
-          {% for workflow in workflows %}
-            <li><a href="{% url 'workflow_detail' place=place.place_code pk=workflow.pk %}">{{ workflow.title }}</a></li>
+          {% for workflow in workflows|dictsort:'title' %}
+            <li><a href="{% url 'workflow_detail' place=place.place_code pk=workflow.pk %}">#{{ workflow.pk }} - {{ workflow.title }}</a></li>
```
goose-life:
```suggestion
            <li><a href="{% url 'workflow_detail' place=place.place_code pk=workflow.pk %}">#{{ workflow.pk }} – {{ workflow.title }}</a></li>
```
Also, why include workflow ids here (and not in dropdown)? (Not sure they need to be included at all?)

indigo_app/views/tasks.py
goose-life: @longhotsummer I'm making a PR onto your branch for these

indigo_app/templates/indigo_api/task_detail.html
longhotsummer: Good point, I think we should have them everywhere. It makes it much easier to refer to them unambiguously, since it looks like we're going to have many open for a place at once. Same reason we expose the task id.

longhotsummer:
> > This is a bit cheeky - it doesn't refresh the list of workflows, it only updates the selected items in the checkbox. it's quick and easy, and I hope it suffices. Another option is to immediately refresh the page when a workflow is chosen. This makes it a bit awkward to choose more than one, or deselect + select, but refreshes the page (this is the behaviour we have in the dropdowns on the task filtering pages). The latter is even easier to implement than what we've done here.
> > Which do you think is better?
> 
> I personally prefer it not refereshing every time, it's confusing as a user when there are multiple choices but it refreshes every time a single change is made?

Agreed, I've left it as it is then, where the whole page doesn't refresh.

PR #486, opened 2019-04-10
send action signals for 'added to' and 'removed from' workflow/s

goose-life:


longhotsummer:
I did this on the plane with a slightly different approach which reduces duplication.

PR #487, opened 2019-04-10
Task permissions review

### LINKED ISSUE #479: Tasks should only be able to be assigned to users that meet permission requirements
Currently I can assign a task to anyone with a country badge.

* Doing a task - Just `indigo_api.change_task`?
* Review a task - `indigo_api.close_task`

We need to be able to handle this in bulk.

goose-life:
closes #479


indigo_app/views/tasks.py
```diff
@@ -62,10 +63,16 @@ def get_context_data(self, **kwargs):
         context['task_groups'] = Task.task_columns(self.form.cleaned_data['state'], context['tasks'])
 
         # potential assignees for tasks. better to batch this here than load it for every task.
-        permitted_users = User.objects.filter(editor__permitted_countries=self.country).all()
+        submit_task_pk = Permission.objects.get(codename='submit_task').pk
+        close_task_pk = Permission.objects.get(codename='close_task').pk
+        potential_assignees = User.objects.filter(editor__permitted_countries=self.country,
```
longhotsummer: This logic looks fine, the use of primary keys surprises me. Does `User.objects.filter(..., user_permissions=submit_task)` where `submit_task` is the permission itself, and not the pk, not work?

Using objects themselves is generally the preferred Django way.

indigo_app/templates/indigo_api/_task_cards.html
```diff
@@ -65,26 +65,48 @@ <h6 class="text-center mb-4">
...
+                            {% endfor %}
+                          {% endif %}
                         </div>
                       </div>
                     </form>
                   {% endif %}
 
                 </div>
-              {% elif task.state == 'open' or task.state == 'pending_review' %}
+              {% elif task.state == 'open' %}
```
longhotsummer: These two blocks are identical except for `potential_assignees` vs `potential_reviewers`. Rather than duplicating so much html, can we not decide on which one to use at the last minute? Alternatively, isn't a `potential_assignees` always `potential_reviewers` for tasks that are waiting to be reviewed?

indigo_app/views/tasks.py
goose-life: thanks I'll try it! I had tried a couple of other things that didn't work, but not this

indigo_app/templates/indigo_api/_task_cards.html
goose-life: quite right – I've changed it so that `potential_assignees` points to the right thing on each task (or at least that was the aim)

indigo_app/views/tasks.py
```diff
@@ -74,12 +74,7 @@ def get_context_data(self, **kwargs):
         context['form'] = self.form
         context['frbr_uri'] = self.request.GET.get('frbr_uri')
         context['task_groups'] = Task.task_columns(self.form.cleaned_data['state'], context['tasks'])
-
-        # potential assignees for tasks. better to batch this here than load it for every task.
-        permitted_users = User.objects.filter(editor__permitted_countries=self.country).all()
-        for task in context['tasks']:
-            # this overwrites the task's potential_assignees method
-            task.potential_assignees = [u for u in permitted_users if task.assigned_to_id != u.id]
+        context['potential_assignees'] = Task.decorate_potential_assignees(context['tasks'], self.country)
```
longhotsummer: Since it decorates the tasks directly, you don't need to assign to context here.

indigo_app/views/tasks.py
```diff
@@ -107,6 +102,7 @@ def get_context_data(self, **kwargs):
         if has_transition_perm(task.close, self):
             context['close_task_permission'] = True
 
+        context['potential_assignees'] = Task.decorate_potential_assignees([task], self.country)
```
longhotsummer: As above.

PR #488, opened 2019-04-12
Tasks on work

goose-life:


longhotsummer:
Fab, nice and easy.

indigo_app/templates/indigo_api/work_tasks.html
```diff
@@ -0,0 +1,37 @@
+{% extends "indigo_api/work_layout.html" %}
+
+{% block title %}Tasks – {{block.super }}{% endblock %}
+
+{% block work-content %}
+  <div class="container mt-3">
```
longhotsummer: I don't think you need this container. It's doubled up from the parent template.

PR #490, opened 2019-04-16
task permissions

goose-life:
This will make it impossible for users to submit tasks that are assigned to someone else, and for users to close their own tasks.


longhotsummer:
Let's allow superusers (later, place admins) to do this too.

PR #492, opened 2019-04-16
Crossheadings, longtitle, b and i

longhotsummer:


longhotsummer:
These are now supported in the grammar.

PR #493, opened 2019-04-17
Fix linking to sections in schedules

longhotsummer:
This fixes linking to sections in schedules. Before this, if a document had a section (or part or chapter) numbered the same in the main body and a schedule, linking to it from the task detail view wouldn't work correctly.

This fixes how we generate and resolve anchors, both in the TOC view in the document view, and on the server when looking up an annotation.

### DIFF
--- a/indigo/analysis/toc/base.py
+++ b/indigo/analysis/toc/base.py
@@ -46,7 +46,7 @@ class TOCBuilderBase(LocaleBasedMatcher):
     """ The locale this TOC builder is suited for, as ``(country, language, locality)``.
     """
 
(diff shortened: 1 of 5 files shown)


goose-life:
https://github.com/laws-africa/indigo/blob/db0eb77dafb8885e4c0aa985c0b975a1ec5f6c26/indigo/documents.py#L46
returns `None` when `anchor` is `{'id': u'schedule/schedule.paragraph0'}`, resulting in an error on the next line:
`AttributeError: 'NoneType' object has no 
Screenshot from 2019-04-17 10-50-08
attribute 'element'`

longhotsummer:
schedule:schedule paragraph0 : aoeu – Laws Africa 2019-04-17 12-03-44

PR #507, opened 2019-04-22
Indigo-web 3.0.0 with PT Serif

longhotsummer:


indigo_app/static/lib/indigo-web-3.0.0/scss/akoma-ntoso.scss
```diff
@@ -0,0 +1,15 @@
+/*
+ * Indigo Web CSS for Akoma Ntoso
+ *
+ * Copyright 2016 Code for South Africa
```
goose-life: still C4SA?

longhotsummer: This is from https://github.com/laws-africa/indigo-web -- it may actually be AfricanLII, I'll have to check when we built it. Otherwise it's my copyright.

PR #517, opened 2019-04-26
Support arbitrary key-value properties

longhotsummer:
This allows us to capture things like Cap numbers in Kenya. It should permit some other useful things in future.

Edit – Advocates Act – Kenya Law 2019-04-26 12-14-57

Advocates Act – Kenya Law 2019-04-26 12-15-10


longhotsummer:
Still need to add support for the create work form.

goose-life:
1 should we leave out 'add new property' for now?
2 if we keep it / reintroduce it later, the form doesn't give the option to enter a new key, only a value

longhotsummer:
Should we change it to "Choose a property to set" or something?

longhotsummer:
We now show all properties, and it's much easier for the user. hooray.

Edit – Advocates Act – Kenya Law 2019-04-29 14-45-01

longhotsummer:
indigo-kenyalaw will update `WorkProperty.KEYS` to have a list of options. At some point we'll allow this to be configured per place.

PR #518, opened 2019-04-26
Correctly set STATIC_ROOT

longhotsummer:
This must be an absolute directory. Otherwise, when
PDF rendering rewrites static file URLs to be absolute, it gets
them wrong


longhotsummer:
Merging since Edith is unavailable.

Issue #495, opened 2019-04-17
As a contributor I'd like an email when a task is assigned to me


longhotsummer:
The magic should happen around about here: https://github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L315

What I'd like to do is to delay the email by 60 seconds or so. That will allow us to unassign a mistaken assignment. This will add some additional complexity. We use [django-background-tasks](https://django-background-tasks.readthedocs.io/en/latest/) to run background tasks. So, I suggest the trigger queues up a background task with a 60 second delay. That can be defined in `indigo_app/tasks.py`. That should then check if the task is still currently assigned to the expected user, and send the email if it is.

We'll need to collaborate on what the email will look like. We don't yet have templates for HTML emails.

longhotsummer:
Actually, it may be a good idea to rely on the `activity` events, which are essentially our audit trail. We work hard to ensure those are created at the right time, and tying into that would simplify things. Coupled with that, I think we're going to quickly need common requirements around email notifications.

So, perhaps an idea is to have `indigo_app/notifications.py` with a singleton `notifier` which then responds to the `save` django model event on the `Action` model, and queues up whatever email is necessary.

So, logically, it would be:

1. task is assigned
2. task view creates the action/audit trail object
3. event handler responds to the save event on the action object
4. handler triggers a delayed task
5. delayed task executor calls the notifier to actually send the email

PR #523, opened 2019-04-30
#495

allanice001:
Adding the ability to send mails on task assignment from django core


longhotsummer:
Looking good! A few thoughts:

1. It'll be useful for us to use templates for the emails, even if they're just plain text. That sets us up nicely to use common layouts, etc. Please can you check how Django's email templates are done and delegate the message body to a template.

2. I'd like to drive the email off of saving the `action` model. We're already ensuring that the actions are created at the right places, so they're a good proxy for understanding when a notification needs to go out -- they also have all the context (who did what to whom etc.)

longhotsummer:
Thanks @allanice001 - as discussed, let's use https://github.com/vintasoftware/django-templated-email as a dependency if at all possible, rather than pulling it in. I'm confident this can be implemented in one file (plus templates) with a few lines of code.

PR #536, opened 2019-05-09
Feature comments

kaysiz:


indigo_app/templates/indigo_api/task_detail.html
```diff
@@ -75,7 +87,31 @@ <h5>{{ task.title }}</h5>
...
-
+      {% get_comment_form for task as form %}
+        <table>
+          <form action="{% comment_form_target %}" method="post">
+            {% csrf_token %}
+            <textarea
+                    id="id_comment"
+                    name="comment"
+                    class="form-control"
+                    required=""
```
longhotsummer: `required` doesn't need a value, you can just have `required`.

I suggest pulling these onto the same line as the textarea (wrapping if it's too wide), rather than having many lines of attributes.

indigo_app/templates/indigo_api/task_detail.html
```diff
@@ -75,7 +87,31 @@ <h5>{{ task.title }}</h5>
           </li>
         {% endif %}
       </ul>
-
+      {% get_comment_form for task as form %}
```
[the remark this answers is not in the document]
kaysiz: i left it intentionally after seeing that non-authenticated users could not see the task, so assumption was anyone seeing the task is already logged in. Advise if i should continue to add it.

indigo_app/templates/indigo_api/task_detail.html
kaysiz: noted.

indigo_app/templates/indigo_api/task_detail.html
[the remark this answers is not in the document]
longhotsummer: The site is read-only for anonymous users, so anyone can see it. So we do need to protect against non-authenticated users from both seeing the form and submitting it (I think submission permissions can be configured with django-contrib-comments).

indigo_app/templates/indigo_api/task_detail.html
```diff
@@ -75,7 +87,26 @@ <h5>{{ task.title }}</h5>
           </li>
         {% endif %}
       </ul>
-
+      {% if user.is_authenticated %}
+        {% get_comment_form for task as form %}
+              <form action="{% comment_form_target %}" method="post">
```
longhotsummer: Pls double check the whitespace: we use two-space indents in .html files.

kaysiz:
Left with merging tasks and comments and rendering a sorted list in the template

indigo_app/templates/indigo_api/task_detail.html
```diff
@@ -1,178 +1,245 @@
...
+            {% csrf_token %}
+            <textarea id="id_comment" name="comment" class="form-control" maxlength="3000" rows="10"
+                      cols="80" placeholder="Add comment" required></textarea>
+            <input type="hidden" name="honeypot" id="id_honeypot">
+            {{ form.content_type }}
+            {{ form.object_pk }}
+            {{ form.timestamp }}
+            {{ form.security_hash }}
+            <div class="form-group">
+              <input type="hidden" name="next"
```
longhotsummer: Minor comment: with HTML 5, unclosed tags (like `br` and `input`) don't need to be self-closed, they can simply exist without a closing tag.

kaysiz: thank you for the info.

kaysiz:
i added the django-contrib-comments package in the setup.py file manually after pip freeze, please advise if this is the correct way.

longhotsummer:
Adding django-contrib-comments to setup.py is fine, thanks. We normally don't anchor the exact version, that allows consumers to pick a different version. We'll normally indicate the package should be at least the version we've tested with (normally the latest version when we first add it): `foo>=1`

setup.py
```diff
@@ -75,6 +75,7 @@
         'requests>=2',
         'unicodecsv>=0.14.1',
         'whitenoise<2,>=1.0.6',
+        'django-contrib-comments==1.9.1',
```
longhotsummer:
```suggestion
        'django-contrib-comments>=1.9.1',
```

Issue #467, opened 2019-04-01
I'd like to progress a task in a workflow without leaving the page


goose-life:
I'd like to make state changes (from `open` to `pending_review` to `closed` etc) to tasks in a workflow without going to individual tasks' detail views – this will save time especially on 'check publication document' tasks, of which we will have many.

longhotsummer:
An option is to have state changes in the dropdown menu for the task.

goose-life:
What we're envisioning is, in this view:
![Screenshot from 2019-04-16 13-07-28](https://user-images.githubusercontent.com/32566441/56204844-ab569200-6048-11e9-9faa-a5d633527d8b.png)

in the dropdown of each task, which currently contains the option to remove form the workflow:
https://github.com/laws-africa/indigo/blob/95c778af9832adc373f9bff1ed51800766a1b701/indigo_app/templates/indigo_api/_task_cards.html#L17-L28

to add the following options:

- [ ] if in the 'Open' column – nothing

- [ ] if in the 'Assigned' column (these are tasks in the `open` state that have an `assigned_to`):
option to 'Submit for review', similar to:
https://github.com/laws-africa/indigo/blob/95c778af9832adc373f9bff1ed51800766a1b701/indigo_app/templates/indigo_api/task_detail.html#L81-L83
- [ ] we want `submit_task_permission` to include the requirement that the task is assigned to the current user – I'll do that part

- [ ] if in the 'Pending review' column:
option to either 'Unsubmit' or 'Close', similar to:
https://github.com/laws-africa/indigo/blob/95c778af9832adc373f9bff1ed51800766a1b701/indigo_app/templates/indigo_api/task_detail.html#L91-L97
- [ ] we want `unsubmit_task_permission` and `close_task_permission` to include the requirement that the current user is both a Senior Drafter and that the task wasn't last_assigned to them – I'll do that part

- [ ] if in the 'Done' or 'Cancelled' columns: nothing.

longhotsummer:
LGTM.

See #490 for task permission changes.

AliciadW:
![Screenshot 2019-05-15 at 15 22 55](https://user-images.githubusercontent.com/13785692/57778931-6ac06600-7725-11e9-836c-91169272f339.png)
![Screenshot 2019-05-15 at 15 22 46](https://user-images.githubusercontent.com/13785692/57778935-6dbb5680-7725-11e9-97f4-f72fe17119f8.png)
[...]

PR #554, opened 2019-05-16
467 progress task without leaving page

AliciadW:


AliciadW:
Fixes #467

longhotsummer:
@goose-life These options show up whether or not the user has permissions to do them, right? Which could be confusing... How do the permissions from #490 get used in the view?

goose-life:
@longhotsummer good point! @AliciadW, we probably want to do something similar to the task detail view (`submit_task_permission`, `reopen_task_permission`, etc):
https://github.com/laws-africa/indigo/blob/d557a975dbe40e59dafe2b62a353fcdd6dd1b5e3/indigo_app/templates/indigo_api/task_detail.html#L83

which gets its context here:
https://github.com/laws-africa/indigo/blob/f07ccbaa869e9775283660419614b5ccf9706192/indigo_app/views/tasks.py#L86

will that be fairly simple to pull through?

goose-life:
@longhotsummer should we move all those permissions things into TaskViewBase?

longhotsummer:
> @longhotsummer should we move all those permissions things into TaskViewBase?

The business logic lives in the task, which is great. The permissions need to show up in TaskListView, WorkflowDetailView and TaskDetailView. They don't have a good common ancestor. So either we need a mixin, or we just copy-and-paste the `get_context_data` code:

```python
context['close_task_permission'] = has_transition_perm(task.close, self)
# etc.
```

AliciadW:
@goose-life looking at this now.

PR #555, opened 2019-05-16
allow progression of tasks according to permissions

goose-life:
I think this should address @longhotsummer's concern on #554


longhotsummer:
I like it. It'll work for now, we can improve later. Simpler than a mixin!

PR #574, opened 2019-05-23
My tasks

AliciadW:
@longhotsummer please review and let me know if there are any further changes that you would like me to make at this time.

I have pushed the version that has the grey `?` icon in the pending tasks cards, but have attached a screenshot of what it will look like if we add the orange block around the icon as per the pending review column on the Workflows page.

I do think we might consider rather using the clock that is currently in the designs instead of the `?`.

Also, please note that the columns are scrollable and I have set a max-height of `70vh`.

Screenshot 2019-05-23 at 19 33 01
Screenshot 2019-05-23 at 19 33 22


indigo_app/static/stylesheets/_tasks-view.scss
```diff
@@ -0,0 +1,39 @@
...
+  justify-content: space-between;
+  align-items: center;
+}
+
+.task-list-card-heading {
+  text-decoration: underline;
+  color: black;
+}
+
+.task-list-card-approved-column {
```
longhotsummer: I suggest we make these border decorators something like `task-border-<<task-state>>`, eg `task-border-pending_review`, `task-border-assigned`, etc. That way they're programmatic and can be used anywhere to add a border coloured based on the state.

indigo_app/templates/indigo_app/tasks/my_tasks.html
```diff
@@ -0,0 +1,94 @@
...
+
+    <div class="col-md-3">
+      <div class="card task-list-card-full-height task-list-card-radius">
+        <h6 class="text-center my-3">Tasks waiting for approval</h6>
+
+        <div class="task-list-card-scroll">
+          {% for task in tasks_pending_approval %}
+          <div class="task-list-card my-2 mx-3 p-2 task-list-card-radius task-list-card-approval-task">
+            <div class="approval-heading-span">
+              <div class="text-muted">#{{ task.pk }}</div><i class="fas fa-fw task-icon-{{ task.state }} text-muted"></i>
```
longhotsummer: I think we should be consistent with iconography and use the same task indicators everywhere. I know right now the icons may not align with the new task design, but let's include them and then evolve them everywhere. They may look a bit better wrapped in `<div class="small">`?

indigo_app/static/stylesheets/_tasks-view.scss
```diff
@@ -0,0 +1,39 @@
+// helper classes
+.task-list-card-full-height {
+  height: 100%;
+  max-height: 70vh;
```
longhotsummer: This works better than I expected - how about making it 80vh?

indigo_app/templates/indigo_app/tasks/my_tasks.html
```diff
@@ -0,0 +1,94 @@
+{% extends "main.html" %}
```
longhotsummer: Pls add a page title (override the `title` block)

longhotsummer:
Oh, and we'll need provision for text in a column if a column is empty.  eg.

assigned: `You don't have any assigned tasks right now.<br><button>Browse open tasks</button>.`

longhotsummer:
And don't forget to include the task labels, pls.

longhotsummer:
And on the pending review tasks, pls show which user it is assigned to (may be none) - perhaps just using the same mechanism as the current task list.

indigo_app/static/stylesheets/_tasks-view.scss
AliciadW: Updated

indigo_app/static/stylesheets/_tasks-view.scss
AliciadW: Done

indigo_app/templates/indigo_app/tasks/my_tasks.html
AliciadW: Done, I've added the icons for all three columns.

indigo_app/templates/indigo_app/tasks/my_tasks.html
AliciadW: Updated to `My tasks`

AliciadW:
Screenshot 2019-05-27 at 08 36 40
Screenshot 2019-05-27 at 08 37 13

longhotsummer:
I like! @goose-life your thoughts? Should we deploy and play with it?

indigo_app/templates/indigo_app/tasks/my_tasks.html
```diff
@@ -9,31 +11,37 @@
...
-            {% endif %}
-
-          </div>
-          {% endfor %}
+            {% endfor %}
+          {% else %}
+            <p class="mx-3">
+              You don't have any assigned tasks right now.
+            </p>
+            <a href="#" class="btn btn-outline-primary ml-3">Browse open tasks</a>
```
AliciadW: should go to:
`{% url 'tasks' place=place.place_code %}`

longhotsummer:
@AliciadW I've merged master into the branch, which adds `task.closed_by_user`, which is the user that approved a task (if it's showing up in the approval column). Please can you factor that in.

Depending on @goose-life's feedback, I think we can then deploy.

longhotsummer:
I'm still partial to having the coloured icons since the colour has significance. @goose-life ?

goose-life:
keen to deploy, we can always tweak as we go!

PR #580, opened 2019-05-26
create models for VocabularyTopic and TaxonomyVocabulary

kaysiz:
- Setup new models for TaxonomyVocabulary and VocabularyTopic. 
- Setup admin interface. 
Next step
 - allow user to choose taxonomy entries in work form
 - enable user to see taxonomy details in work detail view
 - create API for listing taxonomies (/za/taxonomies/)


kaysiz:
Need help writing tests for the Models, following the structure and convention already setup in the codebase.

kaysiz:
@longhotsummer is this a good route to implement 'user can choose taxonomy entries in work form - EditWorkView', https://select2.org/getting-started/basic-usage. The multiple select option will work well with our use case.

indigo_api/models.py
```diff
@@ -151,6 +151,28 @@ def get_queryset(self):
...
+    title = models.CharField(max_length=30, null=False, unique=True, blank=False)
+
+    def __unicode__(self):
+        return unicode(self.name)
+
+
+class VocabularyTopic(models.Model):
+    taxonomy_vocabulary = models.ForeignKey(TaxonomyVocabulary, related_name='vocabularies', null=False, blank=False, on_delete=models.CASCADE)
+    level_1 = models.CharField(max_length=30, null=False, unique=True, blank=False)
+    level_2 = models.CharField(max_length=30, null=False, unique=True, blank=False)
```
longhotsummer: level_2 can be null and black, and doesn't need to be unique.

indigo_api/models.py
```diff
@@ -188,6 +210,9 @@ class Meta:
 
     stub = models.BooleanField(default=False, help_text="Stub works do not have content or points in time")
 
+    # taxonomies
+    taxonomies = models.ManyToManyField(VocabularyTopic, related_name='vocabulary_topics')
```
longhotsummer:
```suggestion
    taxonomies = models.ManyToManyField(VocabularyTopic, related_name='works')
```

longhotsummer:
@kaysiz looking good, some minor comments.

For the UI, we currently use https://developer.snapappointments.com/bootstrap-select/examples/ which is similar to select2, which I think will be good enough.

indigo_api/models.py
kaysiz: level_1 cannot be null and blank?

longhotsummer: Correct: level_1 is not optional. Level 2 IS optional. Neither need to be unique on their own, only when combined together using the `unique_together` index below.

longhotsummer:
Replaced by #625.

PR #587, opened 2019-05-29
[Do not merge] Available tasks view

AliciadW:
@longhotsummer please have a look and let me know what changes you still want for this. 
At the moment I have this view showing as the task list (list format not cards). Next step will be to add the tabs under `Tasks` in the blue nav bar. 

Screenshot 2019-05-29 at 12 29 08


indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
+{% load humanize indigo_app %}
+
+{% if tasks %}
+<table class="table table-striped table-bordered">
```
longhotsummer: The borders add too much noise, let's drop that and add `table-sm` to make it condensed -- we have a LOT of tasks.

indigo_app/templates/indigo_api/task_list.html
```diff
@@ -26,7 +26,7 @@
     {% if tasks %}
 
       {% if form.format.value == 'list' %}
-        {% include 'indigo_api/_task_list.html' with tasks=tasks %}
+        {% include 'indigo_api/_task_list_all.html' with tasks=tasks selectable=True %}
```
longhotsummer: Let's replace `_task_list.html` with the new code above, no need for a second template.

indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
...
+  <thead>
+    <tr>
+      <th scope="col">Task</th>
+      <th scope="col">Place</th>
+      <th scope="col">Work/Document</th>
+    </tr>
+  </thead>
+  <tbody>
+  {% for task in tasks %}
+  {% if task.state != 'pending_review' %}
```
longhotsummer: You can leave this out, we'll do the filtering in Django's view.

indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
...
+    <tr>
+      <td>
+        <div class="task-list-span">
```
longhotsummer: Bootstrap has good flex utility classes. This would be `d-flex align-items-center`

indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
...
+            {% if perms.indigo_api.change_task and task.is_open and selectable%}
+            <label class="mb-0 font-weight-normal c-pointer">
+              <input type="checkbox" name="tasks" value="{{ task.id }}" form="bulk-task-update-form" class="mr-1">
+              #{{ task.id }}
+            </label>
+            {% else %}
+            #{{ task.id }}
+            {% endif %}
+          </div>
+          <div class="text-muted {% if task.labels.all %} mr-2 {% endif %}">{{ task.title }}</div>
```
longhotsummer: This doesn't need to be muted, and should be a link to the task.

indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
...
+          {% for label in task.labels.all %}
+            <span class="badge badge-secondary" title="{{ label.description }}">{{ label.title }}</span>
```
longhotsummer: Needs some margin on the right for multiple labels not to touch.

indigo_app/templates/indigo_api/_task_list_all.html
```diff
@@ -0,0 +1,55 @@
+{% load humanize indigo_app %}
+
+{% if tasks %}
+<table class="table table-striped table-bordered">
+  <thead>
+    <tr>
+      <th scope="col">Task</th>
+      <th scope="col">Place</th>
```
longhotsummer: Let's make the place column optional. That way, the task listing in the place view can opt-out of it.

indigo_app/templates/indigo_api/_task_list_all.html
longhotsummer: Also, I suspect we'll have to do something here so that the titles line up when the numbers are different lengths (eg. task #12 vs #12345). Eg a second column? Or wrap the icon, checkbox and number in a div and use flexbox.

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: Done

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: Done

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: Removed

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: Done

indigo_app/templates/indigo_api/task_list.html
AliciadW: Done

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: I forgot about the bootstrap utility classes. 🙈 Added them.

AliciadW: Available tasks:
Screenshot 2019-05-30 at 15 12 18

Task list:
Screenshot 2019-05-30 at 15 13 24

@longhotsummer thoughts on this layout?

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: @longhotsummer I'm not sure I completely understand what you are asking here?

indigo_app/templates/indigo_api/_task_list_all.html
longhotsummer: I think it's important to keep the number, icon and title all on the same line, otherwise we lose the benefit of the columnar layout. If the title and labels are in their own column, then they'll wrap independently of the number and the icon.

```
|------------------|--------------...
| [ ] (icon) #1234 | Title label label label   |
|------------------|--------------...
```

indigo_app/templates/indigo_api/_task_list_all.html
longhotsummer: Like we do with `selectable`, only render the Place column if the calling template tells it to do so. This will mean that in the Place Task view, we can skip the Place column (because it'll all be the same place).

AliciadW: Okay, thanks. I'll update accordingly.

indigo_app/templates/indigo_api/_task_list_all.html
AliciadW: Makes sense. I'll update this.

AliciadW:
Screenshot 2019-05-31 at 08 41 00


Screenshot 2019-05-31 at 08 41 32

indigo_app/templates/place/tabbed_layout_tasks.html
```diff
@@ -0,0 +1,14 @@
+{% extends "main.html" %}
```
longhotsummer: Please move this into `templates/tasks/` rather than `templates/place/`, it's explicitly NOT place-based.

longhotsummer:
The first column still wrapped, so I've fixed that and moved the task template.

PR #591, opened 2019-05-31
Changes requested email

### LINKED ISSUE #560: As a contributor I want an email when a task gets unsubmitted and assigned back to me to make changes

goose-life:
closes #560

### DIFF
--- a/indigo_api/models.py
+++ b/indigo_api/models.py
@@ -1247,8 +1247,8 @@ class Task(models.Model):
 
     @transition(field=state, source=['pending_review'], target='open', permission=may_unsubmit)
     def unsubmit(self, user):
+        self.assigned_to = self.last_assigned_to
         action.send(user, verb=self.VERBS['unsubmit'], action_object=self, place_code=self.place.place_code)
-        self.assign_to(self.last_assigned_to, user)
 
     # close
     def may_close(self, view):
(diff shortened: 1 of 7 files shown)


indigo_api/models.py
```diff
@@ -1247,8 +1247,8 @@ def may_unsubmit(self, view):
 
     @transition(field=state, source=['pending_review'], target='open', permission=may_unsubmit)
     def unsubmit(self, user):
+        self.assigned_to = self.last_assigned_to
         action.send(user, verb=self.VERBS['unsubmit'], action_object=self, place_code=self.place.place_code)
```
longhotsummer: Let's also stash the user it has been assigned to (which could be None)

indigo_app/notifications.py
```diff
@@ -27,6 +27,13 @@ def send_task_notifications(self, action):
                     'recipient': action.target,
                 })
 
+        elif action.verb == 'requested changes to':
+            self.send_templated_email('task_changes_requested', [action.action_object.assigned_to], {
```
longhotsummer: Guard here in case assigned_to is none.

indigo_api/models.py
longhotsummer: I mean as part of the action.

PR #592, opened 2019-05-31
Auto auto assign

### LINKED ISSUE #534: As a reviewer I want to be able to easily mark a task as assign-to-me-and-all-done
I don't want to have to do the steps separately, when the task is very easy to do. I tend to open the task detail page, do the review, and mark it as done -- without self-assigning the task.

goose-life:
sortof closes #534 but discussion still under way


longhotsummer:
As discussed, we also need to ensure that a reviewer cannot approve a task that is not assigned to them.

goose-life:
closing in favour of #599

Issue #579, opened 2019-05-25
Re-implement work filtering view on the server


longhotsummer:
The work listing view currently filters works on the client, not on the server. This makes the page slow to load because we have a huge number of works.

Instead, move the filtering server-side using Django forms (much like we do filtering of tasks in the TaskListView).

The server-side view is here: https://github.com/laws-africa/indigo/blob/master/indigo_app/views/places.py#L53

The client-side view is here: https://github.com/laws-africa/indigo/blob/master/indigo_app/static/javascript/indigo/views/library.js

The current client-side is big and complex, and we can really simplify that a lot by doing it on the server.

* [x] Move filtering to server-side using a new WorkFilterForm (similar to TaskFilterForm)
* [x] Move rendering of filter components to server-side and add a submit button to the filter form.
* [x] Add pagination
* [x] Move rendering of results to server side. Currently we use Handlebars, this can be relatively easily translated to Django
* [x] No need to show counts in subtype dropdown

PR #619, opened 2019-06-06
Reimplement work filtering on the server side

ro6ley:
#### What does this PR do?
Reimplement work filtering on the server side.

The work listing view currently filters works on the client, not on the server. This makes the page slow to load because we have a huge number of works.

The solution is to move the filtering server-side using Django forms.

- Open the works page
- Use the filters
- Click on submit to receive filtered data

#### Screenshots
image

#579


ro6ley:
@longhotsummer currently the filters are on the PlaceDetailView and not via the WorkFilterForm as intended. I can't seem to get the submitted form data to the WorkFilterForm to handle the filtering, hence my form is not working. Any insights?

indigo_app/forms.py
```diff
@@ -250,6 +251,56 @@ def filter_queryset(self, queryset):
         return queryset
 
 
+class WorkFilterForm(forms.Form):
+    search_string = forms.CharField()
```
longhotsummer: Let's make this `q` please.

indigo_app/forms.py
```diff
@@ -250,6 +251,56 @@ def filter_queryset(self, queryset):
...
+    stub = forms.ChoiceField(choices=[('excl', 'excl'), ('all', 'all'), ('only', 'only')])
+    status = forms.MultipleChoiceField(choices=[('published', 'published'), ('draft', 'draft')])
+    subtype = forms.ChoiceField(choices=[('-', '-'), ('by-law', 'by-law'), ('gn', 'gn'), ('p', 'p'), ('si', 'si')])
```
longhotsummer: Subtypes can be loaded from the `indigo_api.Subtype` model. I think a ModelChoiceField will suffice.

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -0,0 +1,57 @@
+{% load account %}
+
+
+<form method="GET" id="work-filter-form">
+  <div class="d-flex">
+
+    <div class="flex-grow-1 mr-2">
+        <input type="text" class="form-control filter-search" name="search_string" placeholder="Filter by title, year or number" value="">
```
longhotsummer: Don't forget to pull in the current value from the form.

```
value="{% form.search_string.value|default: '' %}"
```

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -0,0 +1,57 @@
...
+    </div>
+
+    <!-- TODO: Types filter -->
+    <div class="mr-2">
+        <select class="form-control filter-subtype" name="subtype">
+          <option value="all">All types</option>
```
longhotsummer: These need to be dynamic and remember what has been selected, so best is to iterate over them:

```
{% for option in form.subtype %}
  {{ option }}
{% endfor %}
```

indigo_app/templates/indigo_api/_work_filter_form.html
longhotsummer: Indents should be two spaces, please :)

indigo_app/templates/indigo_api/_work_filter_form.html
longhotsummer: Similarly for the option select fields.

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -0,0 +1,57 @@
...
+        <select class="selectpicker filter-stub" name="stub" data-style="btn-outline-secondary" data-width="auto">
+          <option value="excl">Exclude stubs</option>
+          <option value="only">Only stubs</option>
+          <option value="all">Everything</option>
+        </select>
+      </div>      
+
+    <!-- Status filter -->
+    <div class="mr-2">
+      <select class="selectpicker notooltip" multiple name="status" data-style="btn-outline-secondary">
```
longhotsummer: Hardcoding the options for these selects is fine. Don't forget to render which option is currently active.

indigo_app/views/places.py
```diff
@@ -50,20 +53,40 @@ def get_context_data(self, **kwargs):
...
 
+# TODO: WorkFilterForm
 class PlaceDetailView(PlaceViewBase, AbstractAuthedIndigoView, TemplateView):
     template_name = 'place/detail.html'
     js_view = 'LibraryView'
     tab = 'works'
 
+    def get(self, request, *args, **kwargs):
+        params = QueryDict(mutable=True)
+        params.update(request.GET)
```
longhotsummer: This is where we need to set defaults, in particular:

* exclude stubs
* draft + published
* all subtypes
* sorting: most recently updated

indigo_app/forms.py
```diff
@@ -250,6 +251,56 @@ def filter_queryset(self, queryset):
...
+                queryset = queryset.filter(document__draft=True)            
+            elif self.cleaned_data['status'] == ['published']:
+                queryset = queryset.filter(document__draft=False)
+
+        if self.cleaned_data.get('sortby'):
+            queryset = queryset.order_by(self.cleaned_data.get('sortby'))        
+        
+        # filter by subtype indicated on frbr_uri
+        if self.cleaned_data.get('subtype'):
+            if self.cleaned_data['subtype'] == 'by-law':
```
longhotsummer: I suggest the below. We trust the value in cleaned_data, and adding the `/act/` will ensure it's matching in the right place.

```python
queryset = queryset.filter(frbr_uri__contains='/act/%s/' % self.cleaned_data['subtype'])
```

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -0,0 +1,52 @@
...
+        {% for option in form.subtype %}
+          {{ option }}
+        {% endfor %}
+      </select>
+    </div>    
+
+    <!-- Stubs filter -->
+    <div class="mr-2">
+      <select class="selectpicker filter-stub" name="stub" data-style="btn-outline-secondary" data-width="auto">
+        <option value="excl">Exclude stubs</option>
```
longhotsummer: Also need to remember which of these are chosen.

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -0,0 +1,52 @@
...
+      <select class="selectpicker notooltip" multiple name="status" data-style="btn-outline-secondary">
+        <option value="published" {% if 'published' in form.status.value %} selected {% endif %} data-content="<i class='fas fa-circle fa-fw published'></i> Published">Published</option>
+        <option value="draft" {% if 'draft' in form.status.value %} selected {% endif %} data-content="<i class='fas fa-circle fa-fw draft'></i> Draft">Draft</option>
+      </select>
+    </div>  
+
+    <!-- Time updated filter -->
+    <div class="mr-2">
+      <select class="form-control sortby" name="sortby">
+        <option value="-updated_at" selected>Most recently updated</option>
```
longhotsummer: Also need to remember which of these is chosen. Probably can just iterate over the options.

longhotsummer:
This is largely complete. We just need to paginate using ListView (and letting it build `context['works']`). There is a `indigo_app/_paginator.html` which will quickly add the pagination widget.

longhotsummer:
This will make a huge difference, it'll take page load times from 4-5 seconds to under a second.

PR #634, opened 2019-06-13
add commencing work details to coverpage

### LINKED ISSUE #489: as a reviewer I want to see the commencement work on the coverpage of a document
This should help at publication stage to check that everything has been recorded.

goose-life:
closes #489 
Not entirely confident on the `blocktrans`es; would a separate block be needed for 'of' on line 53?

image


longhotsummer:
We're going to need to do this a bit differently. Not all works support "X of YYYY" naming, so we need to use a plugin. See https://github.com/laws-africa/indigo/blob/master/indigo_api/models.py#L388 - be sure to guard against it being none.

longhotsummer:
Translations can be a nightmare. Ideally, the whole "Commenced on X by Y unless..." should all be one translation, because in other languages splitting up the wording may not work. So you need to line up your variables, check which variation you're using (ie. do we have a commencing work), and then use the blocktrans tag.

goose-life:
image

PR #636, opened 2019-06-18
Batch import improvements

goose-life:


indigo_app/views/works.py
```diff
@@ -629,9 +629,12 @@ def get_works(self, table, form):
                         work.save_with_revision(self.request.user)
 
                         # link publication document
+                        publication_number = work.publication_number
+                        if " " in work.publication_number:
```
longhotsummer: This feels a little dangerous to me, we'll have to see how other jurisdictions handle it.

Splitting on a string that doesn't have whitespace just returns the string, so you can safely do `publication_number = work.publication_number.split()[-1]` without having to check for whitespace.

indigo_app/views/works.py
```diff
@@ -670,123 +673,154 @@ def get_works(self, table, form):
...
+
+        try:
+            work.save_with_revision(self.request.user)
+        except ValidationError:
+            self.create_task(info, form, task_type='parent_work')
+
+    def find_work_by_title(self, title):
+        potential_match = Work.objects.filter(title=title, country=self.country, locality=self.locality)
+        if len(potential_match) == 1:
+            return potential_match
```
longhotsummer: This returns the queryset, not the actual object. Use `.filter(...).first()` to get the first matching work.

PR #638, opened 2019-06-23
Permissions badges use security groups, pull out default badges

longhotsummer:
This means it's easier to manage and understand badge permissions, and allows us to override the default set of badges.


goose-life:
see https://github.com/laws-africa/indigo-lawsafrica/pull/42

PR #650, opened 2019-06-26
Points plugin

ro6ley:
#### What does this PR do?
As a user I want to earn points when I complete tasks

This PR adds the following:
* [x] mechanism to associate points with tasks (`extra_data` JSONField)


indigo_api/models.py
```diff
@@ -1168,6 +1170,8 @@ class Meta:
 
     labels = models.ManyToManyField('TaskLabel', related_name='+')
 
+    extra_data = JSONField(default={'points': 12})
```
longhotsummer: This indigo package shouldn't need to know about the points at all. This can just be an empty JSONField. (Same applies to the migration).

ro6ley:
@longhotsummer I have modified the `extra_data JSONField`

PR #651, opened 2019-06-27
Bulk creation improvements

goose-life:
- Moves creation of works from spreadsheet into a plugin, which allows for locale-based customisation
- Slight front-end improvements (preview etc to be done separately)
- Moves linking of publication documents to after work creation (to be with other links / tasks)


indigo_app/views/works.py
```diff
@@ -556,131 +558,58 @@ def get_context_data(self, **kwargs):
...
+                self.link_publication_document(info, form)
                 if info.get('commenced_by'):
                     self.link_commencement(info, form)
                 if info.get('repealed_by'):
                     self.link_repeal(info, form)
                 if info.get('parent_work'):
                     self.link_parent_work(info, form)
 
-            if (info['status'] == 'success' or info['status'] == 'duplicate') and info.get('amends'):
+            if not info['status'] == 'error' and info.get('amends'):
```
longhotsummer:
```suggestion
            if info['status'] != 'error' and info.get('amends'):
```

indigo_app/views/works.py
```diff
@@ -556,131 +558,58 @@ def get_context_data(self, **kwargs):
...
         works = None
+        locality_code = self.locality.code if self.locality else None
+        bulk_creator = plugins.for_locale('bulk-creator', self.country.code, None, locality_code)
 
         try:
             table = self.get_table(form.cleaned_data.get('spreadsheet_url'))
-            works = self.get_works(table, form)
-            self.link_works(works, form)
+            works = bulk_creator.get_works(self, table)
+            self.create_links(works, form)
```
longhotsummer: I suspect it'll be useful to have `create_links` and `get_tasks` both exist on the plugin. Time will tell.

indigo/bulk_creator.py
```diff
@@ -0,0 +1,203 @@
...
+        available_subtypes = [s.abbreviation for s in Subtype.objects.all()]
+
+        row_data = row
+        row_data['country'] = view.country.code
+        row_data['locality'] = view.locality.code if view.locality else None
+        form = self.get_row_validation_form(row_data)
+
+        # Extra validation
+        # - if the subtype hasn't been registered
+        if row_subtype and row_subtype not in available_subtypes:
```
longhotsummer: The form could do this for you if you use a `clean_subtype` methods (put your logic in that method on the form, it'll be called automatically), or use a ChoiceField and setup the choices in the form's `__init__` method.

Ditto for the other fields -- moving the validation logic entirely into the form makes it simpler to adjust/override.

But none of this is crucial, we can change it later.

PR #661, opened 2019-07-02
Better task emails

longhotsummer:
* task status and assignment in emails
* wording for 'review' assignments

email html 2019-07-02 16-51-24


indigo_api/models.py
```diff
@@ -1201,6 +1201,7 @@ def assign_to(self, assignee, assigned_by):
         """ Assign this task to assignee (may be None)
         """
         self.assigned_to = assignee
+        self.save()
         if assigned_by == self.assigned_to:
```
goose-life: should https://github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L301 move here 
and https://github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L302 come out?

indigo_api/models.py
```diff
@@ -1262,6 +1263,7 @@ def submit(self, user):
             self.assign_to(user, user)
         self.last_assigned_to = self.assigned_to
         self.assigned_to = None
+        self.save()
```
goose-life: should https://github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L245 move here
and https://github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L266 come out?

goose-life: ditto for updated_by_user below × 4

longhotsummer:
I don't normally like arbitrary `.save()` calls -- these are added because the signal handlers for `post_save` on the `action` signal need to have the updated task stored in the DB. As a new rule of thumb, if we create an action object, the related objects need to be saved first.

indigo_api/models.py
longhotsummer: Hmm. I think L266 (the `.save()`) can be removed.

I'm happy if the `updated_by_user` stays in the view. This method doesn't know which user is doing the updating, necessarily.

longhotsummer:
@goose-life updated

goose-life:
should by the same token github.com/laws-africa/indigo/blob/master/indigo_app/views/tasks.py#L284-L304 change so that L301 moves to L287 and L302 comes out?

longhotsummer:
@goose-life updated, the saving and actions now happens in the post-transition signal handler. I've added the save back in the task view to ensure it happens.

PR #667, opened 2019-07-04
Indigo points ui

AliciadW:
Added UI for points badges.

Closes https://github.com/laws-africa/indigo-points/issues/8

See also: https://github.com/laws-africa/indigo-points/pull/10


longhotsummer:
@AliciadW thanks. I've made some small changes so that we have blocks where we can inject the points stuff from indigo-points, to keep this entirely unaware of it all.

PR #670, opened 2019-07-05
Task reassigned to reviewer when resubmitted

### LINKED ISSUE #632: reassign task to reviewer when resubmitted after changes requested
add a `reviewed_by_user` to the Task model, set that to whoever `unsubmit`s a task, and assign it to them when the task is resubmitted (we can check whenever a task is submitted whether it's got `changes_requested`, and send an email to the `reviewed_by_user` saying something like `User Y has resubmitted the following task to which you requested changes: …`?

goose-life:
closes #632


indigo_api/models.py
```diff
@@ -1263,8 +1263,8 @@ def may_submit(self, view):
     def submit(self, user):
         if not self.assigned_to:
             self.assign_to(user, user)
-        self.last_assigned_to = self.assigned_to
-        self.assigned_to = None
+        self.submitted_by_user = self.assigned_to
+        self.assigned_to = self.reviewed_by_user
```
longhotsummer: Shouldn't this be `self.assign_to` so that the assignment is stored as an action and emails are sent?

longhotsummer: Oh, I see that there's a new email for `submit` with a reviewer. I'm not against having  both emails sent: assigned to you, and the task was submitted back to you for review.

goose-life: I'm not crazy about the idea of two emails – I think of it as the converse of 'changes requested', where only one email is sent to the submitted_by_user. Happy to use the assignment instead, the wording of the email was just a bit trickier (I tried that first).

longhotsummer: Fair point, it mirrors `unsubmit`. Let's do it.

PR #671, opened 2019-07-11
include provision for publication_date_optional in work form

goose-life:


indigo_app/views/works.py
```diff
@@ -224,6 +225,8 @@ def get_context_data(self, **kwargs):
             'locality': self.locality.code if self.locality else None,
         })
         context['subtypes'] = Subtype.objects.order_by('name').all()
+        if self.country.code in self.PUB_DATE_OPTIONAL_COUNTRIES:
```
longhotsummer: `context['publication_date_optional'] = self.country.code in self.PUB_DATE_OPTIONAL_COUNTRIES`

If you're using an `if` to set a boolean variable, you can drop the if and use the condition to set the variable.

PR #677, opened 2019-07-16
Revert "Work and document are readonly after task is created"

goose-life:
This reverts commit 42da0c16cc43e1ebd64a01fc63d24316c3fdc6e2.

### DIFF
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -209,7 +209,7 @@ class ImportDocumentForm(forms.Form):
             raise forms.ValidationError("Invalid json data")
 
 
-class TaskCreateForm(forms.ModelForm):
+class TaskForm(forms.ModelForm):
     class Meta:
         model = Task
         fields = ('title', 'description', 'work', 'document', 'labels', 'workflows')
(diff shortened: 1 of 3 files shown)


indigo_app/forms.py
```diff
@@ -218,12 +218,7 @@ class Meta:
                                             required=False)
     workflows = forms.ModelMultipleChoiceField(queryset=Workflow.objects, widget=forms.CheckboxSelectMultiple,
                                                required=False)
-
-
-class TaskEditForm(TaskCreateForm):
-    class Meta:
-        model = Task
-        fields = ('title', 'description', 'labels', 'workflows')
+    assigned_to = forms.ModelChoiceField(queryset=User.objects, empty_label='Unassigned', required=False)
```
longhotsummer: Actually, I don't think this should be here.

goose-life: you're totally right! we moved that to the detail view in the meantime, thanks for catching

goose-life:
```suggestion
```

PR #679, opened 2019-07-16
Verification notice

longhotsummer:
Samplex @ 2004-05-21 2019-07-16 13-37-19


goose-life:
Looks good to me; is there a way to check that those changes were made in all the places they're meant to be? did it happen automatically when you made the change in one place? (why so many places to start?)

longhotsummer:
The translation files are updated automatically, so it should cover everything.

PR #686, opened 2019-07-18
additional properties

goose-life:
specifically, allow cap 2000 and cap 2004 to be added in Nigeria (see https://github.com/laws-africa/indigo-lawsafrica/pull/48)

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -29,7 +29,6 @@ class RowValidationFormBase(forms.Form):
     year = forms.CharField(validators=[
         RegexValidator(r'\d{4}', 'Must be a year (yyyy).')
     ])
-    cap_number = forms.CharField(required=False)
     publication_name = forms.CharField(required=False)
     publication_number = forms.CharField(required=False)
     publication_date = forms.DateField(error_messages={'invalid': 'Date format should be yyyy-mm-dd.'})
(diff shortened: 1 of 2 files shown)


indigo/bulk_creator.py
```diff
@@ -29,6 +29,7 @@ class RowValidationFormBase(forms.Form):
     year = forms.CharField(validators=[
         RegexValidator(r'\d{4}', 'Must be a year (yyyy).')
     ])
+    cap = forms.CharField(required=False)
```
longhotsummer: ?

indigo_app/views/works.py
```diff
@@ -566,11 +566,15 @@ def form_valid(self, form):
         works = None
         locality_code = self.locality.code if self.locality else None
         bulk_creator = plugins.for_locale('bulk-creator', self.country.code, None, locality_code)
+        extra_properties_getter = plugins.for_locale('extra-properties', self.country.code, None, locality_code)
```
longhotsummer: What if we simplify this and simply have a list of extra properties on the BulkCreator object? We already have plugin support for that.

indigo_app/views/works.py
```diff
@@ -594,6 +598,14 @@ def create_links(self, works_info, form):
                 # (they won't overwrite the existing works but the amendments will be linked)
                 self.link_amendment(info, form)
 
+    def add_extra_properties(self, works_info, extra_properties):
```
longhotsummer: This could then move into the bulk_creator, which makes more sense to me: that's the object which transforms rows into Work objects.

indigo_app/views/works.py
goose-life: wonderful idea!

PR #693, opened 2019-07-22
Google sheets integration

### LINKED ISSUE #644: Authenticate when talking with google docs
Use the Python Google Services client library to authenticate and talk to the Google Sheets API directly, rather than using the public CSV view.

longhotsummer:
Partially fixes #644.

* if google sheets is setup (`settings.INDIGO['GSHEETS_API_CREDS']`) then try to load the list of spreadsheet tabs
* when doing import, fetch via API rather than CSV
* fall back to CSV if gsheets is not enabled
* the spreadsheet must be shared with the email address of the Google API Creds Service Account.

Still pending:
* actually test fetching data table from google sheets


indigo_app/templates/indigo_api/work_new_batch.html
```diff
@@ -104,14 +129,13 @@ <h2 class="mt-3 mb-3">Create a new batch of works</h2>
...
-        {% if place_workflows %}
+        {% if form.fields.workflow.choices %}
           <div class="form-group">
             <div class="row">
               <div class="col-md-6">
-                <select class="selectpicker form-control" required name="{{ form.workflow.name }}" id="{{ form.workflow.id_for_label }}" data-live-search="true">
-                  <option>Choose one</option>
-                  {% for workflow in place_workflows %}
-                    <option value="{{ workflow.id }}" {% if workflow.id in form.workflow.value %}selected{% endif %}>{{ workflow.title }}</option>
+                <select class="selectpicker form-control" required name="{{ form.workflow.name }}" id="{{ form.workflow.id_for_label }}" data-live-search="true" placeholder="Choose a workflow">
```
goose-life: `placeholder` doesn't work

longhotsummer: Oh it's selectpicker. Try `data-liveSearchPlaceholder="foo"`?

goose-life: this gives a placeholder in the search box, but the first workflow is still selected by default, which we don't want
image

longhotsummer: Ah, use `title`:

https://developer.snapappointments.com/bootstrap-select/examples/ "Using the title attribute will set the default placeholder text when nothing is selected. This works for both multiple and standard select boxes:
"

goose-life: I suspect we may have to do something to the form for this to take; just replacing `placeholder` with `title` didn't do anything – the first workflow is still selected by default :/

goose-life:
see https://github.com/laws-africa/indigo/pull/699 for final item in list

PR #697, opened 2019-07-24
include current tasks on work overview

### LINKED ISSUE #682: As a contributor I want to see a list/overview of open tasks for a work on the work overview page

goose-life:
closes #682

image

### DIFF
--- a/indigo_app/templates/indigo_api/_task_list.html
+++ b/indigo_app/templates/indigo_api/_task_list.html
@@ -6,13 +6,13 @@
     <tr>
       {% block table-header %}
       <th scope="col" colspan="2">Task</th>
-      {% if not work_overview %}
-        {% if place %}
-          <th scope="col">Place</th>
-        {% endif %}
-        <th scope="col">Work/Document</th>
-      {% else %}
+      {% if place %}
+        <th scope="col">Place</th>
+      {% endif %}
+      {% if hide_works %}
         <th scope="col">Document</th>
+      {% else %}
+        <th scope="col">Work or Document</th>
       {% endif %}
       <th scope="col">Updated</th>
       {% endblock %}
@@ -52,7 +52,7 @@
       {% endblock %}
     </td>
 
-    {% if not work_overview and place %}
+    {% if place %}
       <td>
         {% block col-place %}
         {{ task.place }}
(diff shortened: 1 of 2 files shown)


indigo_app/templates/indigo_api/_task_list.html
```diff
@@ -6,10 +6,14 @@
     <tr>
       {% block table-header %}
       <th scope="col" colspan="2">Task</th>
-      {% if place %}
-        <th scope="col">Place</th>
+      {% if not work_overview %}
```
longhotsummer: This breaks the concept of encapsulation, that this template should only need to know about its local state. `work_overview` implies it has broader knowledge, that it understands that when it's being used in a particular context, then certain things should happen differently.

The previous `place` indicator inverts this: the including context is telling the template to behave in a certain way. This preserves encapsulation.

Instead, I suggest you keep `place` and either just leave 'Work/Document' as it is (actually I think "Work or Document" is better), or add another variable to control the display of that column.

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -190,4 +190,11 @@ <h3 class="mb-0">{{ work.metrics.p_breadth_complete|floatformat:0 }}%</h3>
   {% include 'indigo_api/_work_points_in_time.html' with work=work editable=False %}
   {% endblock %}
 
+  {% block active_tasks %}
+  <div class="h4 mb-3 mt-5">Current tasks</div>
```
longhotsummer:
```suggestion
  <div class="h4 mb-3 mt-5">Open tasks</div>
```

indigo_app/templates/indigo_api/_task_list.html
goose-life: I would like to not display the work when on the work overview, it feels silly to include it.
When you say 'add another variable to control the display of that column', would it be fine to just call it something else, like `dont_display_works`? I don't want to make it `display_works` for all other instances, the default should be to display them.

goose-life: something like
```
  {% if hide_works %}
    <th scope="col">Document</th>
  {% else %}
    <th scope="col">Work or Document</th>
  {% endif %}
```

longhotsummer: Yes, absolutely. It's exactly the same concept as 'show/don't show the place'.

PR #703, opened 2019-07-28
User profile

kaysiz:
- User able to upload profile photo
- contributors able to view photos of other users

Screenshot 2019-07-28 at 07 47 03


kaysiz:
@goose-life should i add a default image value to the model?

goose-life:
@kaysiz looks good so far, thanks!
If the photo will always be a certain shape (which I think is the right call), we should probably add a cropbox on upload (otherwise it's going to stretch).
It would also be good if it showed up in the UserProfileEditView / if it was obvious that the upload was successful.

Let me know if you have questions about either of these?

PR #704, opened 2019-07-29
Email notification for new reply to an annotation

### LINKED ISSUE #568: When someone replies to an annotation on a document, email me.
An annotation is a bit like a comment, but it uses a different model because it is bound to a particular point in a document.

When an annotation object is created *and* it is a reply to another annotation (`in_reply_to` is not None), send a notification to:

* the user who created the parent annotation (`in_reply_to`)
* all users who also commented on the parent annotation

(unless the user is the user who commented, of course).

This could be done by watching the `post_save` signal on Annotation, much like we do with Comments. See https://github.com/laws-africa/indigo/blob/master/indigo_app/signals.py#L25 and https://github.com/laws-africa/indigo/blob/master/indigo_app/notifications.py#L110

ro6ley:
#### What does this PR do?
Notify users by email when a new reply to an annotation is posted.

When a user replies to an annotation in a document, an email is sent to the user who posted the first annotation, and all other users who have also replied to the parent annotation. The user posting the reply is excluded from the email list.

- Set up Indigo
- Create a few users and documents
- Add an annotation on the document and reply to it with the other user accounts

closes #568

### DIFF
--- a/indigo_app/notifications.py
+++ b/indigo_app/notifications.py
@@ -3,7 +3,6 @@ import logging
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django_comments.models import Comment
-from django.db.models.signals import post_save
 from django.dispatch import receiver
 from templated_email import send_templated_mail
 from background_task import background
@@ -78,11 +77,12 @@ class Notifier(object):
                     'comment': comment,
                 })
 
-    def notify_reply_to_annotation(self, parent_annotation, annotation):
+    def notify_reply_to_annotation(self, annotation):
         """ Send email notifications when there is a reply to an annotation. This
         email is sent to annotation creator and all users who replied, excluding
         the currently replying user.
         """
+        parent_annotation = annotation.in_reply_to
         related_annotations = Annotation.objects.filter(in_reply_to=parent_annotation)
         document = parent_annotation.document
 
@@ -140,13 +140,16 @@ def notify_comment_posted(comment_id):
         log.warning("Comment with id {} doesn't exist, ignoring".format(comment_id))
 
 
-@receiver(post_save, sender=Annotation)
-def post_annotation_reply(sender, **kwargs):
-    if kwargs['instance'].in_reply_to:
-        notifier.notify_reply_to_annotation(kwargs['instance'].in_reply_to, kwargs['instance'])
+@background(queue='indigo')
+def notify_annotation_reply_posted(annotation_id):
+    try:
+        notifier.notify_reply_to_annotation(Annotation.objects.get(pk=annotation_id))
+    except Comment.DoesNotExist:
+        log.warning("Annotation with id {} doesn't exist, ignoring".format(annotation_id))
 
 
 if not settings.INDIGO.get('NOTIFICATION_EMAILS_BACKGROUND', False):
     # change background notification tasks to be synchronous
     notify_task_action = notify_task_action.now
     notify_comment_posted = notify_comment_posted.now
+    notify_annotation_reply_posted = notify_annotation_reply_posted.now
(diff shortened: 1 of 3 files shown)


indigo_app/notifications.py
```diff
@@ -115,6 +140,12 @@ def notify_comment_posted(comment_id):
         log.warning("Comment with id {} doesn't exist, ignoring".format(comment_id))
 
 
+@receiver(post_save, sender=Annotation)
+def post_annotation_reply(sender, **kwargs):
+    if kwargs['instance'].in_reply_to:
+        notifier.notify_reply_to_annotation(kwargs['instance'].in_reply_to, kwargs['instance'])
```
longhotsummer: Minor comment: I think it makes more sense for `notify_reply_to_annotation` to simply receive the entire annotation object. That has all the context, including the parent. It can then decide what to do. In other words, passing in the parent AND the annotation is unnecessary.

indigo_app/notifications.py
```diff
@@ -115,6 +140,12 @@ def notify_comment_posted(comment_id):
         log.warning("Comment with id {} doesn't exist, ignoring".format(comment_id))
 
 
+@receiver(post_save, sender=Annotation)
+def post_annotation_reply(sender, **kwargs):
```
longhotsummer: I've made some changes to the notification signals, so that they can be sent by a background thread. Otherwise, if the SMTP service (eg. mailgun) is having problems, it will cause the action (eg. creating an annotation) to fail.

1. please move the signal receiver into `signals.py`
2. add a new function here which is setup as a background task: it should take the annotation id, load the object from the DB, and then pass that to the notifier method.

See the `notify_comment_posted` method as an example.

indigo_app/notifications.py
```diff
@@ -76,6 +78,29 @@ def notify_comment_posted(self, comment):
                     'comment': comment,
                 })
 
+    def notify_reply_to_annotation(self, parent_annotation, annotation):
+        """ Send email notifications when there is a reply to an annotation. This
+        email is sent to annotation creator and all users who replied, excluding
+        the currently replying user.
+        """
+        related_annotations = Annotation.objects.filter(in_reply_to_id=parent_annotation.id)
```
longhotsummer: When querying related objects, the Django way is to reference the foreign key field, not the foreign key id: `...filter(in_reply_to=parent_annotation)` - Django prefers you not to have to worry about primary keys.

indigo_app/templates/templated_email/annotation_new_reply.email
```diff
@@ -0,0 +1,16 @@
+{% block subject %}Document #{{ document.pk }} – {{ document.title|safe }}{% endblock %}
+
+{% block html %}
+  {% include 'templated_email/_header.html' %}
+
+  <p> {{ annotation.created_by_user.username }} replied to a thread on
+    <a href="{{ SITE_URL }}{% url 'document' doc_id=document.pk %}"> Document #{{ document.pk }} – {{ document.title }} </a>:
```
longhotsummer:
```suggestion
    <a href="{{ SITE_URL }}{% url 'document' doc_id=document.pk %}?anntn={{ annotation.in_reply_to.pk }}">Document #{{ document.pk }} – {{ document.title }}</a>:
```

indigo_app/templates/templated_email/annotation_new_reply.email
```diff
@@ -0,0 +1,16 @@
+{% block subject %}Document #{{ document.pk }} – {{ document.title|safe }}{% endblock %}
+
+{% block html %}
+  {% include 'templated_email/_header.html' %}
+
+  <p> {{ annotation.created_by_user.username }} replied to a thread on
```
longhotsummer:
```suggestion
  <p> {{ annotation.created_by_user.username }} replied to a comment on
```

indigo_app/notifications.py
ro6ley: Yes, it does. Rectifying this

PR #708, opened 2019-08-06
Task Comments in Linked Annotations

### LINKED ISSUE #628: As a user viewing a document, I want to see task comments in linked annotations
In the document view, if an annotation is linked to a task, show the comments from that task as though they were annotation replies.

* document annotations are loaded asynchronously from the api.
* on the server, when loading annotations, check if the root annotation has a task, and if so, also load and return the comments as "fake" reply annotations (don't include an ID)
* in the view, be sure not to allow editing or removal of annotations without an ID

ro6ley:
#### What does this PR do?
As a user viewing a document, I want to see task comments in linked annotations

Annotations can be linked to tasks and annotations can have replies. Linked tasks have comments and this PR pulls the comments from linked tasks into the annotation thread.

- Create a document and create an annotation
- Create a task and link it to the annotation
- Post comments on the task and view them as replies to the annotation

#### Screenshots
Task comments are the ones without dropdowns:
image

closes #628

### DIFF
--- a/indigo_api/views/documents.py
+++ b/indigo_api/views/documents.py
@@ -4,9 +4,11 @@ from actstream import action
 
 from django.views.decorators.cache import cache_control
 from django.db.models import F
+from django.contrib.contenttypes.models import ContentType
 from django.contrib.postgres.search import SearchQuery
 from django.http import Http404
 from django.urls import reverse
+from django_comments.models import Comment
 
 from rest_framework.exceptions import ValidationError, MethodNotAllowed
 from rest_framework.views import APIView
@@ -25,7 +27,7 @@ from lxml.etree import LxmlError
 
 from indigo.analysis.differ import AttributeDiffer
 from indigo.plugins import plugins
-from ..models import Document, Annotation, DocumentActivity
+from ..models import Document, Annotation, DocumentActivity, Task
 from ..serializers import DocumentSerializer, RenderSerializer, ParseSerializer, DocumentAPISerializer, VersionSerializer, AnnotationSerializer, DocumentActivitySerializer, TaskSerializer
 from ..renderers import AkomaNtosoRenderer, PDFResponseRenderer, EPUBResponseRenderer, HTMLResponseRenderer, ZIPResponseRenderer, HTMLRenderer
 from ..authz import DocumentPermissions, AnnotationPermissions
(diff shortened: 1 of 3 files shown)


indigo_api/views/documents.py
```diff
@@ -150,7 +152,33 @@ class AnnotationViewSet(DocumentResourceView, viewsets.ModelViewSet):
     permission_classes = DEFAULT_PERMS + (DjangoModelPermissionsOrAnonReadOnly, AnnotationPermissions)
 
     def filter_queryset(self, queryset):
-        return queryset.filter(document=self.document).all()
+        queryset = list(queryset.filter(document=self.document).all())
+
+        fake_annotations = []
+        for annotation in queryset:
+            if annotation.task and annotation.in_reply_to is None:
+                task_content_type = ContentType.objects.get_for_model(Task)
```
longhotsummer: Rather do this once, outside of the for loop -- it's not going to during iterations.

indigo_app/templates/document/_annotations.html
```diff
@@ -2,25 +2,30 @@
...
-    <a href="#" class="dropdown-toggle btn btn-sm" data-toggle="dropdown"></a>
-    <div class="dropdown-menu">
-      {{#unless in_reply_to}}
-        {{#unless task}}
-          {{#if permissions.can_create_task}}
-            <a href="#" class="dropdown-item create-anntn-task">Create task</a>
+
+    {{#if id}}
+      <a href="#" class="dropdown-toggle btn btn-sm" data-toggle="dropdown"></a>
+
```
longhotsummer: Pedantic: please remove this blank line, the dropdown toggle and its dropdown menu are logically one unit.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -8,7 +8,7 @@
    */
   Indigo.AnnotationView = Backbone.View.extend({
     className: function() {
-      return 'annotation ' + (!this.model.get('in_reply_to') ? 'root' : 'reply');
+      return 'annotation ' + (!this.model.get('in_reply_to') ? 'root' : 'reply_id');
```
longhotsummer: This seems odd to me? Does this really need to change?

indigo_app/templates/document/_annotations.html
```diff
@@ -2,25 +2,30 @@
...
 {{#unless permissions.readonly }}
   <div class="float-right">
-    <a href="#" class="dropdown-toggle btn btn-sm" data-toggle="dropdown"></a>
-    <div class="dropdown-menu">
-      {{#unless in_reply_to}}
-        {{#unless task}}
-          {{#if permissions.can_create_task}}
-            <a href="#" class="dropdown-item create-anntn-task">Create task</a>
+
+    {{#if id}}
```
longhotsummer: I don't think you should have to guard for having an id -- "fake" annotations are always replies, and so should never be top-level?

indigo_app/templates/document/_annotations.html
```diff
@@ -2,25 +2,30 @@
...
+            {{/if}}
+          {{/unless}}
+          {{#if permissions.can_change }}
+            <a href="#" class="dropdown-item close-anntn">Resolve</a>
+            <div class="dropdown-divider"></div>
           {{/if}}
         {{/unless}}
+
         {{#if permissions.can_change }}
-          <a href="#" class="dropdown-item close-anntn">Resolve</a>
```
longhotsummer: It looks like we've lost the Resolve item?

Off the top of my head, I don't think this template needs to change at all?

indigo_app/static/javascript/indigo/views/annotations.js
ro6ley: I forgot to revert this change, at some point fake annotations manifested on the front-end as empty annotation forms with the empty text field and the buttons, I was trying to isolate the issue.

I am reverting it.

indigo_app/templates/document/_annotations.html
ro6ley: It was not lost, just moved in the snippet. Also, the template needs to change to disable `Edit` and `Delete` options for fake annotations.

indigo_app/templates/document/_annotations.html
ro6ley: I put the id check to disable the entire drop-down to not only eliminate `Edit` and `Delete` options, but also visually distinguish fake annotations from real ones by the lack of the dropdown as you can see on the screenshot on the PR. However, I can revert to just disabling `Edit` and `Delete` operations in the dropdown of fake annotations

indigo_app/templates/document/_annotations.html
longhotsummer: cool

indigo_api/views/documents.py
```diff
@@ -151,6 +153,42 @@ class AnnotationViewSet(DocumentResourceView, viewsets.ModelViewSet):
 
     def filter_queryset(self, queryset):
         return queryset.filter(document=self.document).all()
+    
+    def list(self, request, **kwargs):
```
ro6ley: @longhotsummer here is the overriden `list()` method.

longhotsummer: looks good.

PR #715, opened 2019-08-14
Move bulk import logic into bulk creator

longhotsummer:
This pulls all the business logic into one place, rather than splitting it across the view and the bulk creator plugin.

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -193,7 +193,7 @@ class BaseBulkCreator(LocaleBasedMatcher):
 
     def create_work(self, view, row, idx, dry_run):
         # copy all row details
-        info = {k: v for k, v in row.iteritems()}
+        info = row
         info['row'] = idx + 2
 
         row = self.validate_row(view, row)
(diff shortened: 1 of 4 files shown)


indigo_app/static/javascript/indigo/views/work.js
```diff
@@ -346,15 +346,15 @@
     },
   });
 
-  Indigo.BulkImportWorksView = Backbone.View.extend({
+  Indigo.BatchAddWorkView = Backbone.View.extend({
```
goose-life: not sure why but I'm no longer seeing a progress bar

indigo/bulk_creator.py
```diff
@@ -167,77 +170,101 @@ def get_works(self, view, table, dry_run):
...
+                if info['status'] != 'error' and info.get('amends'):
+                    # this will check duplicate works as well
+                    # (they won't overwrite the existing works but the amendments will be linked)
+                    self.link_amendment(info['work'], info)
+
+        return works
+
+    def create_work(self, view, row, idx, dry_run):
+        # copy all row details
+        info = {k: v for k, v in row.iteritems()}
```
goose-life:
```suggestion
        info = row
```

longhotsummer: Heheh.

indigo_app/static/javascript/indigo/views/work.js
longhotsummer: hard refresh?

PR #719, opened 2019-08-16
Load Annotation Replies as Comments to the Associated Task

### LINKED ISSUE #627: As a user I want to see annotation replies in annotation-related task timelines
For tasks that are linked to a document annotation, also pull in annotation replies into the task timeline, as though they were comments. They can be shown exactly as if they were comments.

ro6ley:
#### What does this PR do?
As a user viewing a task, I want to see annotation replies as comments in the task timeline

Annotations can be linked to Tasks and annotations can have replies. This PR pulls a linked annotation's replies into the Task's timelines as if they were comments on the task.

- Create a document and create an annotation
- Create a task and link it to the annotation
- Post replies to the annotation and view them as comments to the task

#### Screenshots
Annotation reply showing up as a comment:
image

closes #627

### DIFF
--- a/indigo_app/views/tasks.py
+++ b/indigo_app/views/tasks.py
@@ -105,23 +105,21 @@ class TaskDetailView(TaskViewBase, DetailView):
             .select_related('user'))
         
         # get the annotation for the particular task
-        task_annotation = Annotation.objects.filter(task=task).first()
+        try:
+            task_annotation = task.annotation
+        except Annotation.DoesNotExist:
+            task_annotation = None
 
         # for the annotation that is linked to the task, get all the replies
         if task_annotation:
-            fake_comments = []
             # get the replies to the annotation
             annotation_replies = Annotation.objects.filter(in_reply_to=task_annotation)\
                     .select_related('created_by_user')
-            for annotation in annotation_replies:
-                fake_comment = Comment(
-                    user=annotation.created_by_user,
-                    comment=annotation.text,
-                    submit_date=annotation.created_at
-                )
-                fake_comments.append(fake_comment)
 
-            comments += fake_comments
+            comments.extend([Comment(user=a.created_by_user,
+                                     comment=a.text,
+                                     submit_date=a.created_at)
+                            for a in annotation_replies])
 
         context['task_timeline'] = sorted(
             chain(comments, actions),
--- a/indigo_app/views/tasks.py
+++ b/indigo_app/views/tasks.py
@@ -111,12 +111,12 @@ class TaskDetailView(TaskViewBase, DetailView):
         if task_annotation:
             fake_comments = []
             # get the replies to the annotation
-            annotation_replies = Annotation.objects.filter(in_reply_to=task_annotation)
+            annotation_replies = Annotation.objects.filter(in_reply_to=task_annotation)\
+                    .select_related('created_by_user')
             for annotation in annotation_replies:
                 fake_comment = Comment(
                     user=annotation.created_by_user,
                     comment=annotation.text,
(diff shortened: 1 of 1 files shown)


indigo_app/views/tasks.py
```diff
@@ -100,9 +100,29 @@ def get_context_data(self, **kwargs):
...
         actions = task.action_object_actions.all()
         task_content_type = ContentType.objects.get_for_model(self.model)
-        comments = Comment.objects\
+        comments = list(Comment.objects\
             .filter(content_type=task_content_type, object_pk=task.id)\
-            .select_related('user')
+            .select_related('user'))
+        
+        # get the annotation for the particular task
+        task_annotation = Annotation.objects.filter(task=task).first()
```
longhotsummer: This is a 1-1 relationship (https://github.com/laws-africa/indigo/blob/master/indigo_api/models.py#L1078) so you can simply use `task.annotation`, but you do need to guard against that throwing an `Annotation.DoesNotExist` exception.

indigo_app/views/tasks.py
```diff
@@ -100,9 +100,29 @@ def get_context_data(self, **kwargs):
...
+
+        # for the annotation that is linked to the task, get all the replies
+        if task_annotation:
+            fake_comments = []
+            # get the replies to the annotation
+            annotation_replies = Annotation.objects.filter(in_reply_to=task_annotation)
+            for annotation in annotation_replies:
```
longhotsummer: There's a simpler way to do this. You're building up a list one by one, then appending that list to another list. Instead, you can just append the items one by one to the other list. Even more pythonic, is this:

```python
comments.extend([Comment(
    user=a.created_by_user,
    comment=a.text,
    # etc.
) for a in annotation_replies])
```

indigo_app/views/tasks.py
```diff
@@ -100,9 +100,29 @@ def get_context_data(self, **kwargs):
...
+            .select_related('user'))
+        
+        # get the annotation for the particular task
+        task_annotation = Annotation.objects.filter(task=task).first()
+
+        # for the annotation that is linked to the task, get all the replies
+        if task_annotation:
+            fake_comments = []
+            # get the replies to the annotation
+            annotation_replies = Annotation.objects.filter(in_reply_to=task_annotation)
```
longhotsummer: You also need the user object, so be sure to use `select_related('user')` otherwise it will go to the database each time you access the user on the reply object.

indigo_app/views/tasks.py
```diff
@@ -100,9 +100,29 @@ def get_context_data(self, **kwargs):
...
+            for annotation in annotation_replies:
+                fake_comment = Comment(
+                    user=annotation.created_by_user,
+                    comment=annotation.text,
+                    object_pk=task_annotation.task,
```
longhotsummer: Is this line correct? Does the fake annotation really need an `object_pk` value and, if so, shouldn't it be an id field such as `task_annotation.task.pk`?

ro6ley: I thought it is needed while rendering on the front end, it actually is not. Taking it off.

indigo_app/views/tasks.py
ro6ley: I usually go for readability first 😅but yes, that's a more Pythonic way of doing it.

indigo_app/views/tasks.py
ro6ley: Yes, I have to guard against that, I had initially done an `Annotation.objects.get()` but I realized I have to guard against that exception too

longhotsummer:
woot!

PR #723, opened 2019-08-20
Fix new annotation bug

### LINKED ISSUE #722: Adding a new annotation makes it look like it's a reply
Bug: when adding a new annotation to a document, it is rendered as though it is a reply to a non-existent root annotation.

To reproduce:

1. open any document
2. add a new annotation

What happens:

National Symbols of the Republic of Namibia Act, 2018 @ 2018-12-28 2019-08-19 20-41-09

What should happen: it should not be rendered as a reply to another annotation.

ro6ley:
#### What does this PR do?
Adding a new annotation makes it look like it's a reply

Adding a new annotation to a document appears as if it is a reply to an empty annotation as can be seen below:
image

This PR fixes that by using Backbone's autogenerated but unique `cid` as the `id` for task comments or fake annotations. See: https://devdocs.io/backbone/index#Model-cid

- Create a new document
- Add a new annotation
- Create a task and post comments on the task (optional)

#### Screenshots
After fixing the bug:
image

New annotations appear as they should.

#### Relevant Issue
closes #722


longhotsummer:
Nice, thanks!

PR #730, opened 2019-08-26
Badges and icons

AliciadW:
@longhotsummer I started with this in `indigo-points`, but have moved it into `indigo` because it isn't dependent on `indigo-points`. Please let me know if I should move it back to `indigo-points` rather. 
I tried to see where the level badges come from, but can't seem to find it (from what I understand and can see `badges.py` include the permissions badges and country badges). I would like to have the levels as numbers (e.g. `badge.level`) to add this to the icon as well as to change the colours of the level badges.

Screenshot 2019-08-26 at 20 28 15


AliciadW:
Also, I think it would be better at this stage to replace the `Badges earned` block with the `Achievements` block instead of duplicating it.

longhotsummer:
Hi Alicia. Yes, this is the right place to do this.

Yes, let's replace the old "badges earned" block.

We need a generic way of rendering badges. The badge object should have sufficient information on it to allow us to render it correctly. This will allow other plugins to provide new badges and have them render correctly.

To render a badge we need:

1. the icon to render (which can be turned into a static path) - eg `badge.icon_name` below
2. a badge description, eg. "Level 3 Contributor" - eg. `badge.description`
3. some way of injecting a class attribute on the icon so that we can tweak colours etc. in CSS. - eg `badge-{{ badge.slug }}` and `badge.icon_classes`

Here's a stab based on what you have:

```html
<div class="achievement d-flex flex-column align-items-center">
  <div class="achievement-icon badge-{{ badge.slug }} {{ badge.icon_classes }}">
    <img src="{% static 'images/badges/'|add:badge.icon_name %}">
  </div>
  <div class="achievement-description">{{ badge.description }}</div>
</div>
```

Is that sufficient to style things appropriately?

AliciadW:
@ro6ley and @longhotsummer please see the list of variables that I need in the badges object:
* slug (already available)
* level_group - this will refer to the level group (levels 0-9 to be level-group-1, levels 10-19 to be level-group-2, etc. @longhotsummer please confirm the grouping)
* level - this is needed to display the levels on the level badges
* svg - we need the complete SVG data to be sent through. The different SVGs that I think we will need initially have been added under `images>badges`. `country.svg` will be used for the country badge; `level.svg` will be used for the level badge; `permission.svg` for the permissions badges.
* description (already available) - can we change this to have a shorter description? (i.e. only the country name instead of 'Can make changes to works for {{country}}'?) Maybe description_short.

I want to add a popover to each badge that contains the date/time when it was awarded as well as the longer descriptions.

longhotsummer:
> level_group - this will refer to the level group (levels 0-9 to be level-group-1, levels 10-19 to be level-group-2, etc. @longhotsummer please confirm the grouping)

The challenge with these is that indigo-points handles levels, and indigo is ignorant of them. So indigo needs to be able to render badges without explicitly knowing about points.

My suggestion is that a badge object has a list of CSS classes in `badge_classes` which are dumped into the surrounding div. The badge object itself can then ensure that it has, say, `badge-level-12 badge-level-group-10` or whatever is necessary for styling.

> level - this is needed to display the levels on the level badges

I see three options:

1. we use scss to set the correct level number based on the badge class -- that means we'll have a hard-coded list of classes similar to:

```css
.badge-level-1::before {content: "1";}
```

which isn't a huge problem - we could even use SCSS magic to script that.

2. we have svg's for ever level number (someone has to create them, or script this)

3. the template pulls in a list of arbitrary data attributes from the badge object, and for level badges that includes the level number.

Option 1 is pretty simple, brute force. Option 2 is probably a lot of work. Option 3 needs some coding but can also work.

AliciadW:
Hi @longhotsummer and @ro6ley. Sorry for only replying now.

I think options 1 and 3 are what we should consider. I'll implement option 1 and see how far we come with that. 

I'll also add the popovers in the meantime so we can see how that will work.

AliciadW:
Screenshot 2019-09-20 at 07 59 26

AliciadW:
Screenshot 2019-09-24 at 20 06 26

AliciadW:
@longhotsummer I think we are almost there! :-)

AliciadW:
> @ro6ley and @longhotsummer please see the list of variables that I need in the badges object:
> 
> * slug (already available)
> * level_group - this will refer to the level group (levels 0-9 to be level-group-1, levels 10-19 to be level-group-2, etc. @longhotsummer please confirm the grouping)
> * level - this is needed to display the levels on the level badges
> * svg - we need the complete SVG data to be sent through. The different SVGs that I think we will need initially have been added under `images>badges`. `country.svg` will be used for the country badge; `level.svg` will be used for the level badge; `permission.svg` for the permissions badges.
> * description (already available) - can we change this to have a shorter description? (i.e. only the country name instead of 'Can make changes to works for {{country}}'?) Maybe description_short.
> 
> I want to add a popover to each badge that contains the date/time when it was awarded as well as the longer descriptions.

I think I've managed to get all the info without the need for the above.

longhotsummer:
Greg Kempe – Laws Africa 2019-09-26 12-54-17

longhotsummer:
Thanks @AliciadW !

PR #738, opened 2019-09-02
Place metrics

longhotsummer:
This starts tracking the amount of activity per place, and opens the path for other per-place metrics. The goal of this is to provide an overview/indication of how much is going on in a place. This is useful for users who'd like to get an idea which places are active/inactive and to what extent.

Every time an Action is created, we increment the day's n_activities counter.

Places – Laws Africa 2019-09-02 11-28-48

Metrics – South Africa – Laws Africa 2019-09-02 11-29-16


goose-life:
nice!

I'm not sure why, I'm pretty sure it's on the rendering and not on the recording that it's going wrong, but I don't think the metrics are displaying as expected.
- there's something not so hot about the behaviour of the 'days' dropdown (max of 4 months displayed at bottom?)
- graphs aren't populating
- info populating inconsistently (e.g. blank of blank works complete on 30 days)

image

image

image

indigo_metrics/signals.py
```diff
@@ -0,0 +1,14 @@
+from django.db.models import signals
+from django.dispatch import receiver
+from actstream.models import Action
+
+from indigo_metrics.models import DailyPlaceMetrics
+
+
+@receiver(signals.post_save, sender=Action)
+def action_created(sender, instance, **kwargs):
+    """ Send 'created' action to activity stream if new task
```
goose-life:
```suggestion
    """ Send 'created' action to activity stream
```

longhotsummer:
The historical data for works and completeness are calculated every night. They won't show up on your system because the jobs aren't done.

Do this to populate metrics for today: `python manage.py update_daily_metrics`

That should fill in the blanks.

> there's something not so hot about the behaviour of the 'days' dropdown (max of 4 months displayed at bottom?)

What do you mean?

goose-life:
> That should fill in the blanks.

Done, thanks! looking great!

>>    there's something not so hot about the behaviour of the 'days' dropdown (max of 4 months displayed at bottom?)

> What do you mean?

The 'days' dropdown – 30, 90, etc – gives the impression that it'll show that range of dates in all 4 graphs in that section, but it doesn't. Is it maybe leaving out periods of inactivity?

longhotsummer:
Oh I see, yes all graphs will change. Currently it doesn't do a good job if there isn't much recorded activity (like in your screenshots). In production, it does a bit of a better job when we have the long-term history. Problem is that for some charts we fill in zeros, but not in others.

We should fix that.

PR #748, opened 2019-09-15
fix paste

goose-life:
if happy see https://github.com/laws-africa/indigo-lawsafrica/pull/56

### DIFF
--- a/indigo_app/static/javascript/indigo/views/document_editor.js
+++ b/indigo_app/static/javascript/indigo/views/document_editor.js
@@ -115,14 +115,15 @@
 
               pasteTable(table);
             }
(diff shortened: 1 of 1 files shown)


indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -115,12 +115,13 @@
 
               pasteTable(table);
             }
```
longhotsummer: I think we need a `return` here, just after the for loop. Then, the `else` statement on line 120 can fall away. We should execute 121-124 in two situations: there's HTML but not tables, or there's no HTML.

PR #762, opened 2019-09-21
Email Notification Context for Annotation Replies

### LINKED ISSUE #731: as a user I want more context when getting an email about a comment on a doc
Currently, emails for replies to document annotations look like this:

```
marinarubidge replied to a comment on Document #2828 – Dilapidated
Buildings and Unsightly and Objectionable Structures:

So I am ignoring this for now?
```

which has no context. Instead, could we also include the original comment? Or maybe the full thread of comments, with the most recent reply highlighted or called out in bold or something?

ro6ley:
#### What does this PR do?
Adds more context to the email notification about a new comment on a doc.

When a new reply to a comment on a document is posted, an email is sent out to all the participants of the current thread. This PR adds more context to the email including the thread of comments and a summary of the task linked to the original comment if available.

The contextual data includes:
- [x] initial comment/annotation
- [x] ellipsis if there's more than 3 replies
- [x] last two replies
- [x] task summary

> Task summary is also added to emails sent when a task receives new comment.

- Create a document,
- Create a comment and an optional task linked to the annotation,
- Reply to that comment as other users and check the email sent

closes #731 

#### Screenshots:
Email sent to other users when there are more than 9 replies to a comment:
image

Email sent to other users when there are less than 9 replies to a comment:
image

Email sent when a task receives a new comment, it now has the task summary at the bottom:
image

#### Any background context you want to provide?
The styling on the `annotation_new_reply` email needs some cleanup.


ro6ley:
I am looking into making the thread appear as it would appear as comments on a task.

ro6ley:
@longhotsummer thoughts on this design:
image

indigo_app/notifications.py
```diff
@@ -86,22 +86,42 @@ def notify_reply_to_annotation(self, annotation):
         the currently replying user.
         """
         parent_annotation = annotation.in_reply_to
-        related_annotations = Annotation.objects.filter(in_reply_to=parent_annotation)
+        related_annotations = Annotation.objects.filter(in_reply_to=parent_annotation)\
+                                                .order_by('created_at')
         document = parent_annotation.document
+        related_task = parent_annotation.task or None
```
longhotsummer: you shouldn't need the `or None`

longhotsummer:
I think this is great. The only problem is that the email clients won't. Email clients don't load external CSS, or at least not all of it.

I haven't got to using https://github.com/roverdotcom/django-inlinecss yet, but maybe this is the right time to do so, for HTML emails?

I'd suggest we keep the layout and CSS simple - I don't know if all clients handle floating divs etc. HTML emails usually heavily use tables.

ro6ley:
I thought embedded CSS would do the trick, let me look into that package.

ro6ley:
@longhotsummer the Django InlineCSS package broke the template and this was the output:
image

This is also coupled with the fact that our email templates are in parts and not one continuous templates.

I have stripped down most of the styling I had before and tried to make the inline-styling as subtle as possible to achieve this:
image

I am looking to trimming down the CSS further and other libraries that can help.

longhotsummer:
@ro6ley if it helps, perhaps keep it easy for now: no complex card layouts. Maybe just the username in bold, then the comment?

ro6ley:
@longhotsummer like the one in the PR description?

ro6ley:
Trimmed it down to this:
image

indigo_app/templates/templated_email/annotation_new_reply.email
```diff
@@ -3,14 +3,68 @@
...
-  <p> {{ annotation.created_by_user.username }} replied to a comment on
+  <p> {{ annotation.created_by_user.get_full_name }} left a new reply to a comment on
     <a href="{{ SITE_URL }}{% url 'document' doc_id=document.pk %}?anntn={{ annotation.in_reply_to.pk }}">Document #{{ document.pk }} – {{ document.title }}</a>:
 
+  <div class="parent-annotation" style="padding: 1rem;">
+    <b> {{ parent_annotation.created_by_user.get_full_name }}: </b> {{ parent_annotation.text|urlize|linebreaksbr }}
+  </div>
+
+
+  <div class="replies">
```
longhotsummer: Is it even worth having classes now? It just seems confusing...

longhotsummer:
Looks good. Just think we could remove the class names Ready to merge?

ro6ley:
Classes removed. It's ready

PR #763, opened 2019-09-23
arbitrary expression dates

goose-life:
to do:

- [x] migration
- [x] `amendments_with_initial` should include fake amendments for each arbitrary expression date
- [x] views to create and update AEDs


indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -3,14 +3,25 @@
...
-            <input type="hidden" name="amending_work" value="">
-            <input type="hidden" name="date" value="">
-          </form>
+            <button type="button" class="btn btn-primary dropdown-toggle dropdown-toggle-split" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"><span class="sr-only">Toggle Dropdown</span></button>
+            <div class="dropdown-menu">
+              <form method="POST">
+                {% csrf_token %}
+                <button class="dropdown-item" type="submit" name="arbitrary_expression_date"
+                        value="{{ consolidation_date|date:'Y-m-d' }}"
+                        formaction="{% url 'new_work_arbitraryexpressiondate' frbr_uri=work.frbr_uri %}">Add consolidation date: {{ consolidation_date|date:'Y-m-d' }}</button>
```
longhotsummer: You can set the action directly on the form, it's not going to change between submit buttons, so it doesn't need to be `formaction` on the submit button.

goose-life:
image

indigo_app/templates/indigo_api/_work_points_in_time.html
goose-life: so this woud become a `submit` button?

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -64,7 +74,31 @@ <h6>Repealed
...
+                  <div class="dropdown-menu">
+                    <a href="#arbitrary-expression-date-form-{{ amendment.id }}" class="dropdown-item" data-toggle="collapse">Change date</a>
+                    {% if amendment.can_delete %}
+                    <button class="dropdown-item" type="submit" name="delete" data-confirm="Really delete this consolidation date?">Delete</button>
+                    {% endif %}
+                  </div>
+                </div>
+              {% endif %}
+
+              <div class="form-inline edit-wrapper mb-3 collapse" id="arbitrary-expression-date-form-{{ amendment.id }}">
```
longhotsummer: the form element already has this id?

indigo_api/models.py
```diff
@@ -458,33 +458,33 @@ def friendly_type(self):
         if plugin:
             return plugin.work_friendly_type(self)
 
-    def amendments_with_initial(self):
+    def expressions_by_date(self):
```
longhotsummer: This doesn't seem right to me since it's not actually returning expressions, it's a lot more than that (contrast with the `expressions` method). Maybe `expression_dates` or `possible_expression_dates`. Neither are great :/

indigo_app/views/works.py
```diff
@@ -363,6 +360,70 @@ def get_success_url(self):
...
 
+class AddArbitraryExpressionDateView(WorkDependentView, CreateView):
+    """ View to add a new arbitrary expression date.
+    """
+    model = ArbitraryExpressionDate
+    permission_required = ('indigo_api.add_amendment',)
+
+    def post(self, request, *args, **kwargs):
+        arbitrary_expression_date = ArbitraryExpressionDate(work=self.work)
+        arbitrary_expression_date.date = request.POST['arbitrary_expression_date']
```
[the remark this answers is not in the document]
goose-life: it does work, but I recognise that it might not be the most correct way of doing it :)

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -64,7 +74,31 @@ <h6>Repealed
         {% endif %}
 
         {% for amendment in event.amendments %}
-          {% if amendment.pk %}
+          {% if not amendment.amending_work and amendment.id %}
+            <h6>Consolidation</h6>
+
+            <form action="{% url 'edit_arbitrary_expression_date' frbr_uri=work.frbr_uri arbitrary_expression_date_id=amendment.id %}" method="POST" id="arbitrary-expression-date-form-{{ amendment.id }}">
```
longhotsummer:
```suggestion
            <form action="{% url 'edit_arbitrary_expression_date' frbr_uri=work.frbr_uri arbitrary_expression_date_id=amendment.id %}" method="POST" id="arbitrary-expression-date-{{ amendment.id }}">
```

indigo_app/templates/indigo_api/_work_points_in_time.html
longhotsummer: Ah, I understand. The ID of the form is wrong. I've pushed a fix.

PR #765, opened 2019-09-25
FIX choosing document expression date

longhotsummer:


longhotsummer:
I don't see any other references to the old method name.

PR #772, opened 2019-09-27
Feedback for errors during creation of works

### LINKED ISSUE #759: When I create a work that already exists, I want better feedback about what went wrong
Try creating a work that already exists (same year and number), and the page just refreshes without an error message.

I don't think we're showing the error for whatever field does uniqueness checks, or showing non-field-errors from the form.

ro6ley:
#### What does this PR do?

Give better feedback when something goes wrong when creating a work

When a user creates a new work with any error such as an existing similar work(same year and number), there is no feedback as the page just refreshes.

This PR adds a notification (as seen in the screenshot down below) to tell the user what went wrong.

- Create a new work
- Recreate the work that already exists (same year and number) 

closes #759 

#### Screenshots

When recreating an existing work:
image

### DIFF
--- a/indigo_app/templates/indigo_api/work_form.html
+++ b/indigo_app/templates/indigo_api/work_form.html
@@ -11,6 +11,18 @@
     <div class="alert alert-danger">You don't have permission to create or change a work.</div>
   {% endif %}
 
+  {% if form.errors %}
+    <div class="alert alert-danger" role="alert">
+      <p>There were some errors in the information you entered. Please correct the following:</p>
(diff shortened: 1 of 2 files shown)


longhotsummer:
Sending a redirect on a POST means that all the details the user entered are lost. A simpler way of displaying these errors is to simply show the form errors in an alert at the top of the page when it loads. Django handles all the redirect logic, we simply add the errors to the template.

A bootstrap `alert-danger` div is quick and easy.

longhotsummer:
Thanks!

PR #774, opened 2019-09-28
arbitrary expression date on bulk import

goose-life:


indigo/bulk_creator.py
```diff
@@ -325,6 +325,17 @@ def add_extra_properties(self, work, info):
             if info.get(extra_property):
                 new_prop = WorkProperty(work=work, key=extra_property, value=info.get(extra_property))
                 new_prop.save()
+            elif extra_property == 'arbitrary_expression_date' and not work.stub:
```
longhotsummer: On first glance I thought that this column would have contained the date to use. I'm worried that as a flag column, it's confusing and links to something that is not in the  spreadsheet at all.

goose-life: `for extra_property in self.extra_properties.keys():` (line 324 above) gets the list of extra properties for this place as given in the bulkimporter.
The idea is to only add `'arbitrary_expression_date'` to that list for a place's bulkimporter if one wants to add the place's as-at-date as an arbitrary consolidation date on bulk import.

So you're right, it doesn't link to the spreadsheet, but it's not a column (flag or otherwise).

It may well be confusing in that a person could do a bulk import in a place without knowing that this is going to happen. I suppose I was only thinking of our current use case, for which I think that's fine, but if you'd rather do it a different way with a checkbox or something let's discuss.

goose-life:
forgot to say: goes with https://github.com/laws-africa/indigo-lawsafrica/pull/63

Issue #678, opened 2019-07-16
Introduce an "as at" date for place (country and locality)


longhotsummer:
Probably do this as a per-place setting, requiring a new per-place settings page.

This will be updated every week (say), once the gazette has been checked for updates.

* [x] model
* [x] Render this on the coverpage after the amendment/commencement information. eg. "Up to date as at XXX"
* [x] indigo-web
* [x] Include it in the API for a work.

longhotsummer:
https://github.com/laws-africa/indigo/pull/711 creates the `as_at_date` setting on a place.

PR #778, opened 2019-10-04
As-at date notice

longhotsummer:
Requires an updated indigo-web to style correctly

Example @ 2010-12-05 2019-10-04 16-27-57


longhotsummer:
Ref #678

PR #779, opened 2019-10-07
internal references

goose-life:
please note: first commit includes a hack to make the test pass that should be reversed


indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,121 @@ class RefsFinderENG(BaseRefsFinder):
...
+        self.ns = root.nsmap[None]
+        self.nsmap = {'a': self.ns}
+        self.ref_tag = "{%s}ref" % self.ns
+        self.ancestor_xpath = etree.XPath('|'.join(f'.//a:{a}' for a in self.ancestors), namespaces=self.nsmap)
+        self.candidate_xpath = etree.XPath(self.candidate_xpath, namespaces=self.nsmap)
+
+    def make_href(self, node, match):
+        """ Turn this match into an href
+        """
+        num = match.group("num")
```
longhotsummer: Technically you should escape any possible `'` chars in `num`, otherwise they will break the xpath. But the chances of the num having an apostrophe are pretty slim.

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,121 @@ class RefsFinderENG(BaseRefsFinder):
...
+        candidate_element = node.xpath(f"//a:section[a:num[text()='{num}.']]", namespaces=self.nsmap)[0]
```
longhotsummer: don't forget to handle when it can't find anything.

indigo/analysis/refs/base.py
goose-life: thanks, that's good to bear in mind! 
the regex shouldn't allow for accidentally including apostrophes though (especially not inside `num`), so I'll just file this away for future reference if that's good with you?

longhotsummer: sure. it'll break obviously if it breaks, which is better than quietly. :)

indigo/tests/test_refsfinders.py
```diff
@@ -0,0 +1,202 @@
...
+        <heading>Passive ref heading</heading>
+        <content>
+          <p>An important provision, but not the one referred to above.</p>
+        </content>
+      </section>
+        """
+            ),
+            language=self.eng)
+
+        expected_content = document.content
```
longhotsummer: ?

longhotsummer: This test doesn't look like it has the correct expected value?

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,132 @@ class RefsFinderENG(BaseRefsFinder):
...
+
+    """
+
+    # country, language, locality
+    locale = (None, 'eng', None)
+
+    section_re = re.compile(
+        r'''\b[sS]ections?\s+
+            (?P<ref>
+             (?P<num>\d+)
```
longhotsummer: What about "Section 26b"?

Perhaps also a test case ensuring that "subsection 26(b)" is not matched? Your regex looks fine, but a test is a good idea.

indigo/tests/test_refsfinders.py
```diff
@@ -0,0 +1,202 @@
...
+            language=self.eng)
+
+        expected = Document(
+            document_xml=document_fixture(
+                xml="""
+      <section id="section-7">
+        <num>7.</num>
+        <heading>Active ref heading</heading>
+        <content>
+          <p>As given in section <ref href="#section-26">26</ref>, blah.</p>
```
longhotsummer: Based on our discussion, we want to make this:

```html
<p>As given in <ref href="#section-26">section 26</ref>, blah.</p>
```

indigo/tests/test_refsfinders.py
goose-life: no it does, we're testing here to make sure that external section refs don't get marked up by accident

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,150 @@ class RefsFinderENG(BaseRefsFinder):
...
+
+    section_re = None
+    """ This must be defined by a subclass. It should be a compiled regular
+    expression, with named captures for `ref` and `num`.
+    """
+    # TODO: add subsection reference
+    candidate_xpath = None  # this must be defined by a subclass
+
+    # the ancestor elements that can contain references
+    ancestors = ['body', 'mainBody', 'conclusions']
```
longhotsummer: This looks fine to me. We won't check for references in preamble or preface.

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,150 @@ class RefsFinderENG(BaseRefsFinder):
...
+                while node is not None and node.tail:
+                    matches = list(self.section_re.finditer(node.tail))
+
+                    # if nothing in the tail matches anymore, we're done
+                    if not matches:
+                        break
+
+                    # there might be just one match that's a baddie,
+                    # in which case we're also done
+                    if len(matches) == 1 and not self.is_valid(matches[0], node):
```
longhotsummer: This smells weird to me. It feels like a specific instance of a general edge case: if there's nothing to do, stop. We should be able to handle it in a generic way. eg. if there are two non-valid matches in a row.

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,150 @@ class RefsFinderENG(BaseRefsFinder):
...
+                        break
+
+                    # mark the reference and continue to check the new tail
+                    node = self.mark_refs_in_matches(matches, node, in_tail=True)
```
longhotsummer: This will loop infinitely if the only matches are bad ones. We need a way for this method to tell us that there were no valid matches and we need to stop, perhaps by returning `None`.

longhotsummer: It may be easier to do the loop here, not in `mark_refs_in_matches`, because iterating over `matches` here is different to iterating over `matches` in the non-tail block above.

goose-life: hmm, but if the first match is bad we don't want to discard the node, which is why we introduced the loop in the first place. this part of it should in theory be the same in both cases: find the first valid reference, mark it and put the rest into the tail, then check the tail.

indigo/analysis/refs/base.py
```diff
@@ -123,3 +123,154 @@ class RefsFinderENG(BaseRefsFinder):
...
+    """
+
+    # country, language, locality
+    locale = (None, 'eng', None)
+
+    section_re = re.compile(
+        r'''(?P<ref>
+        (?P<section_ref>
+        \b[sS]ections?\s+
+        (?P<num>\d+(?!\()([A-Z]+)?)
```
longhotsummer: The negative lookahead for the `(` is redundant, since you know exactly what you want to look for: `\d+[a-z]*`. I suggest making the whole regex case insensitive (`re.IGNORECASE` flag).

indigo/analysis/refs/base.py
longhotsummer: Perhaps this is easier to understand:

```python
if not candidate.is_tail:
    # text directly inside a node
    for match in matches = self.section_re.finditer(node.text):
        if self.is_valid(match, node):
            # mark the reference and continue to check the new tail
            node = self.mark_reference(node, match, False)
            break

while node is not None and node.tail:
    for match in self.section_re.finditer(node.tail):
        if self.is_valid(match, node):
            node = self.mark_reference(node, match, True)
            # the node (and its text) have changed, start again on the tail of the new node
            break
    else:
        # we didn't break out of the loop, so there are no valid matches, give up
        node = None
```

Rather than checking for all the negative cases up front, we handle each match one at a time.

longhotsummer: My motivation for not having `mark_refs_in_matches` is that (a) it doesn't hide anything complex (it's just looping over an array) and (b) it loses the context necessary to know what to do if there *isn't* a match.

goose-life: gotcha, this makes total sense thanks

indigo/analysis/refs/base.py
goose-life: The new negative lookahead for `(` will discard any section references below subparagraph level, so e.g. `section 26(5)(b)(iv)(A)` won't come up as a match at all. we need to know what comes after the whole reference to check if it's an external reference (and thus ignore it).
I'd rather not make it case insensitive as the casing of the different elements is usually significant.

PR #782, opened 2019-10-07
url encode attachment filenames

goose-life:


longhotsummer:
I wouldn't have thought this was necessary -- what problem does it fix?

PR #790, opened 2019-10-12
Work properties, custom work form

longhotsummer:
Allows us to override custom work properties more easily.


indigo_api/models.py
```diff
@@ -272,6 +272,9 @@ class Meta:
 
     stub = models.BooleanField(default=False, help_text="Stub works do not have content or points in time")
 
+    # key-value pairs of extra data, using keys for this place from PlaceSettings.work_properties
+    properties = JSONField(null=False, blank=False, default=dict)
```
goose-life:
```suggestion
    properties = JSONField(null=False, blank=True, default=dict)
```
otherwise we get an error on bulk import at https://github.com/laws-africa/indigo/blob/master/indigo/bulk_creator.py#L238

PR #802, opened 2019-10-14
support template tweaks

goose-life:


indigo_api/models/works.py
```diff
@@ -170,6 +170,19 @@ def repeal(self):
     def place(self):
         return self.locality or self.country
 
+    @property
+    def amended(self):
```
longhotsummer: `.all()` goes to the database and loads all the data, which is very expensive. Instead, use `.exists()`.

Since this method is "expensive" (ie. it goes to the database, it's not quick), don't make it a `@property`. A property should be something that is cheap to call. By making it a method, we're indicating that it's expensive to call.

indigo_api/models/works.py
```diff
@@ -170,6 +170,19 @@ def repeal(self):
...
+        if self.amendments.all():
+            return True
+
+    @property
+    def most_recent_amendment_year(self):
```
longhotsummer: It's far cheaper to do the sorting in the db: `self.amendments.order_by('-date').first()` or something similar.

longhotsummer: Ditto property comment from above.

indigo_api/models/works.py
longhotsummer: https://softwareengineering.stackexchange.com/questions/367161/when-is-it-not-pythonic-to-use-properties-instead-of-a-method-that-doesnt-take?rq=1

goose-life: thanks! if I've understood you, if I use `.exists()` instead I can leave it as a property?

PR #803, opened 2019-10-15
add support for transforming work properties on saving

goose-life:


longhotsummer:
Isn't it easier for subclasses to just override `add_extra_properties`?

PR #812, opened 2019-10-21
Refs tweaks

longhotsummer:
* update UI to unify act refs and internal refs
* move section-specific logic out of base class
* scope target lookups to the closest ancestor (prevents mixing schedules with main content)
* avoid finding target twice for each valid match


indigo/analysis/refs/base.py
```diff
@@ -125,17 +127,18 @@ class RefsFinderENG(BaseRefsFinder):
     candidate_xpath = ".//text()[contains(., 'Act') and not(ancestor::a:ref)]"
 
 
-class BaseSectionRefsFinder(LocaleBasedMatcher):
-    """ Finds internal references to sections in documents.
-
+class BaseInternalRefsFinder(LocaleBasedMatcher):
```
longhotsummer: "section" is tradition-specific.

indigo/analysis/refs/base.py
```diff
@@ -157,31 +160,22 @@ def setup(self, root):
         self.ancestor_xpath = etree.XPath('|'.join(f'.//a:{a}' for a in self.ancestors), namespaces=self.nsmap)
         self.candidate_xpath = etree.XPath(self.candidate_xpath, namespaces=self.nsmap)
 
-    def make_href(self, node, match):
```
longhotsummer: Moved this into subclass since its tradition specific (knowing to find "section" elements).

indigo/analysis/refs/base.py
```diff
@@ -157,31 +160,22 @@ def setup(self, root):
...
         for ancestor in self.ancestor_nodes(root):
             for candidate in self.candidate_nodes(ancestor):
                 node = candidate.getparent()
 
                 if not candidate.is_tail:
                     # text directly inside a node
-                    for match in self.section_re.finditer(node.text):
-                        if self.is_valid(match, node):
+                    for match in self.ref_re.finditer(node.text):
+                        if self.is_valid(node, match):
```
longhotsummer: for consistency: other methods are `node, match`, rather than `match, node`.

indigo/analysis/refs/base.py
```diff
@@ -190,18 +184,13 @@ def find_references(self, root):
                         # we didn't break out of the loop, so there are no valid matches, give up
                         node = None
 
-    def is_valid(self, match, node):
-        # check that it's not an external reference
-        ref = match.group('ref')
```
longhotsummer: tradition/language-specific

indigo/analysis/refs/base.py
```diff
@@ -246,25 +240,50 @@ class SectionRefsFinderENG(BaseSectionRefsFinder):
...
+        ref = match.group(0)
+        if ref.endswith('the ') or ref.endswith('Act '):
+            return False
+
+        return super().is_valid(node, match)
+
+    def find_target(self, node, match):
+        num = match.group('num')
+        # find the closest ancestor to scope the lookups to
+        ancestor = closest(node, lambda e: e.tag in self.ancestor_tags)
```
longhotsummer: `closest()` mimics jquery's `closest` method.

indigo/analysis/refs/base.py
```diff
@@ -246,25 +240,50 @@ class SectionRefsFinderENG(BaseSectionRefsFinder):
...
-        (?P<section_ref>
-        \b[sS]ections?\s+
-        (?P<num>\d+[A-Z]*)
+    ref_re = re.compile(
+        r'''
+        (?P<ref>
+          \b[sS]ections?\s+
+          (?P<num>\d+[A-Z]*)
         )
         (?P<subsection_ref>\s*\(\d+[A-Z]*\))?
```
longhotsummer: I think these named groups add complexity unnecessarily, because they aren't used at all.

indigo/analysis/refs/base.py
```diff
@@ -246,25 +240,50 @@ class SectionRefsFinderENG(BaseSectionRefsFinder):
...
         (?P<paragraph_ref>\s*\([a-z]+[A-Z]*\))?
         (?P<subparagraph_ref>\s*\([ivx]+[A-Z]*\))?
         (?P<item_ref>\s*\([a-z]{2,}[A-Z]*\))?
         (?!\s*\()
         (\s+of\s+(this\s+Act|the\s+|Act\s+)?)?
-        )''',
+        ''',
         re.X)
 
     candidate_xpath = ".//text()[contains(., 'section') and not(ancestor::a:ref)]"
```
longhotsummer: Worth noting that xpath is case sensitive, so this will never find text elements with "Section".

indigo/analysis/refs/base.py
goose-life: I guess it's a bit early as we're not using them yet but I thought it might make it more readable, and they should come in handy once we start matching below section level?

indigo/analysis/refs/base.py
goose-life: thanks! 
```suggestion
     candidate_xpath = ".//text()[contains(translate(., 'S', 's'), 'section') and not(ancestor::a:ref)]"
```

indigo/analysis/refs/base.py
longhotsummer: This is a bit like writing a method or function that you think you may use, but don't know quite yet. It's better to write it later, when you actually need it, than to guess what you'll need in future. In most cases, that guess will be wrong.

Issue #696, opened 2019-07-23
as an importer I want to choose which pdf pages to import


goose-life:
By-laws are often published in gazettes with many pages and currently need to be cropped beforehand

![image](https://user-images.githubusercontent.com/32566441/61711606-b9593480-ad54-11e9-8cc1-a9c26584b27e.png)

PR #818, opened 2019-10-24
Choose PDF pages

longhotsummer:
* the user can type in the page numbers to be imported
* the extracted pages are attached as the imported PDF

Import – Laws Africa 2019-10-24 09-21-44


longhotsummer:
Closes #696

PR #823, opened 2019-10-28
Advanced Work Filters

### LINKED ISSUE #786: As a researcher I'd like advanced work filtering options
* repealed (possibly by date)
* commenced/not commenced (possibly by date)
* primary work / subleg
* assent date
* amended or not (by date?)
* publication date

ro6ley:
#### What does this PR do?

Add advanced work filtering options.

This PR adds an "Advanced filter" section to the work filter page which is hidden by default. When expanded, it shows additional parameters to filter the work listing as can be seen in the screenshot below. 

Full feature listing:
- [x] Primary/Subleg work filter functional
- [x] Assent date filter functional
- [x] Repeal date filter functional
- [x] Commencement date filter functional
- [x] Amendment date filter functional
- [x] Publication date filter functional
- [x] Hidden advanced filter section that comes into view at the click of the advanced button
- [x] For the date filters, checking the adjacent checkbox, automatically makes the fields required 
- [x] Automatic detection of active advanced filters to keep the section open after submitting 
- [x] Date swapping when the start date is later than the end date (_works in the backend, not reflecting on the frontend_)
- [ ] Prefiltered categories
- [x] Reset filters button (_does not empty the form, rather resets it to previously submitted state_)

- Click on the `advanced` button
- Select the parameters needed to filter the works
- Click on `Filter`

closes #786 

#### Screenshots
image


indigo_app/forms.py
```diff
@@ -276,6 +300,104 @@ def filter_queryset(self, queryset):
         if self.cleaned_data.get('taxonomies'):
             queryset = queryset.filter(taxonomies__in=self.cleaned_data.get('taxonomies'))
 
+        # Advanced filters
+        # filter by assent date range
+        if self.cleaned_data.get('assent_date_check') and self.cleaned_data['assent_date_check']:
```
longhotsummer: You shouldn't need the second expression. `if self.cleaned_data.get('assent_date_check')` should be sufficient.

longhotsummer: This repeated block of code for the date checker cries out to be turned into a helper function :)

longhotsummer:
Looking good, Robley! I've made a few comments, even though I know it's still a draft :)

ro6ley:
@longhotsummer I wanted to know your opinion on the date periods defaulting to 6 months period to the current date, I have enforced filling both dates on the form but I felt like it was not enough since it is a GET form.

indigo_app/forms.py
ro6ley: Let me look into it

longhotsummer:
Good question. What if we default the unspecified date to whatever the specified date is? So, if they submit the form with `repeal_date_start = X` and no `repeal_date_end`, then we set `repeal_date_end` to `X`, so that effectively we're saying `repeal_date == X`?

That's perhaps a clearer default than 6 months.

longhotsummer:
If you'd like to ensure that the new values show up in the rendered form, you can implement `clean_assent_date_start` and `clean_assent_date_end` and provide defaults if necessary. Those values will then (I think) show up when the form re-renders.

See https://docs.djangoproject.com/en/2.2/ref/forms/validation/#cleaning-a-specific-field-attribute

ro6ley:
Cleaning the data as specified does not reflect on rerendering, but I think we can still use it to set the default dates.

ro6ley:
Also, as a way to cover for the dates not switching on the frontend, I have ordered the results by the dates starting with the latest.

longhotsummer:
This is looking great, Robley. I've tweaked the view a bit to move more options into the advanced view so the default view is less crowded. I decided the JS to swap dates was unwieldy and have dropped it for now, I shouldn't have suggested it :)

After playing with it a bit, I realise that we need to support different options for the date filters. I've opened https://github.com/laws-africa/indigo/pull/844 to do this for the assent date.

Please can you do the same for:

* publication: any, "No publication date" (none), "Published between..."
* repealed: any, "Not repealed" (no), "Repealed" (yes), "Repealed between..."
* commenced: any, "Not commenced" (no), "Commenced" (yes), "Commenced between..."
* amended: any, "Not amended" (no), "Amended" (yes), "Amended between..."

ro6ley:
on it @longhotsummer

longhotsummer:
Nice. "Commenced between" doesn't toggle the date range.

indigo_app/templates/indigo_api/_work_filter_form.html
```diff
@@ -60,9 +35,173 @@
...
+          <div class="form-row">
+            <div class="col">
+              <select name="{{ form.commencement.html_name }}" class="form-control" id="{{ form.commencement.id_for_label }}" data-toggle="daterange">
+                {% for opt in form.commencement %}
+                  {{ opt }}
+                {% endfor %}
+              </select>
+            </div>
+            <div class="col">
+              <div class="input-group input-daterange {% if form.commencement.value != 'range'%}d-none{% endif %}" id="commencedment-daterange">
```
longhotsummer:
```suggestion
              <div class="input-group input-daterange {% if form.commencement.value != 'range'%}d-none{% endif %}" id="commencement-daterange">
```

ro6ley: I had just spotted it and was about to commit😅

PR #828, opened 2019-10-29
Work chooser pre-selects locality.

longhotsummer:


indigo_app/static/javascript/indigo/views/work_amendments.js
```diff
@@ -20,7 +20,10 @@
     addAmendment: function(e) {
       e.preventDefault();
 
-      var chooser = new Indigo.WorkChooserView({country: this.model.get('country')}),
+      var chooser = new Indigo.WorkChooserView({
+            country: this.model.get('country'),
+            locality: this.model.get('locality'),
```
goose-life:
```suggestion
            locality: this.model.get('locality') || "-",
```
this will select (none) for locality when on country level

longhotsummer:
Nice, thanks. I've added `-` as the default locality for all work choosers, so that it always behaves correctly.

PR #831, opened 2019-10-30
warn about date/s having to be changed manually

### LINKED ISSUE #626: As an editor changing a work's commencement date, I want to be aware that I may need to change the dates of impacted amended and repealed works
If the (main) commencement date of a work is updated, amendments made by that work that have already been recorded don't automatically update to having occured on the new date.
This is probably right, because amendments made by a work don't always and only happen on its main commencement date.
If a commencement date was captured incorrectly and caught later (after related works have been linked), whoever is making the correction should probably be shown which dates will now automatically update (coverpage of any documents on work), and importantly which won't, so that they can be manually updated if needed (date/s of amendment/s to other work/s, repeal date/s of other work/s, etc).
Eventually it might be nice to have checkboxes for all the places where this date has already been recorded in relation to this Act at the place where the main commencement date is changed with the option to select all / none / individual ones to be automatically updated at the same time.
For now just a heads-up would be good I think (it doesn't happen all that often).

goose-life:
closes #626 

Do this if the work being edited
- already has a commencement date
- has amended any other works OR
- has repealed any other works


indigo_app/templates/indigo_api/work_form.html
```diff
@@ -233,6 +233,11 @@ <h5 class="card-header">Commencement and Assent</h5>
             <label for="{{ form.commencement_date.id_for_label }}">Commencement date</label>
             <input type="text" class="form-control" data-provide="datepicker" id="{{ form.commencement_date.id_for_label }}" name="{{ form.commencement_date.html_name }}" placeholder="yyyy-mm-dd" pattern="\d{4}-\d\d-\d\d" value="{{ form.commencement_date.value|date:"Y-m-d"|default:'' }}">
             <p class="form-text text-muted">Date when most of the work comes into force<br>(only leave blank if not yet commenced)</p>
+            {% if work.pk and work.commencement_date %}
+              {% if work.amendments_made.exists or work.repealed_works.exists %}
+                <div class="alert alert-danger">If you change the commencement date here, you'll also have to change any related amendment / repeal dates manually. See this work's <a href="{% url 'work_related' work.frbr_uri %}">related works</a>.</div>
```
longhotsummer:
```suggestion
                <div class="alert alert-warning">If you change this commencement date, you may also have to manually change the dates for related <a href="{% url 'work_related' work.frbr_uri %}">amendments and repeals</a> made by this work.</div>
```

PR #849, opened 2019-11-12
Commencement

### LINKED ISSUE #647: As a user I want to know if something is definitely not commenced
as opposed to us just not having the commencement info.

Make it explicit with `not yet commenced` flag in
- [ ] API (?)
- [ ] coverpage
- [ ] work form
- [ ] bulk creation?

goose-life:
closes #647


goose-life:
image

image

image

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -229,23 +229,34 @@ <h5 class="card-header">Commencement and Assent</h5>
...
-                <div class="alert alert-warning">If you change this commencement date, you may also have to manually change the dates for related <a href="{% url 'work_related' work.frbr_uri %}">amendments and repeals</a> made by this work.</div>
-              {% endif %}
-            {% endif %}
+          <div class="form-group">
+            <div class="form-check">
+              <input type="checkbox" class="form-check-input" id="{{ form.commenced.id_for_label }}" name="{{ form.commenced.html_name }}" value="on" {% if form.commenced.value %}checked{% endif %}>
+              <label class="ml-2 mr-2" for="{{ form.commenced.id_for_label }}">Commenced – this work has come into force</label>
+            </div>
           </div>
+          <div id="commencement_alert" class="alert alert-warning">If you clear this and Save, the commencement date and/or work will be removed.</div>
```
longhotsummer: I'm not sure this is necessary. The entry fields disappear when it's unchecked, which I think makes it fairly obvious. Additionally, it's not a common operation to uncheck this once details have been entered.

indigo_app/templates/indigo_api/work_layout.html
```diff
@@ -72,6 +72,8 @@ <h5 class="main-header-title">
   <span class="work-title">{% if work.pk %}{{ work.title }}{% else %}Create a new work{% endif %}</span> 
   {% if work.stub %}<span class="badge badge-info">stub</span>{% endif %}
   {% if work.repealed_date %} <span class="badge badge-info">repealed</span> {% endif %}
+  {% if not work.commenced %} <span class="badge badge-info">not commenced</span>
+    {% elif not work.commencement_date %} <span class="badge badge-info">commencement uncertain</span> {% endif %}
```
longhotsummer: Normally `elif` is indented to the same level as `if`, so this line is overindented. Then the `endif` would come on a new line at the same indentation level.

indigo_api/templates/indigo_api/akn/coverpage_act.html
```diff
@@ -44,7 +44,15 @@ <h2>{{ subt }}</h2>
...
       {% endif %}
 
-      {% if document.commencement_date %}
+      {% if not document.work.commenced %}
+        <li class="commencement-date">
+          {% blocktrans %}Not yet commenced: This work has not come into force yet.{% endblocktrans %}
+        </li>
+      {% elif not document.commencement_date %}
+        <li class="commencement-date">
+          {% blocktrans %}Commencement uncertain: This work has likely commenced but a commencement date hasn't been captured.{% endblocktrans %}
```
longhotsummer: I'd like to reword this, I'm not sure how yet. I think we should avoid using "work" for general users because it can be a confusing term. But I'm not sure what other word to use.

indigo_app/static/javascript/indigo/views/work.js
```diff
@@ -49,6 +49,8 @@
       '#id_work-publication_name': 'publication_name',
       '#id_work-publication_number': 'publication_number',
       '#id_work-stub': 'stub',
+      // Including the below leads to 'Commenced' being unchecked when the form loads; no idea why.
+      '#id_work-commenced': 'commenced',
```
longhotsummer: This is working for me, perhaps it was earlier on when the `commenced` field wasn't being serialised yet.

longhotsummer:
@goose-life that JS issue seems to be working fine for me.

indigo_app/templates/indigo_api/work_form.html
goose-life: Agree it shouldn't be common; I wanted to warn people in case they think that un/checking the box just shows/hides the commencement details (because if you re-check it the details still show up)

goose-life: to be clear, it only shows up in specific circumstances:
 https://github.com/laws-africa/indigo/pull/849/files#diff-eecce2e94e31cbeb0cb999e041d7a824R233-R239

goose-life: i.e. when there are commencement details and the box has been unchecked

goose-life: that's not what it does, but it should be! fix incoming

longhotsummer: I still don't feel it's necessary. The options disappear, which I think is a clear indication that they're going to "go away" -- ie. be cleared. I feel the warning just makes it more confusing.

If the concern is that, after unchecking and then rechecking, the details are still there, then perhaps clear them instead? Like we do for removing the repeal.

indigo_api/templates/indigo_api/akn/coverpage_act.html
goose-life:
```suggestion
          {% blocktrans %}Commencement uncertain: This legislation has likely commenced but a commencement date hasn't been captured.{% endblocktrans %}
```

PR #852, opened 2019-11-13
Remove the site sidebar

longhotsummer:
This is a more common navigation pattern to make the sidebar links more obvious.

* in the document view, move functionality into a cleaned-up toolbar

Imported from provincial-gazette-za-wc-no-7075-dated-2012-12-14 pdf @ 2012-12-24 2019-11-13 18-58-18

* in work views, use a normal tab interface along the top of the pages

Imported from provincial-gazette-za-wc-no-7075-dated-2012-12-14 pdf – Laws Africa 2019-11-13 18-59-27


goose-life:
The other work-related sidebar buttons aren't a great loss given we can still use the breadcrumbs to go to the Overview, but the Tasks one I think does get used quite a bit by people who have just finished an Import content task

longhotsummer:
I had the same thought. Any ideas for where we could stash a link to the work's tasks? Toolbar?

longhotsummer:
Perhaps to the left of the Settings button. In time, that could be a dropdown to open a tasks box that shows a/the current document task.

indigo_app/views/works.py
```diff
@@ -105,6 +105,7 @@ class EditWorkView(WorkViewBase, UpdateView):
     form_class = WorkForm
     prefix = 'work'
     permission_required = ('indigo_api.change_work',)
+    tab = 'overview'
```
goose-life: if we nuked this line it would make it more obvious that the Overview is somewhere you currently aren't, which I like?
image

indigo_app/templates/main.html
```diff
@@ -9,139 +9,135 @@
...
+      {% block navbar %}
+        <nav class="navbar navbar-expand-md navbar-dark">
+          {% block navbar-brand %}
+          <a class="navbar-brand" href="/"><img src="{% static 'images/logo-inverse.png' %}" title="Indigo"></a>
+          {% endblock %}
+          {% block navbar-toggle %}
+          <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbar-collapse">
+            <span class="navbar-toggler-icon"></span>
+          </button>
+          {% endblock %}
```
goose-life: what does this do?

indigo_app/views/works.py
longhotsummer: nice.

indigo_app/templates/main.html
longhotsummer: the navbar-toggle? it gives you the menu button when on mobile, top right corner.

indigo_app/views/works.py
longhotsummer: done.

longhotsummer:
I've added a link to tasks.

Example @ 2010-12-05 2019-11-15 08-55-38

PR #855, opened 2019-11-18
Include place name in PDF header

longhotsummer:


longhotsummer:
280 pdf 2019-11-18 12-57-33

PR #859, opened 2019-11-18
Ensure that empty p tags take up space in HTML. 

### LINKED ISSUE #612: Prevent subsections/block list items from overlapping if they have numbers but no content.
eg. see: https://edit.laws.africa/places/na/tasks/2572/?c=139

```
(1) Subsection
    (a)
    (b) text
(2)
(3) subsections (2) and (3) deleted.
```

(b) will overlap (a), and (3) will overlap (2).

longhotsummer:
Partially fixes #612, for subsections (a relatively common case)

For block lists, our AKN should have an empty `<p/>` tag, but doesn't. We don't have any content in NA which matches this case.

Linter tester @ 2019-10-23 2019-11-18 15-51-14

### DIFF
--- a/indigo_api/static/xsl/act.xsl
+++ b/indigo_api/static/xsl/act.xsl
@@ -242,6 +242,15 @@
     </section>
   </xsl:template>
 
+  <!-- empty p tags must take up space in HTML, so add an nbsp;
+       empty means no child elements and only whitespace content -->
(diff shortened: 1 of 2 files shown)


longhotsummer:
https://github.com/laws-africa/indigo/issues/860 opened because Slaw generates incorrect empty blockList item XML.

PR #862, opened 2019-11-19
simplify regex, add test

### LINKED ISSUE #837: skip `thereof` in internal refs finder
e.g. the Pensions Act, … make regulations under section 23 thereof, …

goose-life:
closes #837 

regex simplified so that only `of this` is recognised as internal – `of` anything else (`of the`, `of that`, `of Act / Proc / Ord`, etc) will be assumed to be external, as will `thereof`. if there's no `of` or `thereof` it will also be assumed to be internal.

### DIFF
--- a/indigo/analysis/refs/base.py
+++ b/indigo/analysis/refs/base.py
@@ -148,7 +148,7 @@ class SectionRefsFinderENG(BaseInternalRefsFinder):
             (\d+[A-Z0-9]*(\s*\([A-Z0-9]+\))*)
           )*
         )
-        (\s+of\s+(this\s+Act|the\s+|Act\s+)?)?
+        (\s+of\s+(this)?|\s+thereof)?
         ''',
         re.X | re.IGNORECASE)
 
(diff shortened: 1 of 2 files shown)


indigo/analysis/refs/base.py
```diff
@@ -167,7 +167,7 @@ def setup(self, root):
     def is_valid(self, node, match):
         # check that it's not an external reference
         ref = match.group(0)
-        if ref.endswith('the ') or ref.endswith('Act '):
+        if ref.endswith('of ') or ref.endswith('thereof'):
```
longhotsummer: What if it ends with `of this`? Instead, perhaps check if the very last `match.group()` is non-empty? We're basically re-doing the regex with `endswith` otherwise.

longhotsummer:
Nice, I love some additional passing tests :)

indigo/analysis/refs/base.py
goose-life: There were existing tests for `of this` which are passing; I've reshuffled the tests so that hopefully their grouping makes a bit more sense.
I'm not sure I follow you on the last `match.group()` being non-empty – it _will_ be empty for simple references like `section 26`, but will contain `this` in the case of `section 26 of this XXX`.

longhotsummer:
Let's use a new permission on the Work model, `bulk_export_work`.

indigo/analysis/refs/base.py
longhotsummer: Ah, I understand now. "of " is the exception, not "of this". All good.

PR #869, opened 2019-11-28
Itals terms

goose-life:
will close https://github.com/laws-africa/indigo-kenyalaw/issues/63

- [ ] refactor recipe still needed
- [x] tests still needed


indigo/analysis/italics_terms.py
```diff
@@ -0,0 +1,44 @@
+from lxml import etree
+import re
+
+from indigo.analysis.markup import TextPatternMarker
+from indigo.plugins import LocaleBasedMatcher, plugins
+
+
+class BaseItalsFinder(LocaleBasedMatcher, TextPatternMarker):
```
longhotsummer: I would prefer `Italics` over `Itals`

indigo/analysis/italics_terms.py
```diff
@@ -0,0 +1,44 @@
...
+        root = etree.fromstring(document.content)
+        self.setup(root)
+        self.get_pattern_re(document)
+        self.markup_patterns(root)
+        document.content = etree.tostring(root, encoding='utf-8').decode('utf-8')
+
+    def get_pattern_re(self, document):
+        # TODO: do this differently once terms stored as an actual array
+        terms = document.work.country.settings.italics_terms[0]
+        f_string = f'{terms}'.replace('\r\n', '|').replace(' ', '\s*')
```
longhotsummer: `f'{terms}'` is just `terms`.

I suggest splitting on newlines, stripping trailing and leading whitespace, removing empty lines, the use `re.escape` to escape any special characters. Then join with `|`.

```python
terms = [t.strip() for t in terms.split()]
terms = [re.escape(t) for t in terms if t]
```

indigo/analysis/italics_terms.py
```diff
@@ -0,0 +1,44 @@
...
+        self.pattern_re = re.compile(rf'(?P<itals>{f_string})', re.X)
```
longhotsummer:
```suggestion
        self.pattern_re = re.compile(rf'(?P<itals>{f_string})')
```

longhotsummer: Actually, you don't even really need to name the capture group, since the entire thing is the capture group.

I suspect you'll want `\b` on either side of the terms: `\b(foo|bar)\b`

indigo/analysis/italics_terms.py
```diff
@@ -0,0 +1,44 @@
...
+        itals.text = match.group('itals')
+        return itals, match.start('itals'), match.end('itals')
+
+
+@plugins.register('italics_terms')
+class ItalsFinderENG(BaseItalsFinder):
+    """ Finds references to italics terms in English documents.
+    """
+    # country, language, locality
+    locale = (None, 'eng', None)
```
longhotsummer: In this case, I think `BaseItalsFinder` can be the actual plugin, because there's nothing language-specific about the functionality.

indigo_app/templates/indigo_api/document/_toolbar.html
```diff
@@ -23,6 +23,7 @@
     <div class="dropdown-menu">
       <a class="dropdown-item" href="#defined-terms-modal" data-toggle="modal">Defined terms...</a>
       <a class="dropdown-item" href="#references-modal" data-toggle="modal">References...</a>
+      <a class="dropdown-item" href="#italics-modal" data-toggle="modal">Italics terms...</a>
```
longhotsummer: `Italicise terms...`?

indigo_app/templates/place/settings.html
```diff
@@ -51,6 +51,19 @@ <h4 class="card-header">Settings</h4>
             </div>
           {% endif %}
         </div>
+
+        <div class="form-group">
+          <label for="{{ form.italics_terms.id_for_label }}">Italics terms</label>
+          <textarea class="form-control" rows="4" name="{{ form.italics_terms.name }}" id="{{ form.italics_terms.id_for_label }}">{{ form.italics_terms.value|default:'' }}</textarea>
+          <p class="form-text text-muted">Newline-separated list of italics terms for {{ place.country.name }}.</p>
```
longhotsummer:
```suggestion
          <p class="form-text text-muted">Terms that should be italicised for {{ place.country.name }}, one per line. Case sensitive.</p>
```

indigo_app/templates/indigo_api/document/_toolbar.html
goose-life: I was matching to 'Defined terms' and 'References' above, i.e. nouns as opposed to verbs?

goose-life: going with `Italicised` per in-person discussion

indigo_app/templates/indigo_api/document/_italics_terms.html
```diff
@@ -0,0 +1,39 @@
+<div class="modal fade" id="italics-modal" tabindex="-1">
+  <div class="modal-dialog modal-lg">
+    <div class="modal-content">
+      <div class="modal-header">
+        <h5 class="modal-title">Italics terms</h5>
```
longhotsummer:
```suggestion
        <h5 class="modal-title">Italicised terms</h5>
```

indigo_app/views/places.py
```diff
@@ -553,6 +561,19 @@ def form_valid(self, form):
         placesettings = self.object
         placesettings.updated_by_user = self.request.user
 
+        # save changes to italics_terms 1) as a list 2) to the country
+        if 'italics_terms' in form.changed_data:
```
longhotsummer: It feels like doing this in the form is the better place to do it. Perhaps https://docs.djangoproject.com/en/2.2/ref/contrib/postgres/forms/#simplearrayfield would simplify this?

longhotsummer: I've pushed a fix to do this, reduces our logic a bunch.

longhotsummer:
Looks good. I think it would be useful to have the list of automatically identified terms in the modal, since most users won't have access to the settings page. But let's deploy this as it is.

indigo_app/templates/place/settings.html
```diff
@@ -51,6 +51,19 @@ <h4 class="card-header">Settings</h4>
             </div>
           {% endif %}
         </div>
+
```
longhotsummer: This can all disappear.

indigo_app/tests/test_places.py
```diff
@@ -45,6 +45,10 @@ def setUp(self):
 
     def test_place_settings(self):
         form = self.app.get('/places/za/settings').forms[0]
+        form['spreadsheet_url'].value = 'https://docs.google.com/spreadsheets/u/1/d/1a2o-842lGliSwlLo3gSbYSRbaOYu-2PZhC1rOf8MgA4/edit'
```
longhotsummer: nuke

PR #875, opened 2019-12-05
Better annotations

### LINKED ISSUE #793: improve commenting experience
make it so I can highlight a piece of text rather than a whole element

longhotsummer:
Support annotations on text snippets, rather than just block elements with IDs.

* existing annotations still highlight the whole block
* user creates new annotations by highlighting text and then clicking the annotation button
* annotations are fairly robust to changes in text (eg. typo corrections, or a list item being renumbered)
* the `anchor_id` on the `Task` model is removed, relying instead on the task's associated annotation, until such a point that we really need to be able to link a task to something inside a document.
* annotations on table content, schedules and headings are now possible
* annotations on coverpage elements are not yet possible

Table-based annotations position and behave a bit weirdly, due to how we insert the annotation floater into the content of the document. Notably:

* annotations will be hidden when editing a table, and then re-appear when editing is cancelled or saved
* annotation floaters for tables appear inside the table

Fixes #793.

Imported from cap95 rtf @ 1980-01-01 2019-12-05 14-21-10

Imported from cap95 rtf @ 1980-01-01 2019-12-05 14-20-37

Imported from cap95 rtf @ 1980-01-01 2019-12-05 14-23-05

Imported from cap95 rtf @ 1980-01-01 2019-12-05 14-24-37

Imported from cap95 rtf @ 1980-01-01 2019-12-05 14-25-36


goose-life:
looks awesome! there does seem to be either a 
`TypeError: Argument 1 of Range.setStart is not an object.` or a 
`TypeError: selector is null` when the selected text is deleted or moved into a new element.

We could bump up against this quite quickly, e.g. in the case of 
image

longhotsummer:
I've pushed a fix, please check again @goose-life

indigo_app/static/javascript/indigo/dom.js
```diff
@@ -0,0 +1,269 @@
...
+
+    var start = selector.start || 0;
+    var end = selector.end || start;
+    var count = domSeek(iter, start);
+    var remainder = start - count;
+
+    if (iter.pointerBeforeReferenceNode) {
+      range.setStart(iter.referenceNode, remainder);
+    } else {
+      range.setStart(iter.nextNode(), remainder);
```
goose-life: `iter.nextNode()` should be an object? 
`TypeError: Argument 1 of Range.setStart is not an object.`

longhotsummer: This is third party code. Do you know how you triggered this error?

goose-life: oh dear. In the example screenshot from above, I moved the subs (2) down.

PR #876, opened 2019-12-06
tweaks

goose-life:
The only substantive change I think is changing 'Stub and 'Non-stub' to be 'Principal' and 'Stub'

image

image

### DIFF
--- a/indigo_app/templates/place/detail.html
+++ b/indigo_app/templates/place/detail.html
@@ -8,10 +8,10 @@
 
   <div class="row">
     <div class="col">
-      <div class="card mt-3">
+      <div class="card mt-4">
         <div class="card-header">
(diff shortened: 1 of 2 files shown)


indigo_app/templates/place/detail.html
```diff
@@ -8,10 +8,10 @@
 
   <div class="row">
     <div class="col">
-      <div class="card mt-3">
+      <div class="card mt-4">
         <div class="card-header">
           <a class="float-right" href="{% url 'tasks' place=place.place_code %}">Tasks →</a>
-          <h5 class="mb-0">Open tasks</h5>
+          <h5 class="mb-0">Current tasks ({{ place.place_tasks.unclosed.count }})</h5>
```
longhotsummer: @ro6ley I think you already have a variable with this number?

ro6ley: Yes, I have the number

PR #879, opened 2019-12-09
mark up more references

### LINKED ISSUE #649: As an editor I want to see links to works other than Acts
e.g. "LN 25 of 2009"
we'll have to check for all subtypes in a place (e.g. `GN`, `LN`) plus `\d+` plus `of \d{4}`

goose-life:
closes #649 

image


indigo/analysis/italics_terms.py
```diff
@@ -10,19 +10,23 @@ class BaseItalicsFinder(LocaleBasedMatcher, TextPatternMarker):
...
         # we have to re-parse it
         root = etree.fromstring(document.content)
+        self.get_candidate_xpath(italics_terms)
         self.setup(root)
         self.get_pattern_re(italics_terms)
         self.markup_patterns(root)
         document.content = etree.tostring(root, encoding='utf-8').decode('utf-8')
 
+    def get_candidate_xpath(self, terms):
+        xpath_contains = " or ".join([f"contains(., '{term}')" for term in terms])
```
longhotsummer: You'll need to escape apostrophes in `term`.

indigo/analysis/italics_terms.py
```diff
@@ -10,19 +10,23 @@ class BaseItalicsFinder(LocaleBasedMatcher, TextPatternMarker):
...
+        self.candidate_xpath = f".//text()[{xpath_contains} and not(ancestor::a:i)]"
```
longhotsummer: I think you'll want brackets around the contains clause, so that it's `(A or B or C) and not(ancestor::a:i)` otherwise the boolean logic precedence is incorrect (it'll be `A or B or (C and not(...))` because `and` binds more tightly than `or`).

indigo/analysis/refs/base.py
```diff
@@ -63,6 +64,51 @@ class RefsFinderENG(BaseRefsFinder):
...
+        P 52 of 2001
+        Ordinance no. 52 of 1998
+        GN 1/2009
+
+    """
+
+    # country, language, locality
+    locale = (None, 'eng', None)
+
+    subtypes = [s for s in Subtype.objects.all()]
```
longhotsummer: Don't access the database when declaring a class. This needs to be done in the `setup` method or similar.

This is probably the cause of the test failures.

indigo/analysis/refs/base.py
```diff
@@ -63,6 +64,51 @@ class RefsFinderENG(BaseRefsFinder):
...
+                (?P<num>\d+)
+                (\s+of\s+|/)
+                (?P<year>\d{{4}})
+            )
+        ''', re.X | re.I)
+
+    xpath_contains = " or ".join([f"contains(translate(., '{subtype.upper()}', '{subtype.lower()}'), "
+                                  f"'{subtype.lower()}')"
+                                  for subtype in subtype_names + subtype_abbreviations])
+    candidate_xpath = f".//text()[{xpath_contains} and not(ancestor::a:ref)]"
```
longhotsummer: wrap the contains clause in brackets in the xpath expression.

indigo/analysis/refs/base.py
```diff
@@ -63,6 +64,51 @@ class RefsFinderENG(BaseRefsFinder):
...
+    subtypes = [s for s in Subtype.objects.all()]
+    subtype_names = [s.name for s in subtypes]
+    subtype_abbreviations = [s.abbreviation for s in subtypes]
+
+    subtypes_string = '|'.join([re.escape(s) for s in subtype_names + subtype_abbreviations])
+    pattern_re = re.compile(
+        fr'''
+            (?P<ref>
+                (?P<subtype>{subtypes_string})\s*
+                ([nN]o\.?\s*)?
```
longhotsummer: You don't need both `n` and `N` if you're using the `re.I` (ignore case) flag.

indigo/analysis/refs/base.py
```diff
@@ -78,26 +78,36 @@ class RefsFinderSubtypesENG(BaseRefsFinder):
...
+        self.get_subtypes()
+        self.get_candidate_xpath()
+        self.get_pattern_re()
+        super().setup(root)
 
-    xpath_contains = " or ".join([f"contains(translate(., '{subtype.upper()}', '{subtype.lower()}'), "
-                                  f"'{subtype.lower()}')"
-                                  for subtype in subtype_names + subtype_abbreviations])
-    candidate_xpath = f".//text()[{xpath_contains} and not(ancestor::a:ref)]"
+    def get_subtypes(self):
```
longhotsummer: A style comment: methods name `get_` usually don't have side effects (eg. setting attributes), they normally just return a value. In this case, since you're setting things, a better name would be `setup_`.

PR #883, opened 2019-12-12
Completeness Level

### LINKED ISSUE #877: As an advanced user I want to see the completeness level of a work in the work listing page
Include the completeness ball in the work listing table, just to the left of the "last updated at" column. It doesn't need to be included for documents, only for works. It doesn't need a percentage, just the SVG graphic.

Also, add an advanced filter to show "Complete" and "Incomplete" works.

ro6ley:
#### What does this PR do?

Add completeness level of a work in the work listing page.

This PR adds the completeness ball(just the SVG graphic) in the work listing table, just to the left of the "last updated at" column. 

It also adds an advanced filter to show "Complete" and "Incomplete" works.

Open a place and click on the works tab. The progress balls should be visible next to the last updated column. 
Click on advanced filters to filter by 'Complete and incomplete works'

closes #877 

#### Screenshots

image

### DIFF
--- a/indigo_app/templates/place/works.html
+++ b/indigo_app/templates/place/works.html
@@ -104,11 +104,11 @@
                     {% endif %}
                   </div>
 
-                  {% if work.metrics %}
-                    <div class="col-1">
+                  <div class="col-1">
+                    {% if work.metrics %}
                       {% include 'indigo_metrics/_progress_ball.html' with size=40 progress=work.metrics.p_breadth_complete %}
-                    </div>
-                  {% endif %}
+                    {% endif %}
+                  </div>
 
                   <div class="col-2">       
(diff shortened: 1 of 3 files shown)


indigo_app/templates/place/works.html
```diff
@@ -104,7 +104,13 @@ <h3 class="mb-0">{{ latest_completeness_stat.n_complete_works }} of {{ latest_co
                     {% endif %}
                   </div>
 
-                  <div class="col-2">
+                  {% if work.metrics %}
+                    <div class="col-1">
```
longhotsummer: It's better to always include the column (ie. move the `if` inside the `div`), otherwise in the case where there are no metrics, that row will have the incorrect number of columns and it won't line up with the other rows.

PR #885, opened 2019-12-13
include locality in FRBR URI if there is one

goose-life:


goose-life:
image

indigo/analysis/refs/base.py
```diff
@@ -117,6 +117,9 @@ def make_href(self, match):
                 subtype = s.abbreviation
                 break
 
+        if self.frbr_uri.locality:
```
longhotsummer: Looks good. In the spirit of DRY, it's better to build up the URI progressively, so that we don't have to double-check for typos (or make changes later) in two places:

```python
href = None
if self.frbr_uri.locality:
    href = f'/{...}-{...}'
else:
    href = f'/{...}'

return f'{href}/act/....'
```

PR #886, opened 2019-12-17
User Tasks View

### LINKED ISSUE #881: as a manager I want to see who is currently working on what
Use case: I'm starting a new municipality and want to get an idea of who has which tasks currently assigned to them.

Suggestion: Include 'Tasks currently assigned to {user}' on their profile, as a separate tab / above 'Recent activity'

ro6ley:
#### What does this PR do?

This PR adds functionality for users to view what other users are working on.

This PR adds a new tab next to Activity that reads Tasks and includes all the info about that user from whatever is in their 'My tasks' view

- Set up and start indigo.
- Navigate to http://127.0.0.1:8000/contributors/
- Open a user's profile and select the `Tasks` section

closes #881 

#### Screenshots (if appropriate)

User tasks view with tasks:
image

User tasks view with no tasks and no full name:
image

### DIFF
--- a/indigo_app/templates/indigo_app/tasks/_user_tasks.html
+++ b/indigo_app/templates/indigo_app/tasks/_user_tasks.html
@@ -1,6 +1,6 @@
 
 <div class="row">
-  <div class="col-md-4">
+  <div class="{% if current_wide %}col-md-6{% else %}col-md{% endif %}">
     <div class="card task-list-card">
       <div class="card-body">
         <h6 class="text-center mb-3">
@@ -27,7 +27,7 @@
     </div>
   </div>
 
-  <div class="col-md-4">
+  <div class="col-md">
     <div class="card task-list-card">
       <div class="card-body">
         <h6 class="text-center mb-3">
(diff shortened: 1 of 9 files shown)


goose-life:
this looks right, but I wonder, is there a way we can {% include %} a template in  indigo_social/templates/indigo_social/user_tasks as well as in indigo_app/templates/indigo_app/tasks/my_tasks to avoid code duplication? and also change `MyTasksView` in indigo_app/views/tasks to be more generic and use that instead of `UserTasksView`?

ro6ley:
@goose-life I had thought about that, but it meant making a few changes to the `My Tasks` view so that it can look uniform and fit in the profile section. Let me try making it one

ro6ley:
@goose-life I found a workaround regarding the reuse of `MyTasksView` . Only major thing is, the dimensions of the cards are now the same size since the same view is used in the profile. 

This is how My Tasks looks like right now:
image

This is how the same view looked before:
image

You can see the difference in sizes of the columns. I coulnd't maintain the same sizing since the profile view is smaller.

goose-life:
Thanks for looking into this! I don't dislike the cards as equal thirds, but I didn't make that design decision in the first place so I can't overturn it now. I guess GK can decide which he prefers when he's back next year. Would you mind pushing the commit that includes your workaround in the meantime? no rush on that though :)

ro6ley:
@goose-life I have pushed the commits

PR #896, opened 2020-01-03
lock in django-allauth below 0.41.0

goose-life:


goose-life:
merging this now so we can all stop doing it in separate PRs

PR #898, opened 2020-01-03
add Cap refs finder

goose-life:
image

### DIFF
--- a/indigo/analysis/refs/base.py
+++ b/indigo/analysis/refs/base.py
@@ -153,7 +153,7 @@ class RefsFinderCapENG(BaseRefsFinder):
         country = document.work.country
         locality = document.work.locality
         place = locality or country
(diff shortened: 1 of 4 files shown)


indigo/analysis/refs/base.py
```diff
@@ -120,6 +120,45 @@ def make_href(self, match):
...
+
+    def setup(self, root):
+        self.setup_cap_numbers(self.document)
+        super().setup(root)
+
+    def setup_cap_numbers(self, document):
+        country = document.work.country
+        locality = document.work.locality
+        place = locality or country
+        cap_strings = [p for p in place.settings.work_properties if 'cap' in p]
```
longhotsummer: this should probably be `p.startswith('cap')` otherwise `foocapfoo` will match.

PR #904, opened 2020-01-10
Multiple commencements

### LINKED ISSUE #260: Need a mechanism to capture commencement details (notice, etc.) when sections of acts commence at different times.
eg.

Amended by "Fertilizers, Farm Feeds, Agricultural Remedies and Stock Remedies  Amendment Act 24 of 1977 (RSA) (RSA GG 5461)", sections 1-6, 8-14 and 16-22 brought into force on 10 June 1977  by RSA Proc. R.119/1977 (RSA GG 5583); section 7 brought into force  on 1 December 1977 by RSA Proc. R.311/1977 (RSA GG 5800)

image

goose-life:
will close #260


indigo_api/models/works.py
```diff
@@ -411,6 +463,31 @@ def save(self, *args, **kwargs):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    created_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+    updated_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+
+    class Meta:
+        ordering = ['date']
+        unique_together = (('commenced_work', 'commencing_work', 'date'),
+                           ('commenced_work', 'main'))
```
longhotsummer: I don't think this will work -- surely we can have duplicate Commencement objects for a work with `main=False`?

indigo_api/migrations/0125_migrate_commencements.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def for_document(topic, document):
+        country_code = document.work.country.country.iso.lower()
+        locality = document.work.locality.code if document.work.locality else None
+        language_code = document.language.language.iso_639_2B
+        return plugins.for_locale(topic, country=country_code, locality=locality, language=language_code)
+
+    def table_of_contents_for_document(self, document):
+        def make_act(xml):
+            id = re.sub(r'[^a-zA-Z0-9]', '-', settings.INDIGO_ORGANISATION)
```
longhotsummer: this `id` stuff isn't necessary for the TOC, so you can drop it.

indigo_api/migrations/0125_migrate_commencements.py
```diff
@@ -0,0 +1,95 @@
...
+            doc = Act(xml)
+            doc.source = [settings.INDIGO_ORGANISATION, id, settings.INDIGO_URL]
+            return doc
+        doc = make_act(document.document_xml)
+        django_language = document.language.language.iso_639_1
+        return self.table_of_contents(doc, django_language)
+
+    def table_of_contents(expression):
+        builder = for_document('toc', expression)
```
longhotsummer:
```
builder = ...
expression.doc = Act(expression.document_xml)
expression.django_language = expression.language.language.iso_639_1
return build.table_of_contents_for_document(expression)
```

Then I think most of the stuff above this block can go.

indigo/bulk_creator.py
```diff
@@ -382,19 +381,27 @@ def link_publication_document(self, work, info):
...
-        work.commencing_work = commencing_work
-        try:
-            work.save_with_revision(self.user)
-        except ValidationError:
-            self.create_task(work, info, task_type='link-commencement')
+        # if the work has commencement details, try linking it
+        # make a task if a `commenced_by` title is given but not found
+        commencing_work = None
+        date = info.get('commencement_date')
+        all_provisions = work.all_provisions()
```
longhotsummer: this will probably be empty since no points in time exist, yet. That's okay, just double checking.

indigo_api/models/works.py
```diff
@@ -88,20 +88,23 @@ class Meta:
     publication_number = models.CharField(null=True, blank=True, max_length=255, help_text="Publication's sequence number, eg. gazette number")
     publication_date = models.DateField(null=True, blank=True, help_text="Date of publication")
 
-    commenced = models.BooleanField(null=False, default=False, help_text="Has this work commenced? (Date may be unknown)")
-    commencement_date = models.DateField(null=True, blank=True, help_text="Date of commencement unless otherwise specified")
     assent_date = models.DateField(null=True, blank=True, help_text="Date signed by the president")
 
+    # comm
```
longhotsummer:
```suggestion
    # commencement
```

longhotsummer: `commencement_date` and `commencing_work` fall away, I think?

indigo_api/models/works.py
```diff
@@ -359,6 +362,55 @@ def as_at_date(self):
         if dates:
             return max(dates)
 
+    def all_provisions(self):
```
longhotsummer: I suggest this function simply finds the first expression and then delegates to the same method on the expression document, which then introspects the TOC etc. Provisions are inherently tied to a point in time.

indigo_api/models/works.py
```diff
@@ -359,6 +362,55 @@ def as_at_date(self):
...
+        # a list of the element ids of the earliest PiT (in whichever language) of a work
+        # does not include Schedules as they don't have commencement dates
+        ids = []
+
+        def add_ids(toc):
+            for e in toc:
+                if e.id:
```
longhotsummer: Do we need to only include `section` or the tradition-specific equivalent here?

indigo_api/models/works.py
```diff
@@ -359,6 +362,55 @@ def as_at_date(self):
...
+
+        return None
+
+    def main_commencement_date(self):
+        if self.commencements.exists():
+            main = self.commencements.filter(main=True).first()
+            if main:
+                return main.date
+
+            first = self.commencements.first()
```
longhotsummer: I think we should return None if there isn't a main commencement date.

indigo_api/models/works.py
```diff
@@ -359,6 +362,55 @@ def as_at_date(self):
...
+    def main_commencement(self):
+        if self.commencements.exists():
+            main = self.commencements.filter(main=True).first()
+            if main:
+                return main
+
+        return None
+
+    def main_commencement_date(self):
+        if self.commencements.exists():
```
longhotsummer: This goes to the database twice (.exists() and then .first()) -- instead, just use .first() and check for None.

These methods should use `main_commencement()`

indigo_api/models/works.py
```diff
@@ -359,6 +362,55 @@ def as_at_date(self):
...
+
+        return None
+
+    def main_commencing_work(self):
+        if self.commencements.exists():
+            main = self.commencements.filter(main=True).first()
+            if main:
+                return main.commencing_work
+
+            first = self.commencements.first()
```
longhotsummer: As above, I think at this point we give up and return None

indigo_api/models/works.py
```diff
@@ -411,6 +463,31 @@ def save(self, *args, **kwargs):
...
 
 
+class Commencement(models.Model):
+    """ The commencement details of (provisions of) a work,
+    optionally performed by a commencing work or a provision of the work itself.
+
+    """
+    commenced_work = models.ForeignKey(Work, on_delete=models.CASCADE, null=False, help_text="Principal work being commenced", related_name="commencements")
+    commencing_work = models.ForeignKey(Work, on_delete=models.CASCADE, null=True, help_text="Work that provides the commencement date for the principal work", related_name="commencements_made")
+    date = models.DateField(null=True, blank=True, help_text="Date of the commencement")
```
longhotsummer:
```suggestion
    date = models.DateField(null=True, blank=True, help_text="Date of the commencement, or null to indicate explicity uncommenced.")
```

indigo_app/forms.py
```diff
@@ -114,6 +119,26 @@ def save_publication_document(self):
...
+            main_commencement.delete()
+
+        # otherwise, amend the existing one (or create a new one) with the work / date given in the form
+        else:
+            if 'commencement_date' in self.changed_data or 'commencing_work' in self.changed_data:
+                if not main_commencement:
+                    all_provisions = self.instance.all_provisions()
+                    main_commencement = Commencement(commenced_work=self.instance, main=True, provisions=all_provisions)
+
+                main_commencement.commencing_work = self.cleaned_data.get('commencing_work')
```
longhotsummer: Just sanity check that it does what we expect if one of `commencing_work` or `commencement_date` are None.

indigo_api/migrations/0125_migrate_commencements.py
goose-life: don't I need it to make `doc` so that `.table_of_contents()` on l36 works as expected?

goose-life: nvm, just read the next comment

indigo_api/models/works.py
goose-life: I'd like to keep Chapters and Parts for now too – would be cool if we can use them in the form (e.g. select Chapter 2 and it automatically selects all the sections up to the start of Chapter 3)

indigo/bulk_creator.py
```diff
@@ -382,19 +381,26 @@ def link_publication_document(self, work, info):
...
+        # if the work has commencement details, try linking it
+        # make a task if a `commenced_by` title is given but not found
+        date = info.get('commencement_date')
+
+        commencing_work = None
+        title = info.get('commenced_by')
+        if title:
+            commencing_work = self.find_work_by_title(title)
+            if not commencing_work:
+                self.create_task(work, info, task_type='link-commencement')
```
longhotsummer: Do we need to return here?

indigo/bulk_creator.py
```diff
@@ -382,19 +381,26 @@ def link_publication_document(self, work, info):
...
+
+        commencement = Commencement(
+            commenced_work=work,
+            commencing_work=commencing_work,
+            date=date,
```
longhotsummer: Is it okay if `date` is None?

indigo/bulk_creator.py
```diff
@@ -405,7 +411,7 @@ def link_repeal(self, work, info):
         if not repealing_work:
             return self.create_task(work, info, task_type='link-repeal')
 
-        repeal_date = repealing_work.commencement_date
+        repeal_date = repealing_work.main_commencement_date()
```
longhotsummer: I'm beginning to wonder if we should simply make `commencement_date` an alias for `main_commencement_date()`?

indigo_api/migrations/0125_migrate_commencements.py
```diff
@@ -0,0 +1,84 @@
...
+    for w in Work.objects.using(db_alias).all():
+        if w.commencing_work or w.commencement_date:
+            all_provisions = get_all_provisions(w)
+            commencement = Commencement(
+                commenced_work=w,
+                commencing_work=w.commencing_work,
+                date=w.commencement_date,
+                main=True,
+                all_provisions=True,
+                provisions_list=all_provisions,
```
longhotsummer: If we're introducing logic somewhere else that will "expand" the all provisions list, we could skip doing it here in the migration and simply delegate to that. Just a thought for now.

indigo_api/models/documents.py
```diff
@@ -285,7 +285,7 @@ def expression_frbr_uri(self):
 
     @property
     def commencement_date(self):
-        return self.work.commencement_date
+        return self.work.main_commencement_date()
```
longhotsummer: We could do something similar on the work model, then it would retain the old API and we wouldn't have to change all the references to `work.commencement_date`.

indigo_api/models/works.py
```diff
@@ -359,6 +350,44 @@ def as_at_date(self):
...
+    def all_provisions(self):
+        # a list of the element ids of the earliest PiT (in whichever language) of a work
+        # does not include Schedules as they don't have commencement dates
+        first_expression = self.expressions().first()
+        if first_expression:
+            return first_expression.all_provisions()
+        return None
+
+    def main_commencement(self):
+        main = self.commencements.filter(main=True).first()
```
longhotsummer: `return self.commencements.filter(main=True).first()`

indigo_api/models/works.py
```diff
@@ -359,6 +350,44 @@ def as_at_date(self):
...
+        if main:
+            return main
+        return None
+
+    def main_commencement_date(self):
+        main = self.main_commencement()
+        if main:
+            return main.date
+        return None
```
longhotsummer: A python method returns None if it doesn't have an explicit return, so it's not strictly necessary to have `return None` at the end.

indigo_api/models/works.py
```diff
@@ -411,6 +440,52 @@ def save(self, *args, **kwargs):
         return super(PublicationDocument, self).save(*args, **kwargs)
 
 
+class Commencement(models.Model):
+    """ The commencement details of (provisions of) a work,
+    optionally performed by a commencing work or a provision of the work itself.
```
longhotsummer: We don't currently explicitly support a work commencing itself, right?

indigo_api/models/works.py
```diff
@@ -411,6 +440,52 @@ def save(self, *args, **kwargs):
...
+    updated_at = models.DateTimeField(auto_now=True)
+
+    created_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+    updated_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+
+    class Meta:
+        ordering = ['date']
+        unique_together = (('commenced_work', 'commencing_work', 'date'),)
+
+    def commences(self):
```
longhotsummer: Safe to make this a `@property`

indigo_api/models/works.py
```diff
@@ -411,6 +440,52 @@ def save(self, *args, **kwargs):
...
+        return bool(self.date)
+
+    def only(self):
```
longhotsummer: It feels like this needs a slightly longer/more descriptive name (and probably a documentation comment). Perhaps `is_only_commencement`?

indigo_api/models/works.py
```diff
@@ -411,6 +440,52 @@ def save(self, *args, **kwargs):
...
+        return bool(
+            self.main is True and
+            self.all_provisions is True and
+            not self.other_commencements()
```
longhotsummer:
```suggestion
            not self.other_commencements().exists()
```

indigo_api/serializers.py
```diff
@@ -531,12 +530,26 @@ def create(self, validated_data):
         return super(DocumentActivitySerializer, self).create(validated_data)
 
 
+class CommencementSerializer(serializers.ModelSerializer):
```
longhotsummer: Nice, easy peasy :)

indigo_api/serializers.py
```diff
@@ -550,9 +563,9 @@ class Meta:
             # readonly, url is part of the rest framework
             'id', 'url',
             'title', 'publication_name', 'publication_number', 'publication_date', 'publication_document',
-            'commenced', 'commencement_date', 'assent_date', 'stub',
+            'commenced', 'assent_date', 'stub',
```
longhotsummer: I'm looking at how useful it is for us in our code to have easy access to the commencement date (and possibly other details) of the main commencement. I wonder if we should retain these fields in the API and populate them from the main commencement object. It would make the API backwards compatible, and make working with the commencement common case (commenced, single work), much simpler. Otherwise, an API user must iterate over the commencement objects, find the main one, and work with that.

indigo/bulk_creator.py
goose-life: we've changed the logic slightly: 
previously, this was only for linking the commencing work, which is why we had a return here.
now, this is the only place a commencement object is created.
we could change it so that if there's no `commencing_work` and no `date`, we return `here`; but if there's no `commencing_work` and there is a `date`, we create a task here but continue to create the commencement object. 
but if we don't find the commencing_work and there is no date, and we create a shell of a commencement object, that's also not the end of the world? it'll get edited once the task is done, or deleted if the work is marked as uncommenced after all.
lmk if you'd like me to add the extra logic, I don't think it's necessary but if you'd prefer to be cautious we can do it!

indigo/bulk_creator.py
goose-life: yep, the date is allowed to be None: 
`date = models.DateField(null=True, blank=True, help_text="Date of the commencement, or null if listed provisions are uncommenced")`

goose-life: we're obviously not _expecting_ a work to have a commencing work but no date on a bulk import, but if it does happen the work will be flagged as 'commencement date unknown' and we take it from there

indigo_api/models/works.py
```diff
@@ -411,6 +444,82 @@ def save(self, *args, **kwargs):
...
 
 
+class Commencement(models.Model):
+    """ The commencement details of (provisions of) a work,
+    optionally performed by a commencing work or a provision of the work itself.
+
+    """
+    commenced_work = models.ForeignKey(Work, on_delete=models.CASCADE, null=False, help_text="Principal work being commenced", related_name="commencements")
+    commencing_work = models.ForeignKey(Work, on_delete=models.CASCADE, null=True, help_text="Work that provides the commencement date for the principal work", related_name="commencements_made")
+    date = models.DateField(null=True, blank=True, help_text="Date of the commencement, or null if listed provisions are uncommenced")
```
longhotsummer:
```suggestion
    date = models.DateField(null=True, blank=True, help_text="Date of the commencement, or null if it is unknown")
```

indigo_api/models/works.py
```diff
@@ -411,6 +444,82 @@ def save(self, *args, **kwargs):
...
+        if existing_all_provisions_commencement and existing_all_provisions_commencement != self:
+            self.all_provisions = False
+
+        return super(Commencement, self).save(*args, **kwargs)
+
+
+class UncommencedProvisions(models.Model):
+    """ The details of uncommenced provisions of a work
+    """
+    commenced_work = models.OneToOneField(Work, on_delete=models.CASCADE, null=False, help_text="Principal work with uncommenced provisions", related_name="uncommenced_provisions")
```
longhotsummer:
```suggestion
    work = models.OneToOneField(Work, on_delete=models.CASCADE, null=False, help_text="Principal work with uncommenced provisions", related_name="uncommenced_provisions")
```

indigo_api/models/works.py
```diff
@@ -411,6 +444,82 @@ def save(self, *args, **kwargs):
...
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+
+    created_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+    updated_by_user = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name='+')
+
+    def save(self, *args, **kwargs):
+        # ensure only one commencement / uncommencement with all_provisions=True on commenced work
+        if self.all_provisions:
+            existing_all_provisions_commencement = self.commenced_work.commencements.filter(all_provisions=True).first()
```
longhotsummer:
```suggestion
            existing_all_provisions_commencement = self.work.commencements.filter(all_provisions=True).exists()
```

longhotsummer: From a style standpoint, the long variable name may be descriptive but it's long to read and long to type. It could safely be `existing` or `commencement`.

longhotsummer: Faster to check `.exists()` than to load from the database.

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -223,7 +223,7 @@ <h5 class="card-header">Publication</h5>
 
     {% block work-form-commencement %}
     <div class="card mb-3" id="commencement">
-      <h5 class="card-header">Commencement and Assent</h5>
+      <h5 class="card-header">Commencement and assent</h5>
```
longhotsummer: One day, these will all be correct. Dare to dream.

indigo_api/models/works.py
longhotsummer: Rather cache this method, than the "alias" property. This is the one that goes to the DB.

goose-life:
> Rather cache this method, than the "alias" property. This is the one that goes to the DB.

Were my changes what you had in mind?

indigo_api/migrations/0125_migrate_commencements.py
goose-life: yeah this should actually come out, I realised recently getting the first PiT for the ToC won't necessarily be right because we don't always have a document for the first PiT (e.g. starting with a consolidation). 

Can we make it so that only works with multiple commencements need `provisions` for their Commencement objects, and that these must be entered manually (at least to start)?

indigo_api/models/works.py
```diff
@@ -462,14 +462,30 @@ class Meta:
...
-        existing_all = self.commenced_work.commencements.filter(all_provisions=True).first()
-        if existing_all and existing_all != self:
+        # if more than one commencement / uncommencement exists on the work, all_provisions is False on all of them
+        other_commencements = work.commencements.exclude(pk=self.pk)
+        uncommenced = work.uncommenced_provisions
+
+        if other_commencements or uncommenced:
+            self.all_provisions = False
+
+        for commencement in other_commencements:
```
longhotsummer: On second thought, I think we should move this logic into the view where we update a commencement object. Otherwise we're going to be chasing our tails, as each `.save()` triggers this logic again and again.

Perhaps as a rule of thumb, `.save()` should only adjust the current object. Adjusting other objects should happen at a higher level, where the code has a better "view" of all the changes being made to a commencement object.

indigo_api/models/works.py
```diff
@@ -462,14 +462,30 @@ class Meta:
...
+            if commencement.all_provisions:
+                commencement.all_provisions = False
+                commencement.save()
+
+        if uncommenced and uncommenced.all_provisions:
```
longhotsummer: As above.

indigo_api/models/works.py
```diff
@@ -462,14 +462,30 @@ class Meta:
...
+            uncommenced.all_provisions = False
+            uncommenced.save()
+
+        # if it has a provisions list, all_provisions must be False
+        if self.provisions:
```
longhotsummer: This is safe and can stay.

goose-life:
commencements on timeline:

image

goose-life:
multiple commencements in work overview page summary:

image

goose-life:
work popup:

image

indigo_api/models/works.py
```diff
@@ -331,16 +337,18 @@ def points_in_time(self):
...
 
-        for date, group in groupby(amendments_expressions, key=lambda x: x.date):
+        for date, group in groupby(events_expressions, key=lambda x: x.date):
             group = list(group)
             pits.append({
                 'date': date,
                 'initial': any(getattr(a, 'initial', False) for a in group),
-                'amendments': group,
-                'expressions': set(chain(*(a.expressions().all() for a in group))),
+                'amendments': [a for a in group if a._meta.object_name == 'Amendment'],
```
longhotsummer: Could also do `if instanceof(a, Amendment)` which is the pythonic way to test if an object has a particular class.

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -73,6 +48,17 @@ <h6>Repealed
           </h6>
         {% endif %}
 
+        {% for commencement in event.commencements %}
+          <h6>
+            Commenced{% if not commencement.all_provisions %} in part{% endif %}
+            {% with commencement.commencing_work as commencing_work %}
```
longhotsummer: There's no good reason to stash this with `with`.

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -73,6 +48,17 @@ <h6>Repealed
           </h6>
         {% endif %}
 
+        {% for commencement in event.commencements %}
+          <h6>
+            Commenced{% if not commencement.all_provisions %} in part{% endif %}
```
longhotsummer: I like the 'in part' wording, we probably want to re-use that elsewhere.

indigo_app/templates/indigo_api/work_commencements.html
```diff
@@ -0,0 +1,157 @@
...
+  {% if has_all_provisions %}
+    <p>All provisions have commenced, no additional commencement events are required.</p>
+  {% endif %}
+
+  <ol class="timeline">
+    {% if perms.indigo_api.add_commencement and not has_all_provisions %}
+      <li class="timeline-item card">
+        <div class="card-body">
+          {% block new-commencement-card %}
+            <button class="btn btn-primary add-commencement"><i class="fas fa-plus"></i> Add commencement</button>
```
longhotsummer: We can add a dropdown here to create a commencement without a commencing work.

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -60,15 +60,16 @@ <h4 class="text-muted">{{ work.frbr_uri }}</h4>
...
-                      Commenced on
-                      {% if work.commencement_date %}
-                        {{ work.commencement_date|date:"Y-m-d" }}
-                      {% else %}
-                        <i>(unknown)</i>
-                      {% endif %}
-                      {% if work.commencing_work %}
-                        by <a href="{% url 'work' frbr_uri=work.commencing_work.frbr_uri %}" data-popup-url="{% url 'work_popup' frbr_uri=work.commencing_work.frbr_uri %}">{{ work.commencing_work.title }}</a>
-                      {% endif %}
+                      {% with work.commencements.count as n_commencements %}
```
longhotsummer: `.count()` will go to the db, you can also use `work.commencements|length` which will test the length of the object, which will be evaluated once. Although it's moot because you use the variable, so this is just an FYI.

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -60,15 +60,16 @@ <h4 class="text-muted">{{ work.frbr_uri }}</h4>
...
+                        {% if n_commencements > 1 %}Main commencement: {% endif %}
```
longhotsummer: Is the "main" really adding anything? Can we not just have "Commenced on XXX" and then the supplementary notice if there are other commencements? The "main" in that case is implied. Let's chat through this...

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -60,15 +60,16 @@ <h4 class="text-muted">{{ work.frbr_uri }}</h4>
...
+                        Commenced on {{ work.commencement_date|date:"Y-m-d"|default:"<i>(unknown)</i>" }}
+                          {% if work.commencing_work %}
+                            by <a href="{% url 'work' frbr_uri=work.commencing_work.frbr_uri %}" data-popup-url="{% url 'work_popup' frbr_uri=work.commencing_work.frbr_uri %}">{{ work.commencing_work.title }}</a>
+                          {% endif %}
+                        {% if n_commencements > 1 %}
+                          <div>This work has mixed commencements, see below for details or <a href="{% url 'work_commencements' frbr_uri=work.frbr_uri %}">edit them in the Commencements page</a>.</div>
```
longhotsummer:
```suggestion
                          <div>This work has <a href="{% url 'work_commencements' frbr_uri=work.frbr_uri %}">multiple commencements</a>.</div>
```

indigo_app/templates/indigo_api/work_popup.html
```diff
@@ -42,10 +42,15 @@ <h6>
...
 
-  {% if work.commencement_date %}
-    <p class="mb-2">Commenced on {{ work.commencement_date|date:"Y-m-d" }}{% if work.commencing_work %} by
-      <a href="{% url 'work' frbr_uri=work.commencing_work.frbr_uri %}">
-        {{ work.commencing_work.title }}</a>{% endif %}.
+  {% if work.commenced %}
+    <p class="mb-2">
+      {% if work.commencements.count > 1 %}
+        This work has <a href="{% url 'work_commencements' frbr_uri=work.frbr_uri %}">multiple commencements</a>.
+      {% else %}
```
longhotsummer: If we have a main commencement date, we should always show it. We can then supplement that with the multiple commencements notice.

indigo_app/views/works.py
```diff
@@ -59,23 +60,28 @@ def get_work_timeline(self):
...
-            for name, date in other_dates:
-                if entry['date'] == date:
-                    entry[name] = True
-        # add new events (e.g. if assent is before any of the other events)
-        existing_dates = [entry['date'] for entry in work_timeline]
+
+        # add new events (e.g. if assent is before any of the other events),
+        # only one event per date
+        def existing_dates():
+            return [entry['date'] for entry in work_timeline]
```
longhotsummer: If a function is one line (and in this case a simple list comprehension), it's okay not to create a whole new function. You could easily just put this directly into the if statement.

longhotsummer: This would all be way simpler if work_timeline was a dictionary mapping dates to descriptions!

indigo_app/views/works.py
```diff
@@ -238,6 +244,153 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class WorkCommencementsView(WorkViewBase, DetailView):
+    template_name_suffix = '_commencements'
+    tab = 'commencements'
+
+    def get_context_data(self, **kwargs):
+        context = super(WorkCommencementsView, self).get_context_data(**kwargs)
+        # TODO: these need to be attached and filtered for each commencement object,
```
longhotsummer: Whoops, this comment can be removed.

indigo_app/views/works.py
```diff
@@ -480,9 +633,9 @@ def get_context_data(self, **kwargs):
             })
         commencement = commencement + [{
             'rel': 'commenced',
-            'work': w,
-            'date': w.commencement_date,
-        } for w in self.work.commenced_works.all()]
+            'work': c.commenced_work,
```
longhotsummer: In addition, we need to include all the works that commenced this one, currently we just pull in the primary one.

PR #908, opened 2020-01-13
Update pinax-badges

longhotsummer:
This is a fix for https://github.com/pinax/pinax-badges/issues/27


longhotsummer:
This will allow Indigo to run on python 3.7+

PR #915, opened 2020-01-21
Move stubs filter out of advanced, change wording for "all works"

longhotsummer:
* This makes it clearer that we're hiding stubs by default.
* Move "published / draft" down into advanced
* "All works" -> "All types"
* "Acts only" -> "Act"

South Africa – Works – Laws Africa 2020-01-21 11-57-56

South Africa – Works – Laws Africa 2020-01-21 11-58-11

### DIFF
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -241,7 +241,7 @@ class TaskFilterForm(forms.Form):
 class WorkFilterForm(forms.Form):
     q = forms.CharField()
     stub = forms.ChoiceField(choices=[('', 'Exclude stubs'), ('only', 'Only stubs'), ('all', 'Everything')])
-    status = forms.MultipleChoiceField(choices=[('published', 'published'), ('draft', 'draft')])
+    status = forms.MultipleChoiceField(choices=[('published', 'published'), ('draft', 'draft')], initial=['published', 'draft'])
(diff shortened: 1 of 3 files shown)


indigo_app/forms.py
```diff
@@ -272,19 +272,25 @@ class WorkFilterForm(forms.Form):
...
 
     advanced_filters = ['assent', 'publication', 'repeal', 'amendment', 'commencement',
-                        'primary_subsidiary', 'taxonomies', 'stub', 'completeness']
+                        'primary_subsidiary', 'taxonomies', 'completeness', 'status']
 
     def __init__(self, country, *args, **kwargs):
         self.country = country
         super(WorkFilterForm, self).__init__(*args, **kwargs)
-        self.fields['subtype'] = forms.ChoiceField(choices=[('', 'All works'), ('acts_only', 'Acts only')] + [(s.abbreviation, s.name) for s in Subtype.objects.all()])
+        self.fields['subtype'] = forms.ChoiceField(choices=[('', 'All types'), ('acts_only', 'Act')] + [(s.abbreviation, s.name) for s in Subtype.objects.all()])
```
goose-life: I realise this is a bit of an edge case, but just for the record I had it as 'Acts only' instead of 'Act' for when someone adds 'Act' as a subtype (which did happen with one installation).

longhotsummer:
```suggestion
        self.fields['subtype'] = forms.ChoiceField(choices=[('', 'All types'), ('acts_only', 'Acts only')] + [(s.abbreviation, s.name) for s in Subtype.objects.all()])
```

PR #917, opened 2020-01-22
Place name on coverpage

### LINKED ISSUE #857: Include place name in generated work coverpage
Unless there's a coat of arms, such as in NA.

If we do this by, say, adding an `h3` tag just above the act title, then the generated PDF picks up that content as (a) as the first entry in the TOC and (b) as the running header for the whole document. So this needs to be done without an `h3` tag, but still needs to be styled appropriately by https://github.com/laws-africa/indigo-web

goose-life:
closes #857 

- Above Work title on coverpage
- Doesn't affect ToC or running heads in pdf

Document view on platform:
image

Pdf first page:
image

Pdf later page:
image

With locality:
image

Pdf first page:
image

Pdf later page:
image

### DIFF
--- a/indigo_api/templates/indigo_api/akn/coverpage_act.html
+++ b/indigo_api/templates/indigo_api/akn/coverpage_act.html
@@ -4,6 +4,10 @@
 <div class="coverpage">
   {% block coat_of_arms %}{% endblock %}
 
+  {% block place_name %}
+    <p class="place-name">{{ document.work.country.name }}{% if document.work.locality %} – {{ document.work.locality.name }}{% endif %}</p>
+  {% endblock %}
+
   {% if document.work.parent_work %}
     {% block parent_title %}
       <h1>{{ document.work.parent_work.title }}</h1>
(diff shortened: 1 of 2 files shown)


indigo_api/templates/indigo_api/akn/coverpage_act.html
```diff
@@ -4,6 +4,10 @@
 <div class="coverpage">
   {% block coat_of_arms %}{% endblock %}
 
+  {% block place_name %}
+    <p class="place-name">{{ document.work.country.name }}{% if document.work.locality %} – {{ document.work.locality.name }}{% endif %}</p>
```
longhotsummer: Another option would be `Cape Town, South Africa`, which I think "sounds" a bit better?

Alternatively: `Cape Town<br>South Africa`, which would look better when the locality name is long enough to wrap.

goose-life:
image

longhotsummer:
Oh, please just double check that the place name doesn't show up in the PDF TOC, and that the body pages of the PDF have the Act name in the top left corner, not the place name. That happened last time when I tried it with `h2` tags.

PR #946, opened 2020-02-19
Fix for `not repealed` filter bug on Works filter

ro6ley:
#### What does this PR do?

Fix a bug where the `not repealed` works filter returned repealed works.

The queryset filter when the `not repealed` option was selected fetched works with repeal dates. The fix is to filter the works where the repeal date is null. More details on file changed.

#937


longhotsummer:
Thanks!

PR #948, opened 2020-02-19
commencement activity

### LINKED ISSUE #943: as a manager I want to see who added/changed commencement information
We should record commencement-related changes on the work activity page just as we do amendment-related changes. Probably the same way as we do amendments.

* [ ] create the activity stream entries
* [ ] update updated_at and updated_by on the work object

goose-life:
closes #943 

Work changes:

image

Work reflects as recently updated:

image

### DIFF
--- a/indigo_api/apps.py
+++ b/indigo_api/apps.py
@@ -8,7 +8,7 @@ class IndigoApiConfig(AppConfig):
     def ready(self):
         from actstream import registry
         from django.contrib.auth.models import User
-        from indigo_api.models import Amendment, Document, Task, Work, Workflow, PlaceSettings, ArbitraryExpressionDate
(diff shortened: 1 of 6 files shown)


goose-life:
- We can make coalesce_entries smarter (currently the commencement object is created, then the work is saved, then the commencement object is updated, all of which end up on 'Work changes')
- As always with this stuff the whole entry disappears when a commencement object is deleted

goose-life:
would the best place for that be also in post_save?

I had it in but removed it because we're already updating the work during 'rationalise' – but we may not always end up updating the work (if it doesn't go from 'uncommenced' to 'commenced')

longhotsummer:
> would the best place for that be also in post_save?
> 
> I had it in but removed it because we're already updating the work during 'rationalise' – but we may not always end up updating the work (if it doesn't go from 'uncommenced' to 'commenced')

We generally manipulate those fields in the form or view logic, since at that point we know the user doing the manipulation.

PR #951, opened 2020-02-20
Fix expression dates dropdown

goose-life:
currently nothing is showing because that method was renamed:
image

but using the method as changed would incorrectly allow a user to change the expression date of a document to a date that has just a commencement, so I made some tweaks to get around that


longhotsummer:
We've decided not to fix it like this, instead we're going to adjust how we use both these methods. See https://github.com/laws-africa/indigo/issues/952

PR #960, opened 2020-02-25
resolve points in time / timeline confusion

### LINKED ISSUE #952: Bug: confusion between points_in_time and work expressions
`work.points_in_time()` and `work.amendments_initial_commencement_arbitrary()` are misbehaving and confusing.

* the expression date dropdown in the document view is broken
* metrics counts things incorrectly (n_points_in_time)

## Proposal

1. add `work.possible_expression_dates()` that returns a list of dicts with a date and a brief description, for use in the expression date dropdown and metrics
2. move logic for building the work timeline into the view itself

goose-life:
closes #952 

still to do:
- [ ] use possible_expression_dates() in metrics
- [ ] check what else uses points_in_time() and whether it can use possible_expression_dates() instead, or needs something else
- [ ] get rid of amendments_initial_commencement_arbitrary() and points_in_time() on Work

### DIFF
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -617,6 +617,11 @@ class ArbitraryExpressionDate(models.Model):
     def can_delete(self):
         return not self.expressions().exists()
 
+    def save(self):
+        if self.work.amendments.filter(date=self.date).all():
+            return
+        return super().save()
+
     @property
     def amended_work(self):
         return self.work
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 from itertools import chain, groupby
+from datetime import datetime
 
 from actstream import action
 from django.contrib.postgres.fields import JSONField
@@ -311,6 +312,27 @@ class Work(models.Model):
         plugin = plugins.for_work('work-detail', self)
         if plugin:
             return plugin.work_friendly_type(self)
+
+    def possible_expression_dates(self):
+        initial = self.publication_date or self.commencement_date
+        amendment_dates = [a.date for a in self.amendments.all()]
+        consolidation_dates = [c.date for c in self.arbitrary_expression_dates.all()]
+        dates = [
+            {'date': date,
+             'amendment': date in amendment_dates,
+             'consolidation': date in consolidation_dates,
+             'initial': date == initial}
+            for date in amendment_dates + consolidation_dates
+        ]
+
+        if initial not in amendment_dates + consolidation_dates:
+            dates = dates + [
+                {'date': initial,
+                 'initial': True}
+            ]
+
+        dates.sort(key=lambda x: x['date'], reverse=True)
+        return dates
 
     def amendments_initial_commencement_arbitrary(self):
         """ Return a list of Amendment, Commencement and ArbitraryExpressionDate objects, including a fake one at the end
--- a/indigo_app/views/works.py
+++ b/indigo_app/views/works.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 import json
 import logging
-from itertools import chain
+from itertools import chain, groupby
 from datetime import timedelta
 
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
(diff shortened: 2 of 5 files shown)


indigo_api/models/works.py
```diff
@@ -312,6 +313,27 @@ def friendly_type(self):
         if plugin:
             return plugin.work_friendly_type(self)
 
+    def possible_expression_dates(self):
```
longhotsummer: Docstring please -- in particular that it returns a list of dates in reverse order.

indigo_api/models/works.py
```diff
@@ -312,6 +313,27 @@ def friendly_type(self):
...
+        initial = self.publication_date or self.commencement_date
+        amendment_dates = [a.date for a in self.amendments.all()]
+        consolidation_dates = [c.date for c in self.arbitrary_expression_dates.all()]
+        dates = [
+            {'date': date,
+             'amendment': date in amendment_dates,
+             'consolidation': date in consolidation_dates,
+             'initial': date == initial}
+            for date in amendment_dates + consolidation_dates
```
longhotsummer: There could be duplicate dates. I suggest doing something like `all_dates = set(amendment_dates + consolidation_dates)`

indigo_api/models/works.py
```diff
@@ -312,6 +313,27 @@ def friendly_type(self):
...
+        ]
+
+        if initial not in amendment_dates + consolidation_dates:
+            dates = dates + [
```
longhotsummer: For style, a common way of formatting dicts that span multiple lines is to put the opening `{` and closing `}` on their own lines. It makes it easier to read and add/remove entries. A trailing comma on the last entry also makes it easier to add/shuffle lines around later without having to tweak commas. eg:

```
dates = dates + [{
    'date': initial,
    'initial': True,
}]
```

indigo_api/models/works.py
```diff
@@ -595,6 +617,11 @@ def expressions(self):
     def can_delete(self):
         return not self.expressions().exists()
 
+    def save(self):
```
longhotsummer: Are you trying to prevent this object from being created in certain circumstances?

Doing it like this is dangerous because `save()` will simply not happen in some cases, and silently at that. It breaks the contract of `Model.save()` and will have unforeseen impacts on other code.

If we need to prevent the creation, it should be part of a validation step wherever the creation happens (eg. in a form or view). It's too late to be doing it here because we can't stop it cleanly.

indigo_app/views/works.py
```diff
@@ -1,7 +1,7 @@
 # coding=utf-8
 import json
 import logging
-from itertools import chain
+from itertools import chain, groupby
```
longhotsummer: I think this is unused?

indigo_app/views/works.py
```diff
@@ -55,32 +55,48 @@ def get_context_data(self, **kwargs):
...
-                    for entry in work_timeline:
+                    for entry in dates:
                         if entry['date'] == date:
                             entry[name] = True
 
-        return sorted(work_timeline, key=lambda k: k['date'], reverse=True)
+        dates.sort(key=lambda x: x['date'], reverse=True)
+
+        # add expressions and commencement and amendment objects
+        work_timeline = [{
```
longhotsummer: Rather than copying everything from the other dict, would it be easier to:

```python
for event in dates:
    event.commencements = ...
    etc.
```

indigo_api/models/works.py
longhotsummer: Also, rather than concatenating two lists, `dates.append({...})` is better.

indigo_api/models/works.py
goose-life: Cool, I've canned this for now, it's not necessary. We probably don't want consolidations and amendments on the same date, but that's actually a separate issue which we can deal with if and when it actually arises.

indigo_app/views/works.py
goose-life: yep, thanks!

indigo_app/views/works.py
goose-life: um yes that's much better thank you!

goose-life:
@longhotsummer, can you confirm that https://github.com/laws-africa/indigo-lawsafrica/blob/master/indigo_liiweb/updater.py is using `points_in_time` from WorkExpression, in which case it's fine to leave as is?

PR #966, opened 2020-02-28
Remember finished session nonces across browser tabs

longhotsummer:
This fixes the issue where the window unload event cannot send its
DELETE call to the server. Instead, we remember that a session was finished
by storing the document id and nonce in the browser's localStorage.

The next time the same user opens that document in the same browser,
it tells the server that those old sessions are dead.

### DIFF
--- a/indigo_app/static/javascript/indigo/views/document_activity.js
+++ b/indigo_app/static/javascript/indigo/views/document_activity.js
@@ -50,13 +50,25 @@
         this.nonce = Math.floor(Math.random() * (max - min) + min).toString();
       }
 
+      // at the same time, we clean up finished sessions from other tabs that didn't get sent in time,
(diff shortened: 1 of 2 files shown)


indigo_app/static/javascript/indigo/views/document_activity.js
```diff
@@ -101,15 +113,51 @@
...
+
+      for (var i = 0; i < localStorage.length; i++) {
+        var key = localStorage.key(i);
+
+        if (key && key.startsWith('indigo-document-activity-finished-')) {
+          var data = localStorage.getItem(key);
+          if (data) {
+            try {
+              data = JSON.parse(data);
+            } catch {
```
longhotsummer:
```suggestion
            } catch(err) {
```

PR #972, opened 2020-03-08
FIX cap numbered title

longhotsummer:
Fixes "Chapter cap65 of 1840" to "Chapter 65 of 1840". I suspect it shouldn't even have "of 1840"?

This is pretty hard-coded right now. We could do better, like move it out of the base and until a regional plugin.

Move tests that should be in `indigo_api.tests` out of `indigo.tests`


goose-life:
I think we should re-look at this, we shouldn't be using `cap` in the number of a work at all anymore now that we have support for Cap number in different places, or is there a scenario I'm overlooking?

goose-life:
If I'm wrong we should also be guarding against e.g. a by-law with a number that starts with `cap`.

I also agree that the year shouldn't be included in the numbered title for Chapters.

goose-life:
Oh I see, there's just 3 in Namibia that have it. sigh.

indigo/analysis/work_detail/base.py
```diff
@@ -15,10 +15,14 @@ def work_numbered_title(self, work):
...
         This usually differs from the short title. May return None.
         """
-        number = work.number
+        uri = work.work_uri
+        number = uri.number
         if number == 'constitution':
             return None
-        uri = work.work_uri
+        # TODO: Should be in a locale-specific place
+        if number.startswith('cap'):
```
goose-life:
```suggestion
        if number.startswith('cap') and not uri.subtype:
```

goose-life: and switch the if / elif at lines 34-44?

longhotsummer:
I agree with you that we should ideally move to use the cap property logic we're using for Kenya. I'd really rather not have this in the global plugin.  Maybe we just move this to the NA plugin as a special case for Nam?

longhotsummer:
Closed in preference of doing this in a locale-specific plugin for NA.

PR #976, opened 2020-03-10
Judge completeness of works in a place by breadth completeness

longhotsummer:
We currently display breadth completeness everywher, ignoring depth
completeness (which requires a point-in-time for every work) for now.

We therefore want to calculate the daily number of complete works
based on breadth completeness, not just expressions.

This corrects, for example, the fact that in za-cpt it shows 59 of 70
works complete, but filtering on incomplete works shows only 4.

Also:

* log background task errors such that the admins are emailed when they fail
* protect against a None initial date for a work (this is breaking the metrics task)


longhotsummer:
In future, we could track `n_complete_works`, `n_breadth_complete_works` and `n_depth_complete_works` separately.

indigo_api/models/works.py
```diff
@@ -184,7 +184,7 @@ def possible_expression_dates(self):
             for date in all_dates
         ]
 
-        if initial not in all_dates:
+        if initial and initial not in all_dates:
```
goose-life: isn't it impossible to create a work without either a publication or a commencement date?

longhotsummer:
The database doesn’t enforce it, only the work form, and we have historical docs that seem to have this. 

Greg
--
(iPhone)

PR #977, opened 2020-03-10
Fix revisions bug

goose-life:
jsonpatch now uses json.dumps, so we need to turn our date objects into strings before getting the patch

### DIFF
--- a/indigo_app/revisions.py
+++ b/indigo_app/revisions.py
@@ -15,6 +15,13 @@ def decorate_versions(versions, ignore=IGNORE_FIELDS):
                 if fld in d:
                     del d[fld]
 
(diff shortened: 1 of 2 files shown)


indigo_app/revisions.py
```diff
@@ -15,6 +15,13 @@ def decorate_versions(versions, ignore=IGNORE_FIELDS):
                 if fld in d:
                     del d[fld]
 
+        for d in [curr_d, prev_d]:
+            for fld in d:
+                try:
+                    d[fld] = d[fld].isoformat()
```
longhotsummer: Bit brute-force, but it works I guess...

PR #984, opened 2020-03-13
Annotation generator

### LINKED ISSUE #913: As an amending editor, I want support generating amendment annotations
Add ‘Insert amendment annotation’ to menu, which inserts `[[section / subsection / paragraph <number> substituted / inserted / amended / deleted by section <number> of <amending work>]]` or similar

`<amending work>` can be determined by the current point in time’s amending work, it should default to the numbered title (eg. Act 5 of 2006).

goose-life:
adds support for adding annotation generators and partially closes #913 

@longhotsummer, is it okay that annotationGeneratorBase is added as the default here in `traditions`? I tried to do it in https://github.com/laws-africa/indigo-la-tradition/blob/master/indigo_la_tradition/static/javascript/indigo/indigo_la_tradition/traditions.js but it seemed to get overridden?


longhotsummer:
Urg, in Indigo 'annotation' means a comment made by a user on a document. So we should use a different word internally. 'Remark' is the AKN term, so perhaps 'auto remarks'?

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -69,6 +70,10 @@
 
       this.setupRenderers();
       this.textEditor.getSession().setMode(this.parent.model.tradition().settings.grammar.aceMode);
+
+      // get the appropriate annotation style for the tradition
+      this.annotationGenerator = Indigo.annotations[this.parent.model.tradition().settings.annotationStyle];
+      this.annotationAmendingWork = this.getAnnotationAmendingWork(this.parent.model);
```
longhotsummer: I think we could safely call this `amendingWork`, since it's not annotation specific.

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -667,6 +672,40 @@
       this.textEditor.focus();
     },
 
+    getAnnotationAmendingWork: function(document) {
+      var date = document.get('expression_date'),
+          documentAmendments = document.get('amendments'),
+          amendments = [];
+
+      documentAmendments.forEach(function(amendment) {
```
longhotsummer: Use underscore's `findWhere`: https://underscorejs.org/#findWhere

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -667,6 +672,40 @@
...
+      if (amendments) {
+        return amendments[0];
+      }
+
+    },
+
+    insertAnnotation: function(e) {
+      e.preventDefault();
+      var element = 'section/subsection/paragraph/subparagraph/<other> ()',
+          verb = 'substituted/amended/inserted/deleted/repealed/<other>',
```
longhotsummer: What if the menu had a sub-menu that let the user choose the verb?

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -667,6 +672,40 @@
...
+          amendingSection = 'section <insert>',
+          annotation;
+
+      if (this.annotationGenerator && this.annotationAmendingWork) {
+        annotation = this.annotationGenerator(element, verb, amendingSection, this.annotationAmendingWork);
```
longhotsummer: It feels to me that the generator should generate all of the text, given the various context elements. eg, provide it with:

* current document object
* amending work object
* verb (if chosen from menu, otherwise null)
* current element being edited (if any)

Then the generator can make use of those things to do something smart, including taking into account the document language. So all hardcoded strings are removed from this code.

indigo_app/static/javascript/indigo/views/document_editor.js
goose-life: Nice! I initially thought a pop-up form where you choose a verb and give it the other details as a second-generation version, but this could work nicely too. So we'd store a list of verbs per language somewhere and show the relevant language's list as sub options?

indigo_app/static/javascript/indigo/views/document_editor.js
goose-life: That would be very cool. I wanted to ask you about a (hopefully simple?) way of figuring out where we currently are in the document based on the information to hand.

In case it's relevant, just mentioning it won't be the amending work object but rather a serialization of it, i.e. its title and FRBR URI. Unless we want more, in which case I assume we'd have to change the serializer?

indigo_app/static/javascript/indigo/views/document_editor.js
goose-life: I wanted to be careful because it's a guess at the amending work that brought about the current document (if there's more than one you're just getting the first) and can't be relied on by our future selves for anything other than annotations. but I guess I can make a comment to that effect?

longhotsummer: If it's ambiguous, then I suggest looking it up at the last moment when the menu item is selected. There's no advantage to storing it globally then. Additionally, it can change if the user changes the expression date for the current document (unlikely, but worth bearing in mind).

indigo_app/static/javascript/indigo/views/document_editor.js
longhotsummer: I don't think it needs to be translated. Think of the menu as the user showing their intent: "I want to generate a remark about an insertion...". We then generate an appropriate, localised, remark.

indigo_app/static/javascript/indigo/views/document_editor.js
longhotsummer: > That would be very cool. I wanted to ask you about a (hopefully simple?) way of figuring out where we currently are in the document based on the information to hand.

`this.fragment`

> In case it's relevant, just mentioning it won't be the amending work object but rather a serialization of it, i.e. its title and FRBR URI. Unless we want more, in which case I assume we'd have to change the serializer?

That's fine for now.

goose-life: `this.fragment` only gives down to the section number, so we can't use it to generate e.g. `paragraph (b) substituted by …`

longhotsummer: Yeah, it's the best we have right now (could also be the whole document!) Useful to pass to the generator, though. It could make the assumption that the currently edited node is what changed, and the user could correct.

AFAIK, legislation.gov.uk only make amendment remarks down to the section level.

goose-life: It would be cool to use the amending work's `numbered_title` in the annotation (for e.g. `Act 3 of 2019`, `GN 45 of 2002`, and for localisation), but we'd have to inject that into a document's `amendment_events`, for which we'd have to change cobalt's `AmendmentEvent`. I think it might be worth doing, especially once we start working in multiple languages and/or with more provincial and national legislation.

longhotsummer: Simplest is to serialize the complete amending work information into the page as it loads, like we do with other expressions, and then look it up based on frbr uri.

longhotsummer: Oh, and then include `numbered_title` in the serialisation, which is actually useful even externally.

indigo_app/static/javascript/indigo/traditions.js
```diff
@@ -133,8 +133,8 @@
         },
       },
     },
-    // default annotation style
-    annotationStyle: 'annotationGeneratorBase',
+    // default remark style
+    remarkStyle: 'remarkGeneratorBase',
```
longhotsummer: Perhaps `remarkGenerator` is better than `remarkStyle`? The generate determines the style, but this is a reference to the generator.

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -672,37 +671,28 @@
...
     },
 
-    insertAnnotation: function(e) {
+    insertRemark: function(e) {
       e.preventDefault();
-      var element = 'section/subsection/paragraph/subparagraph/<other> ()',
-          verb = 'substituted/amended/inserted/deleted/repealed/<other>',
-          amendingSection = 'section <insert>',
-          annotation;
+      var verb = e.currentTarget.className.split(" ").pop(),
```
longhotsummer: This is very fragile. Much better to put the data into a data attribute (eg. `<span ... data-verb="foo">`) and use `e.currentTarget.getAttribute('data-verb')`.

goose-life:
Localisation, using numbered_title

image

image

image

indigo_app/views/documents.py
```diff
@@ -49,9 +49,15 @@ def get_context_data(self, **kwargs):
 
         context['document_content_json'] = json.dumps(doc.document_xml)
 
-        context['amendments_json'] = json.dumps(
-            WorkAmendmentSerializer(context={'request': self.request}, many=True)
-            .to_representation(doc.work.amendments))
+        # add 'numbered_title_localised' to each amendment
```
longhotsummer: My thinking was to add this to `indigo_api.serializers.WorkSerializer`, then it's included everywhere. You can just add `numbered_title` to the `fields` list on that serializer.

goose-life: I've done that as well, this is to localise the numbered title for the annotations only.

longhotsummer: Ah, I see.

longhotsummer: I wonder if we're trying to be too smart, but let's see how it goes.

PR #986, opened 2020-03-17
Subpart

longhotsummer:
Requires updated slaw (`bundle update`)

Advocates (Remuneration) Order @ 2017-11-17 2020-03-17 11-06-03

Advocates (Remuneration) Order @ 2017-11-17 2020-03-17 11-06-15
Advocates (Remuneration) Order @ 2017-11-17 2020-03-17 11-06-27


goose-life:
Looks good! My only concern / question is around subparts without numbers. Will these be supported differently / at a later stage?

longhotsummer:
We will have to add it later.

Issue #991, opened 2020-03-19
As a table editor, I want to be able to easily mark text as bold and italics


longhotsummer:
Can we just use the CKEditor support for this?

PR #992, opened 2020-03-19
Toggle bold and italics in table editor view

longhotsummer:
We add our own buttons because the CKEditor toolbar is placed
over our own toolbar, and we can't move it.


longhotsummer:
Fixes #991

PR #1009, opened 2020-03-30
support adding taxonomy when creating works in bulk

### LINKED ISSUE #1008: As a bulk importer I want to link the taxonomies

goose-life:
closes #1008

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -15,7 +15,7 @@ from googleapiclient.errors import HttpError
 from google.oauth2 import service_account
 
 from indigo.plugins import LocaleBasedMatcher, plugins
-from indigo_api.models import Subtype, Work, PublicationDocument, Task, Amendment, Commencement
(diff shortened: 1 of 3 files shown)


goose-life:
image

image

image

image

indigo/bulk_creator.py
```diff
@@ -466,6 +470,21 @@ def link_amendment(self, work, info):
             amendment.date = date
             amendment.save()
 
+    def link_taxonomy(self, work, info):
+        topics = info.get('taxonomy').split(', ')
```
longhotsummer: This will break when a user enters "foo,bar, boom,   baz" accidentally. I suggest splitting on either a comma and then stripping whitespace, or splitting on whitespace and then stripping commas.

longhotsummer:
```suggestion
        topics = [x.strip(",") for x in info.get('taxonomy').split()]
```

PR #1014, opened 2020-04-01
link repeals on bulk import

### LINKED ISSUE #1012: as a spreadsheet importer I want repeals to link automatically
Currently, repeals only link automatically on first import. After that, not only are they not linked automatically, there's no task for manually linking it.

Suggestions: 
1. Change the relationship in the spreadsheet from 'repealed_by' on the repealed work to 'repeals' on the repealing work (same as 'amends').

2. 'Re-import' the repealed work from the spreadsheet along with its repealing work (or after the repealing work has been created). The repealed work won't be re-imported, but we can link the repeal, as we currently do with 'amends' (https://github.com/laws-africa/indigo/blob/master/indigo/bulk_creator.py#L195-L198).

goose-life:
closes #1012

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -179,10 +179,12 @@ class BaseBulkCreator(LocaleBasedMatcher):
 
         if not dry_run:
             for info in works:
-                if info['status'] == 'success':
-                    if info.get('commencement_date') or info.get('commenced_by'):
-                        self.link_commencement(info['work'], info)
+                # link all commencements first so that amendments and repeals will have dates to work with
+                if info['status'] == 'success' and info.get('commencement_date') or info.get('commenced_by'):
+                    self.link_commencement(info['work'], info)
 
+            for info in works:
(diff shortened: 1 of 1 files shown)


indigo/bulk_creator.py
```diff
@@ -409,25 +409,32 @@ def link_commencement(self, work, info):
...
+
         if not repealing_work:
-            return self.create_task(work, info, task_type='link-repeal')
+            # an exact match on the given title wasn't found
+            self.create_task(work, info, task_type='no-repeal-title-match')
 
-        repeal_date = repealing_work.commencement_date
-        if not repeal_date:
-            return self.create_task(work, info, task_type='link-repeal')
+        elif work.repealed_by and work.repealed_by.id != repealing_work.id:
```
longhotsummer, line 419: you should be able to use `elif work.repealed_by and work.repealed_by != repealing_work`

goose-life: cool, thanks!

Issue #982, opened 2020-03-13
As an editor creating a new PIT from an old one, I want the attachments to be copied over to the new one.


longhotsummer:
eg. see https://edit.laws.africa/places/za-wc015/tasks/5634/

PR #1022, opened 2020-04-06
Copy Attachments to New PIT automatically.

ro6ley:
#### What does this PR do?
Copy attachments to new PIT created from an existing one.

As an editor creating a new PIT from an old one, I want the attachments to be copied over to the new one.

Create a new PIT and check the attachments section.

Issue #982

### DIFF
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -4,6 +4,7 @@ from datetime import datetime
 
 from actstream import action
 from django.contrib.postgres.fields import JSONField
+from django.core.files.base import File
 from django.db import models
 from django.db.models import signals, Q
 from django.core.exceptions import ValidationError
@@ -412,6 +413,7 @@ class Work(WorkMixin, models.Model):
         for attachment in attachments:
             attachment.pk = None
             attachment.document = doc
+            attachment.file.save(attachment.filename, File(attachment.file))
             attachment.save()
 
         return doc
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -4,7 +4,6 @@ from datetime import datetime
 
 from actstream import action
 from django.contrib.postgres.fields import JSONField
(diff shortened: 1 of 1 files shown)


indigo_api/models/works.py
```diff
@@ -407,6 +407,13 @@ def create_expression_at(self, user, date, language=None):
         doc.created_by_user = user
         doc.save()
 
+        # copy over attachments to new expression
+        attachments = Attachment.objects.filter(document=template)
```
longhotsummer, line 411: Are you sure this doesn't *move* the existing attachments?

ro6ley: No, according to the docs, when you set the `pk` to `None` a new one will be created maintaining all previous details of the model.

Ref: https://docs.djangoproject.com/en/1.11/topics/db/queries/#copying-model-instances

longhotsummer: Ah, clever.

Hrm, I just tested it, and it doesn't quite work as expected because the file itself (`attachment.file`) doesn't get copied. In my local database, the attachment filename should change:

indigo_api_attachment – indigo – localhost (1) 2020-04-06 18-41-13

158 is the old document id, 975 is the new one. So I think you'll need to actually create new attachment objects and/or copy the one file to the other. I always have to lookup how to do that with Django's FileField fields, which give you FieldFile objects.

ro6ley: Let me recreate the attachment objects entirely instead

ro6ley: Managed to copy the files: 
image

Just a single extra line, will push tomorrow

ro6ley: @longhotsummer  Accessing a FileField returns a FieldFile object that wraps around around the file, and this can be saved a new file on the new document

longhotsummer:
Very simple, nice :)

PR #1033, opened 2020-04-15
Profile photos

### LINKED ISSUE #456: As a contributor I want to see photos of other users

ro6ley:
#### Description
This PR adds profile pictures to Indigo. User will be able to upload, change and remove profile pictures that will appear in select areas of the platform.

Deliverables:
- [x] Profile picture visible on profile page
- [x] Remove/Edit profile picture functionality on profile edit page
- [x] Import Google profile pictures upon sign-up using Google
- [x] Migration to import photos of existing users with connected Google accounts
- [x] Thumbnails on the leaderboard on the Overview Page
- [x] Thumbnails on the user popup used across Indigo
- [x] Confirmation popup when removing current profile picture
- [x] Post processing uploaded images i.e resizing, cropping
- [x] Caching of profile pictures

#### Manual Testing
- A default avatar is displayed for users without profile pictures and should be visible on the user profile page, user profile edit page, user popup. Navigating to any of these pages should display the default user avatar or their custom avatar if one is available.
- You should be able to upload a new profile picture or remove an existing one in the edit profile section.
- Signing up a new user using Google should import their profile photo, save it, and display it as their profile photo across Indigo.

#### Relevant Issue
closes #456 

#### Context
- Wording on the disclaimer on edit/upload section subject to review
- Default profile picture is a fontawesome free icon in SVG format

#### Screenshots

User Profile page with a profile photo:
Screenshot_2020-04-14 Robley Adrian – Indigo Platform

User profile using the default profile picture:
image

Profile edit section for a user without a profile photo:
image

Profile edit section for a user changing an existing profile photo:
image

Leaderboard with thumbnails:
image

User Popup with thumbnail:
Screenshot from 2020-04-15 16-57-46


indigo_social/models.py
```diff
@@ -25,3 +54,17 @@ def create_user_profile(sender, **kwargs):
     if not hasattr(user, "userprofile") and not kwargs.get("raw"):
         user_profile = UserProfile(user=user)
         user_profile.save()
+
+
+@receiver(user_signed_up, sender=User)
+def retrieve_profile_photo_on_signup(sender, **kwargs):
+    user = kwargs['user']
+    user_profile = UserProfile.objects.get(user=user)
```
longhotsummer, line 62: This will throw `UserProfile.DoesNotExist` if the object doesn't exist. Either handle the exception, or use `.filter(...).first()`.

indigo_social/templates/indigo_app/user_account/edit.html
```diff
@@ -88,17 +88,36 @@
...
-              </div>
-            {% endif %}
-            -->
+            <p>For the best experience, kindly ensure your profile picture is square and smaller than 2MB in size. Uploading a new image will replace the existing one.</p>
+
+            <div class="mt-2">
+              <img src="{{ user.userprofile.profile_photo_url }}" class="img-thumbnail" width="300" height="300">
+              {% if user.userprofile.profile_photo %}
+                <div class="btn-group-toggle mt-2" data-toggle="buttons">
+                  <label class="btn btn-danger" for="profile_photo-clear_id">
```
longhotsummer, line 98: Rather make this a button, that's more natural for an action than a checkbox.

longhotsummer:
When the user uploads an image, we should post-process it to resize it to our preferred size and then crop it to be square. Django exposes `.image` which is a [Pillow](https://pillow.readthedocs.io/en/stable/) image object.

indigo_social/models.py
```diff
@@ -1,9 +1,32 @@
...
 
+from urllib.request import urlretrieve, urlcleanup
+
+from allauth.account.signals import user_signed_up
+from allauth.socialaccount.models import SocialLogin, SocialAccount
+from allauth.socialaccount.signals import pre_social_login, social_account_added, social_account_updated
+
+
+def user_profile_photo_path(instance, filename):
+    return 'avatars/{}/{}'.format(instance.id, os.path.basename(filename))
```
longhotsummer, line 18: I don't even think it needs a filename. We don't want to keep old photos around -- just save to `avatars/99` for UserProfile with pk=99.

indigo_social/models.py
```diff
@@ -13,10 +36,16 @@ class UserProfile(models.Model):
...
-    profile_photo = models.ImageField(blank=True, null=True)
+    profile_photo = models.ImageField(upload_to=user_profile_photo_path, blank=True, null=True)
     twitter_username = models.CharField(max_length=256, blank=True, null=True, default='')
     linkedin_profile = models.URLField(blank=True, null=True, default='')
 
+    @property
+    def profile_photo_url(self):
+        if not self.profile_photo:
+            return "/static/images/avatars/default_avatar.svg"
+        return self.profile_photo
```
longhotsummer, line 47: This is where we'd return something like `reverse('user_profile_photo', kwargs={'username': self.user.username, 'nonce': self.profile_photo_nonce}` which would be a url such as `/contributors/greg/avatar/abc123`

indigo_social/models.py
```diff
@@ -1,8 +1,61 @@
...
+from urllib.request import urlretrieve, urlcleanup
+
+from allauth.account.signals import user_signed_up
+from allauth.socialaccount.models import SocialLogin, SocialAccount
+from allauth.socialaccount.signals import pre_social_login, social_account_added, social_account_updated
+
+
+def user_profile_photo_path(instance, filename):
+    file_name, file_extension = os.path.splitext(filename)
+    return 'avatars/{}'.format(instance.id)
```
longhotsummer, line 24: I've been burned by `instance.id` being None in the past. I suggest check for None and raise `ValueError`.

indigo_social/models.py
```diff
@@ -1,8 +1,61 @@
...
+    """ Retrieve and save a retrieved user's profile photo from a url
+    """
+    try:
+        filename, _ = urlretrieve(url)
+        user_profile.profile_photo.save(filename, File(open(filename, 'rb')))
+    finally:
+        urlcleanup()
+
+
+def resize_photo(profile_photo, user):
```
longhotsummer, line 37: I suggest putting this into the UserProfile class, then it has all the context it needs.

longhotsummer: It can then resize itself, effectively.

indigo_social/models.py
```diff
@@ -1,8 +1,61 @@
...
+    """ Resize and crop an uploaded profile photo to meet specifications
+    """
+    uploaded_profile_photo = Image.open(profile_photo)
```
longhotsummer, line 40: FYI, the Django docs say you can use `profile_photo.image`. Haven't tried it :)

indigo_social/models.py
```diff
@@ -13,10 +66,23 @@ class UserProfile(models.Model):
...
+    def save(self):
+        # TODO: Generate and save the nonce here
+        self.profile_photo = resize_photo(self.profile_photo, self.user)
+
+        super(UserProfile, self).save()
+
+    @property
+    def profile_photo_url(self):
+        if not self.profile_photo:
+            return "/static/images/avatars/default_avatar.svg"
```
longhotsummer, line 83: https://stackoverflow.com/a/17738606

longhotsummer: Using `static` means we get fingerprinting for the assets for cache purposes.

indigo_social/views.py
```diff
@@ -292,3 +293,17 @@ class UserPopupView(AbstractAuthedIndigoView, DetailView):
     slug_url_kwarg = 'username'
     template_name = 'indigo_social/user_popup.html'
     queryset = User.objects
+
+
+class UserProfilePhotoView(View):
```
longhotsummer, line 298:
```suggestion
class UserProfilePhotoView(AbstractAuthedIndigoView):
```

That view will ensure a user is logged in if the site requires it.

indigo_social/views.py
```diff
@@ -292,3 +293,17 @@ class UserPopupView(AbstractAuthedIndigoView, DetailView):
...
+    def get(self, request, **kwargs):
+        username = kwargs['username']
+        nonce = kwargs['nonce']
+
+        user = User.objects.filter(username=username).first()
```
longhotsummer, line 303: Use `get_object_or_404` otherwise you'll get a 500 internal error if the user doesn't exist. Once you have a user, you can use `user.user_profile`.

indigo_social/views.py
```diff
@@ -292,3 +293,17 @@ class UserPopupView(AbstractAuthedIndigoView, DetailView):
...
+        user_profile = UserProfile.objects.filter(user=user).first()
+
+        response = HttpResponse(user_profile.profile_photo.read())
```
longhotsummer, line 306: The photo might not exist, redirect to the static image in that case.

indigo_social/migrations/0005_import_google_profile_pictures.py
```diff
@@ -0,0 +1,40 @@
...
+
+from urllib.request import urlretrieve, urlcleanup
+
+
+def import_google_profile_pictures(apps, schema_editor):
+	SocialAccount = apps.get_model('allauth.socialaccount', 'SocialAccount')
+	UserProfile = apps.get_model('indigo_social', 'UserProfile')
+
+	def fetch_and_save_profile_photo(user_profile, url):
+		try:
```
longhotsummer, line 16: Pls change tabs to spaces.

The for loop on line 22 is over-indented and so won't run.

indigo_social/models.py
ro6ley: The image is usually stored in `profile_photo.photo` but it still raises an error because it has not been saved yet

longhotsummer: I suspect this method could be simpler: it could take an `image` object modify it in place. The caller would then be responsible for writing it back to the file object.

longhotsummer: Alternatively return a new image object. Basically, you can avoid writing files and do it in memory.

indigo_social/models.py
```diff
@@ -25,3 +91,17 @@ def create_user_profile(sender, **kwargs):
...
         user_profile = UserProfile(user=user)
         user_profile.save()
+
+
+@receiver(user_signed_up, sender=User)
+def retrieve_profile_photo_on_signup(sender, **kwargs):
+    user = kwargs['user']
+
+    try:
+        user_profile = UserProfile.objects.get(user=user)
```
longhotsummer, line 101: Use `user.user_profile` instead.

indigo_social/views.py
ro6ley: The physical photo? We already redirect to the static photo when getting the photo_url in the model

longhotsummer: There's no guarantee this object will exist in the db by the time the request comes here. A user could load a page from their cache which is old, etc.

indigo_social/models.py
ro6ley: I have actually moved it to the form, and have it take an image object only instead without all the other context, then returns a processed image

indigo_social/models.py
```diff
@@ -71,17 +49,15 @@ class UserProfile(models.Model):
...
     linkedin_profile = models.URLField(blank=True, null=True, default='')
 
-    def save(self):
-        # TODO: Generate and save the nonce here
-        self.profile_photo = resize_photo(self.profile_photo, self.user)
-
-        super(UserProfile, self).save()
+    def generate_nonce(self):
+        self.profile_photo_nonce = ''.join(random.sample(string.ascii_lowercase, 8))
+        self.save()
```
longhotsummer, line 54: A good rule of thumb is not to call `save()` from inside a model. In some cases you may not want to save it, or it's being saved prematurely, etc. Instead, the part of the code closest to the user's request (eg. the view) should call `save()` once it knows everything is ready.

indigo_social/views.py
```diff
@@ -292,3 +294,19 @@ class UserPopupView(AbstractAuthedIndigoView, DetailView):
...
+    def get(self, request, **kwargs):
+        username = kwargs['username']
+        nonce = kwargs['nonce']
+
+        user = get_object_or_404(User, username=username)
+
+        if not user.userprofile.profile_photo:
+            return redirect(static('images/avatars/default_avatar.svg'))
+
+        response = HttpResponse(user.userprofile.profile_photo.read())
```
longhotsummer, line 309: You could probably get away with a one-liner:

```
return FileResponse(user.userprofile.profile_photo)
```

indigo_social/views.py
```diff
@@ -292,3 +294,19 @@ class UserPopupView(AbstractAuthedIndigoView, DetailView):
...
+        response['Content-Disposition'] = 'inline; filename={}'.format(user.userprofile.profile_photo.name)
```
longhotsummer, line 310: This header isn't necessary, since we don't really care about the user downloading the file to the filesystem.

indigo_social/forms.py
```diff
@@ -63,6 +75,39 @@ def badge_choices():
...
+    if width != height:
+        new_width = min(width, height)
+
+        left = (width - new_width)/2
+        top = (height - new_width)/2
+        right = (width + new_width)/2
+        bottom = (height + new_width)/2
+        uploaded_photo = uploaded_photo.crop((left, top, right, bottom))
+        uploaded_photo = uploaded_photo.resize((200, 200), Image.ANTIALIAS)
+    else:
```
longhotsummer, line 98: line 97 is duplicated, you can drop the else and make line 99 always happen, and remove line 97

ro6ley: fixed

indigo_app/templatetags/indigo_app.py
```diff
@@ -30,6 +30,14 @@ def user_profile(user):
     return html
 
 
+@register.simple_tag
+def user_profile_photo_thumbnail(user, height=32, width=32):
+    if not user:
+        return ''
+
+    return format_html('<img src="{}" height="{}" width="{}" style="border-radius: 50%;">'.format(user.userprofile.profile_photo_url, height, width))
```
longhotsummer, line 38:
```suggestion
    return format_html('<img src="{}" height="{}" width="{}" class="user-profile-photo">'.format(user.userprofile.profile_photo_url, height, width))
```

Then add a class in `_core.scss` that gives it a 5px radius, so we're consistent everywhere.

PR #1036, opened 2020-04-16
Handle deletion of profile photo; saving user profile form

longhotsummer:
* Only submit the photo file if it is changing. Otherwise a user could change the input field to use another user's photo.
* Accept only image files.
* Delete submits the form

### DIFF
--- a/indigo_social/forms.py
+++ b/indigo_social/forms.py
@@ -9,7 +9,6 @@ from pinax.badges.models import BadgeAward
 
 
 class UserProfileForm(forms.ModelForm):
-
     validate_username = RegexValidator(
         _lazy_re_compile(r'^[-a-z0-9_]+\Z'),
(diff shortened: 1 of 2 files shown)


indigo_social/forms.py
```diff
@@ -50,7 +49,14 @@ def clean_username(self):
         return username
 
     def save(self, commit=True):
+        if self.cleaned_data.get('new_profile_photo'):
+            self.instance.profile_photo = self.cleaned_data['new_profile_photo']
```
longhotsummer, line 53: This is the perfect point to resize the photo, rather than doing it on every save. At this point we know it's new, we know it must be resized, and we know it exists.

longhotsummer: We can also resize it based on the current in-memory image file, rather than writing to s3, reading back, resizing, writing back to s3.

longhotsummer:
Merge at your convenience.

PR #1037, opened 2020-04-17
Cobalt imports

goose-life:
goes with laws-africa/cobalt#24


setup.py
```diff
@@ -42,7 +42,7 @@
     install_requires=[
         'django>=1.11.27,<2',
         'boto3>=1.7',
-        'cobalt>=3.1',
+        'cobalt @ git+https://github.com/laws-africa/cobalt@0e42ef02239c18e21715e163d0b61483270a0a02#egg=cobalt',
```
longhotsummer, line 45: It may be useful to refer just to the akn3 branch, otherwise we'll need to change this the whole time as we iterate on Cobalt.

Issue #31, opened 2017-03-18
Deprecate and remove the LIME editor


longhotsummer:
The editor is not used in AfricanLIIs indigo, is complicated to integrate, and out of date.

PR #1038, opened 2020-04-20
Redirect to GazetteMachine Post-login

ro6ley:
#### Description
If a user tries to go to https://gazettes.laws.africa/next-task to do a task and they aren't logged in, they're sent to edit.laws.africa to sign in (by design). However, they are sent to https://edit.laws.africa/accounts/login/?next=/next-task which doesn't have the correct domain.

This PR adds a middleware to facilitate the redirection back to gazettes domain post-login.

#### Manual Testing
Log out of Indigo and access Gazettes. After logging in again, the page will redirect to Gazettes and not raise an error.

#### Relevant Issue
[Gazettemachine #31](https://github.com/laws-africa/gazettemachine/issues/31)

### DIFF
--- a/indigo/middleware.py
+++ b/indigo/middleware.py
@@ -1,0 +1,11 @@
+from django.http import HttpResponsePermanentRedirect
+
+
+class GazettesRedirectMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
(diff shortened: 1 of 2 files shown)


longhotsummer:
Hacky, but it'll do. However, it needs to be on indigo-lawsafrica please :)

PR #1041, opened 2020-04-20
extend work view when viewing a task on a work

### LINKED ISSUE #1024: As a task user, I want to see the Work tabs in a task linked to a work
eg. if the task is linked to a work, then when showing the task, show it as part of the work layout, with the Task tab selected.

This will change the template that the task template extends. I don't think that can be determined in the template itself with an if statement, but I'm not sure.

If not:

* [ ] create a new `work_task.html` to complement `task.html` but for when the task is linked to a work. `work_task` should extend from the work layout.
* [ ] create a new `task_detail.html` or similar which both the new template and the old template can include.
* [ ] task detail view must decide which top-level template to render.

goose-life:
closes #1024

task on a work
image

after nuking `Create task`
image

task without a work
image


indigo_app/views/tasks.py
```diff
@@ -132,8 +132,17 @@ def get_context_data(self, **kwargs):
...
         Task.decorate_permissions([task], self)
 
+        # add work to context
+        if task.work:
+            context['work'] = task.work
+
         return context
 
+    def get_template_names(self):
+        names = super().get_template_names()
```
longhotsummer, line 142: 1. rather than calling super() and then ignoring it, do the if statement first, then call super() if necessary.
2. for symmetry, please rename `work_task.html` to `work_task_detail.html`.

PR #1044, opened 2020-04-22
Fix commenced provisions

### LINKED ISSUE #1034: Commencement section ordering
See https://openbylaws.org.za/za/act/gn/2020/318/eng/

Is it correct that it says that "1A to 11G"  commenced on X date? That ordering seems strange.

Disaster Management Act: Regulations relating to COVID-19 - Open By-laws South Africa 2020-04-15 14-55-02

goose-life:
closes #1034


goose-life:
Multiple:
image

Table:
image

indigo_api/models/works.py
```diff
@@ -243,10 +257,8 @@ def commenceable_provisions(self):
             plugin = plugins.for_document('toc', doc)
             if plugin:
                 toc = plugin.table_of_contents_for_document(doc)
-                for item in plugin.commenceable_items(toc):
-                    if item.id and item.id not in id_set:
-                        id_set.add(item.id)
-                        provisions.append(item)
+                items = plugin.commenceable_items(toc)
+                provisions = self.insert_provisions(provisions, id_set, items)
```
longhotsummer, line 261: you're going to have to walk me through what's going on here. 

i'm also not sure that `insert_provisions` is a useful helper method on the work object.

longhotsummer: ah, I get it. two things:

1. `insert_provisions` is a very specific helper function. i don't think it has external value. i suggest just pulling that loop into the `commenceable_provisions`
2. some additional comments on what `insert_provisions` is doing would make it a bit clearer. I've added some suggestions.

indigo_api/models/works.py
```diff
@@ -229,6 +229,20 @@ def first_commencement_date(self):
         if first:
             return first.date
 
+    def insert_provisions(self, provisions, id_set, items):
+        # take note of any removed items to compensate for when inserting
+        removed_indexes = [i for i, p in enumerate(provisions) if p.id not in [i.id for i in items]]
+        for i, item in enumerate(items):
+            if item.id and item.id not in id_set:
+                id_set.add(item.id)
+                # compensate for removed items
```
longhotsummer, line 238:
```suggestion
                # We need to insert this provision at the correct position in the work provision list. If any provisions from a previous document have been removed in this document (indexes stored in removed_indexes), bump the insertion index up to take them into account.
```

indigo_api/models/works.py
goose-life: Oh, I remember now. I moved `insert_provisions` out so I could test it. What's the better way of doing that?

longhotsummer: That's a good reason. Can we move lines 259 onwards into the TOC plugin class? I think it makes more sense to have the helper function there, which  makes it easier to test. Something like:

```python
plugin.insert_commenceable_provisions(doc, provisions, id_set)
```

and the plugin then splits out those two functions. It doesn't need to return anything because `provisions` and `id_set` are updated in-place.

goose-life: Great idea, thanks!

setup.py
```diff
@@ -67,6 +67,7 @@
         'django-wkhtmltopdf>=2.0.3,<2.1',
         'djangorestframework-xml>=1.3.0',
         'djangorestframework>=3.11.0',
+        'dotmap>=1.3.8',
```
longhotsummer, line 70: This is only needed for tests, so can safely be moved into the `test` part of `extras_require` below.

PR #1045, opened 2020-04-22
support /akn URI prefix

### LINKED ISSUE #953: Indigo must support /akn FRBR URIs throughout
* [x] Handle new cobalt package layout (not AKN3-specific)
* [x] Migration to change all existing URIs: Works and Documents
* [x] Migration to change existing URIs: Work and Document historical versions
* [x] Migration to change ref/ passiveRef tags in all documents (incl. deleted)
* [x] Migration to change ref/ passiveRef tags: historical versions of all documents
* [x] Existing code to deal with new URIs
* [x] Creating new documents (bulk import, single work page)
* [x] Website URLs must change
* [x] API URLs must change

goose-life:
closes #953


indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,139 @@
...
+        work_component=uri.work_component,
+        language=uri.language,
+        expression_date=uri.expression_date,
+        expression_component=uri.expression_component,
+        expression_subcomponent=uri.expression_subcomponent,
+        format=uri.format
+    )
+
+
+def set_frbr_uri_meta_ident(doc, uri):
```
longhotsummer, line 42: Why not just load the document as a Cobalt doc, and then set the new FRBR URI, and let cobalt do all this work?

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,139 @@
...
+    """ Generates a cobalt FrbrUri object using all the existing URI's attributes.
+        If `forward` is True, the optional /akn prefix is included; otherwise it isn't.
+    """
+    if not isinstance(uri, FrbrUri):
+        uri = FrbrUri.parse(uri)
+    prefix = None
+    if forward:
+        prefix = 'akn'
+
+    return FrbrUri(
```
longhotsummer, line 24: You can also use `frbr_uri = frbr_uri.clone(); frbr_uri.prefix = 'xxx'`

longhotsummer: I don't even think you need to generate a new object, you can simply set the prefix on the existing object.

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: sounds great! how do I set the new FRBR URI?

goose-life: got it!

goose-life: (I assumed it'd be more complicated than just literally setting it)

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,92 @@
...
+    - Update URIs of all existing works and documents
+    - Update meta/identification block of each AKN document
+    - Update hrefs in all documents
+    """
+    db_alias = schema_editor.connection.alias
+    Work = apps.get_model("indigo_api", "Work")
+    Document = apps.get_model("indigo_api", "Document")
+
+    for work in Work.objects.using(db_alias).all():
+        work.frbr_uri = new_frbr_uri(work.frbr_uri, forward)
```
longhotsummer, line 64: I *think* you'll have to call `.work_uri()` on this, because at this point the model doesn't have custom logic on it, only the raw fields, and this is a string field.

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,92 @@
...
+    """ Sets prefix on uri:
+        'akn' if forward is True, None if it's False.
+    """
+    if not isinstance(uri, FrbrUri):
+        uri = FrbrUri.parse(uri)
+    uri.prefix = 'akn' if forward else None
+    return uri
+
+
+class RefsFinderRef(BaseRefsFinder):
```
longhotsummer, line 24: I'm not sure that BaseRefsFinder is doing anything useful for you here. If your xpath is restricted to `//a:ref`, then `ref` tags can only occur in textual elements, so you don't need any of the ancestor stuff. At that point, `handle_refs` doesn't need anything from the parent class and it can be a simple function.

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    db_alias = schema_editor.connection.alias
+    Work = apps.get_model("indigo_api", "Work")
+    Document = apps.get_model("indigo_api", "Document")
+
+    for work in Work.objects.using(db_alias).all():
+        work.frbr_uri = new_frbr_uri(work.frbr_uri, forward)
+        work.save()
+
+    for doc in Document.objects.using(db_alias).all():
```
longhotsummer, line 67: We also need to update all the saved historical versions of Work and Document objects 😱 . I suggest doing that as a separate migration after this one.

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: If I restrict the xpath to `//a:ref`, the hrefs in e.g. `passiveRef` in `meta` would need to be updated separately, right?

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: I don't follow? I agree this is a string, but `new_frbr_uri()` should handle that?
(Also am I wrong in thinking that `.work_uri()` is unavailable because it's a method on Work?)

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: omg 🤦

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: But if I were to keep the existing `candidate_xpath`, are you saying I could just do something like,
for node in root where there's an href starting with '/', do the right thing?

indigo_api/migrations/0129_migrate_refs_uris.py
longhotsummer: `new_frbr_uri()` returns an FrbrUri object, whereas `work.frbr_uri` expects a string.

indigo_api/migrations/0129_migrate_refs_uris.py
longhotsummer: > If I restrict the xpath to `//a:ref`, the hrefs in e.g. `passiveRef` in `meta` would need to be updated separately, right?

I had forgotten about those. As long as you're sure that all `href` attributes are safe to process as AKN URIs, then we could do them all in one go. 

> But if I were to keep the existing candidate_xpath, are you saying I could just do something like, for node in root where there's an href starting with '/', do the right thing?

Yes. Although I don't think you need the candidate xpath at all. Just process every `ref` (or element with an `href`) in the entire document.

indigo_app/static/javascript/indigo/models.js
```diff
@@ -422,19 +422,19 @@
...
-    // /za-cpt/act/by-law/2011/foo
+    // /akn/za-cpt/act/by-law/2011/foo
     var parts = frbr_uri.split('/'),
-        tmp = parts[1].split('-'),
-        country = tmp[0],
-        locality = tmp.length > 1 ? tmp[1] : null,
-        bump = parts.length > 5 ? 1 : 0;
+        place = parts[2].split('-'),
+        country = place[0],
+        locality = place.length > 1 ? place[1] : null,
```
longhotsummer, line 429: If the place code is `za-foo-bar` this breaks. split() in JS isn't nearly as nice as in python, you can't tell it to split once. Easiest is probably to take the substring after the first `-`.

indigo_app/views/works.py
```diff
@@ -50,7 +51,14 @@ class WorkViewBase(PlaceViewBase, AbstractAuthedIndigoView, SingleObjectMixin):
 
     def determine_place(self):
         if 'place' not in self.kwargs:
-            self.kwargs['place'] = self.kwargs['frbr_uri'].split('/', 2)[1]
+            try:
+                uri = FrbrUri.parse(self.kwargs['frbr_uri'])
+                self.kwargs['place'] = uri.place
+            except ValueError:
+                # not a valid URI; make some guesses
+                parts = self.kwargs['frbr_uri'].split('/')
```
longhotsummer, line 59: Because all our FRBR URIs now must start with /akn, can we not always just do the exception part (without parsing the whole FRBR URI)?

indigo_content_api/v1/views.py
```diff
@@ -78,17 +78,31 @@ def parse_frbr_uri(self, frbr_uri):
...
+            place = parts[2] if parts[1] == 'akn' else parts[1]
+            if '-' in place:
+                country, locality = place.split('-')
+            else:
+                country = place
 
         # country
         try:
-            self.country = Country.for_code(parts[0])
+            self.country = Country.for_code(country)
```
longhotsummer, line 99: There's a helper method on Country that will take a place code and give you a country and locality all at once now.

indigo_api/migrations/0129_migrate_refs_uris.py
goose-life: so just checking, I can be namespace agnostic here? 
```
    root = etree.fromstring(document.document_xml)
    # path = etree.XPath("//a:*[starts-with(@href, '/')]", namespaces={'a': cobalt_doc.namespace}) <-- don't need this?
    path = etree.XPath("//*[starts-with(@href, '/')]")
    for node in path(root):
```

indigo_app/views/works.py
goose-life: We can for sure always just do the exception part and still be backwards compatible even, but I thought parsing the URI would be safer / take the guesswork out of it (the exception part was just added to get the tests passing). Are you saying it would be better not to parse it? Is it expensive?

indigo_api/migrations/0129_migrate_refs_uris.py
longhotsummer: Sorry, no, I meant all AKN elements with a an `href` attribute. I can't think of a time when we'd want to ignore namespaces.

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,85 @@
...
+        href="/za/act/2012/22">Act no 22 of 2012
+        href="/akn/za/act/2012/22">Act no 22 of 2012
+
+        For each of these,
+        if `forward` is True, it adds the /akn prefix
+        if `forward` is False, it removes it
+
+    """
+    root = etree.fromstring(document.document_xml)
+    path = etree.XPath("//*[starts-with(@href, '/')]")
```
longhotsummer, line 35: This definitely needs to be namespace aware. Otherwise we're potentially modifying XML that we have no understanding of.

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,85 @@
...
+    for node in path(root):
+        ref = node.get('href')
+        match = FRBR_URI_RE.match(ref)
```
longhotsummer, line 38: (minor code style): you're not using the `match` object, so line 38 can be moved into the `if` statement.

indigo_api/migrations/0129_migrate_refs_uris.py
```diff
@@ -0,0 +1,85 @@
...
+    for work in Work.objects.using(db_alias).all():
+        work.frbr_uri = new_frbr_uri(work.frbr_uri, forward)
+        work.save()
+
+    for doc in Document.objects.using(db_alias).all():
+        doc.frbr_uri = doc.work.frbr_uri
+        cobalt_doc = Act(doc.document_xml)
+        cobalt_doc.frbr_uri = doc.frbr_uri
+        doc.document_xml = cobalt_doc.to_xml()
+        handle_refs(doc, forward)
```
longhotsummer, line 65: to avoid parsing and serialising the XML twice, you could also pass `cobalt_doc` into `handle_refs` so that the latter doesn't have to re-parse and re-serialise it.

indigo_app/static/javascript/indigo/models.js
```diff
@@ -422,19 +422,19 @@
...
-    // /za-cpt/act/by-law/2011/foo
+    // /akn/za-cpt/act/by-law/2011/foo
     var parts = frbr_uri.split('/'),
-        tmp = parts[1].split('-'),
-        country = tmp[0],
-        locality = tmp.length > 1 ? tmp[1] : null,
-        bump = parts.length > 5 ? 1 : 0;
+        place = parts[2].split('-'),
+        country = place[0],
+        locality = place.length > 1 ? place.slice(1,).join('-') : null,
```
longhotsummer, line 429:
```suggestion
        locality = place.length > 1 ? place.slice(1).join('-') : null,
```

nice.

indigo_content_api/v1/views.py
```diff
@@ -78,19 +78,9 @@ def parse_frbr_uri(self, frbr_uri):
...
-            raise Http404
-
-        # locality
-        if len(parts) > 1:
-            self.locality = self.country.localities.filter(code=parts[1]).first()
-            if not self.locality:
-                raise Http404
+        parts = self.kwargs['frbr_uri'].split('/')
+        place = parts[2] if parts[1] == 'akn' else parts[1]
+        self.country, self.locality = Country.get_country_locality(place)
```
longhotsummer, line 83: `get_country_locality` raises a `DoesNotExist` exception for Country or Locality, which needs to be turned into an Http404

indigo_api/models/works.py
```diff
@@ -487,7 +491,8 @@ class PublicationDocument(models.Model):
     updated_at = models.DateTimeField(auto_now=True)
 
     def build_filename(self):
-        return '{}-publication-document.pdf'.format(self.work.frbr_uri[1:].replace('/', '-'))
+        # don't include /akn/ from FRBR URI in filename
+        return f"{{{self.work.frbr_uri[5:].replace('/', '-')}}}-publication-document.pdf"
```
longhotsummer, line 495: Is this correct to have so many curly brackets?

longhotsummer:
I think if you push your updated AKN3 branch, with master merged in, GitHub will realise this is actually a smaller change.

longhotsummer:
Nice progress, that migration is looking good now.

goose-life:
The test that's still failing is for indigo_metrics; it's not super obvious to me what's being tested so I'm not sure if the test should be updated?

longhotsummer:
> The test that's still failing is for indigo_metrics; it's not super obvious to me what's being tested so I'm not sure if the test should be updated?

My guess is we have removed a new work or place that is being included in the daily metrics test. Check what the `dailies` variable contains. Is it expected?

indigo_content_api/v1/urls.py
```diff
@@ -16,18 +17,18 @@
 
     # Document/work media
     # Work publication document
-    re_path(r'^(?P<frbr_uri>[a-z]{2}[-/]\S+?)/media/publication/(?P<filename>.*)$', views.PublishedDocumentMediaView.as_view({'get': 'get_publication_document'}), name='published-document-publication'),
+    re_path(r'^(?P<frbr_uri>(akn/)?[a-z]{2}[-/]\S+?)/media/publication/(?P<filename>.*)$', views.PublishedDocumentMediaViewV1.as_view({'get': 'get_publication_document'}), name='published-document-publication'),
```
longhotsummer, line 20: I *think* that we can remove the akn completely, rather than making it optional. The V1 api shouldn't have `/akn` at the start of the FRBR URIs at all -- if one is provided, it should return a 404 because the document can't be found.

indigo_content_api/v2/urls.py
```diff
@@ -18,18 +17,18 @@
...
+    re_path(r'^(?P<frbr_uri>akn/[a-z]{2}[-/]\S+?)/media\.(?P<format>[a-z0-9]+)$', views.PublishedDocumentMediaView.as_view({'get': 'list'}), name='published-document-media'),
+    re_path(r'^(?P<frbr_uri>akn/[a-z]{2}[-/]\S+?)/media$', views.PublishedDocumentMediaView.as_view({'get': 'list'}), name='published-document-media'),
 
     # Expression details
     # eg. /akn/za/act/2007/98/toc.json
-    re_path(r'^akn/(?P<frbr_uri>[a-z]{2}[-/].*)/toc\.(?P<format>[a-z0-9]+)$', v1_views.PublishedDocumentTOCView.as_view({'get': 'get'}), name='published-document-toc'),
+    re_path(r'^(?P<frbr_uri>akn/[a-z]{2}[-/].*)/toc\.(?P<format>[a-z0-9]+)$', views.PublishedDocumentTOCView.as_view({'get': 'get'}), name='published-document-toc'),
     # eg. /akn/za/act/2007/98
-    re_path(r'^akn/(?P<frbr_uri>[a-z]{2}[-/].*)$', v2_views.PublishedDocumentDetailViewV2.as_view({'get': 'get'}), name='published-document-detail'),
+    re_path(r'^(?P<frbr_uri>(akn/)?[a-z]{2}[-/].*)$', views.PublishedDocumentDetailView.as_view({'get': 'get'}), name='published-document-detail'),
```
longhotsummer, line 31: Why is this one optional?

indigo_content_api/v1/urls.py
goose-life: ah. in that case, some more work needs to be done.

PR #1046, opened 2020-04-23
Django 2.2 LTS Upgrade

ro6ley:
This PR upgrades the Django version from `1.11` to `2.2.12 LTS`.

#### Manual Testing

Reinstall the dependencies and run the tests using:
```
$ pip install -e '.[test]' 
$ python -Wa manage.py test
```

#### Upgraded Packages
- Django updated to version 2.2.12 LTS
- Django-Rest-Framework upgraded to version 3.11.0
- Django-Languages-Plus upgraded to version 1.0.0
- Django-Countries-Plus upgraded to version 1.3.1
- Django-Filter upgraded to version 2.2.0
- Django-Allauth upgraded to version 0.41.0
- Django-Reversion upgraded to version 3.0.1
- Django-Taggit upgraded to version 1.2.0

#### Notable Changes
- `SessionAuthenticationMiddleware` was removed in Django 2 - [reference](https://docs.djangoproject.com/en/2.2/releases/2.0/#miscellaneous)
- `Django-Languages-Plus` introduced a new `languages_data` fixture from version 1.0.0 that must be manually loaded - [reference](https://github.com/cordery/django-languages-plus/blob/master/CHANGELOG.md#100)
- `Django-Reversion` version `3.0.2` squashes previous migrations and  that breaks migrations. Currently, the upgrade is limited to version `3.0.1` which still maintains previous migrations - [reference](https://github.com/etianen/django-reversion/blob/master/CHANGELOG.rst#302---2018-11-05)
- `Django-Rest-Framework` replaces `@detail_route` decorator with `@action` decorator from version `3.10` onwards - [reference](https://www.django-rest-framework.org/community/3.9-announcement/#action-decorator-replaces-list_route-and-detail_route)
- `Django-Rest-Framework` also requires a default value passed for dotted sources in serializers - [reference](https://www.django-rest-framework.org/api-guide/fields/#source)
- `on_delete` values added to models as required by `Django 2+`
- `application names` added to url tuples with namespaces as required by `Django 2+`
- Update url definitions to use `path` and `re_path`


setup.py
```diff
@@ -40,33 +40,33 @@
...
     install_requires=[
-        'django>=1.11.27,<2',
+        'django>=2.2.12,<3',
         'boto3>=1.7',
         'cobalt>=3.1',
         'django-ckeditor>=5.8',
         'dj-database-url>=0.3.0',
         'django-activity-stream>=0.7.0',
-        'django-allauth>=0.36.0,<0.41.0',  # 0.41.0 moved to django 2
+        'django-allauth==0.41.0',
```
longhotsummer, line 49: As a library vendor, it's better to be flexible in the libraries that you support (see https://stackoverflow.com/a/44938662). This means that consumers of your library have flexibility as well. For example, if `0.42.0` came out, a consumer wouldn't be able to upgrade because we have pinned our dependency.

So, this should change to `>=0.41.0` (ditto for other versions that are pinned with `==`).

The flip side of this is in your end application that pulls in a set of libraries, they should be pinned as far as possible, so that you don't get surprise minor version upgrades during your deployments. But that's a different issue :)

ro6ley: I had to pin some of the packages to identify which versions were bringing in breaking changes.  Let me pin the ones that are stable

ro6ley:
Hoping to get those done tomorrow

PR #1047, opened 2020-04-23
Try to work around intermittent JS bug in Firefox

longhotsummer:
There is an intermittent bug (seen mostly with Kenya) where sometimes `Indigo.view` is null, even though we're at a point where it should definitely not be null. I can't reproduce it, but this attempts to work around that.


longhotsummer:
Without this patch, the core bootstrap script assumes that `Indigo` is already in the window namespace, which is bad practice. So check for it explicitly.

PR #1060, opened 2020-05-05
Enhance Place Listing

### LINKED ISSUE #1019: As a manager I want a quick overview of as-at date and completeness in place and locality listings
In https://edit.laws.africa/places/za/localities and https://edit.laws.africa/places, change the place cards to add:

* % complete
* as at date

eg:

Localities – Laws Africa 2020-04-06 13-19-40

I suggest a `row` div with `col` divs with `.text-muted`, then `h5` for the number, and plain text for the description. Use an emdash if there is no as-at date.

ro6ley:
#### Desc
This PR adds as-at date and completeness to overview cards for country and locality listings as can be seen in the screenshot.

#### Manual Testing
Navigate to `/places`, then navigate to the `localities` listing for a particular country.

#### Relevant Issue
closes #1019 

#### Screenshots
Countries listing:
image

Localities listing:
image

### DIFF
--- a/indigo_app/views/places.py
+++ b/indigo_app/views/places.py
@@ -74,9 +74,15 @@ class PlaceListView(AbstractAuthedIndigoView, TemplateView, PlaceMetricsHelper):
 
         context['countries'] = Country.objects\
             .prefetch_related('country')\
-            .annotate(n_works=Count('works'))\
+            .annotate(n_works=Subquery(
+                Work.objects.filter(country=OuterRef('pk'), locality=None)
+                .values('country')
+                .annotate(cnt=Count('pk'))
+                .values('cnt'),
+                output_field=IntegerField()
+            ))\
             .annotate(n_open_tasks=Subquery(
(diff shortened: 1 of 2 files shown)


indigo_app/views/places.py
```diff
@@ -82,6 +82,13 @@ def get_context_data(self, **kwargs):
                 .values('cnt'),
                 output_field=IntegerField()
             ))\
+            .annotate(p_breadth_complete=Subquery(
```
longhotsummer, line 85: nice

longhotsummer:
Great!

indigo_app/views/places.py
```diff
@@ -82,6 +82,13 @@ def get_context_data(self, **kwargs):
                 .values('cnt'),
```
longhotsummer, line 82: This subquery should filter for `locality=None` on the tasks.

longhotsummer: The work count should do the same.

PR #1063, opened 2020-05-06
Bulk creator changes

goose-life:


indigo/bulk_creator.py
```diff
@@ -614,12 +643,16 @@ def create_task(self, work, info, task_type, repealing_work=None):
             task.workflows.set([self.workflow])
             task.save()
 
+        if 'pending-commencement' in task_type:
+            # add the `pending commencement` label
+            p_c_label = TaskLabel.objects.filter(slug='pending-commencement').first()
```
longhotsummer, line 648: This task label may not exist.

longhotsummer:
What's the implication for the spreadsheet for changing from looking up based on frbr-uri vs title?

indigo/bulk_creator.py
```diff
@@ -51,6 +52,30 @@ def clean_title(self):
...
 
+class RowValFormCap(RowValidationFormBase):
+    """ Includes (optional) Chapter (cap) field.
+    For this field to be recorded on bulk creation, add `'cap': 'Chapter (Cap.)'`
+    for the relevant country in settings.INDIGO['WORK_PROPERTIES']
+    """
+    cap = forms.CharField(required=False)
+
+
+class RowValFormPubDateOpt(RowValidationFormBase):
```
longhotsummer, line 63: Style: avoid cryptic abbreviations. `PublicationDateOptional` or `PublicationDateOptionalRowValidationForm` (verbose, but clear). The former is probably sufficient given that it's in the context of the bulk creator. In fact, we could make this a mixin.

Take a look at how Django does its class naming and hierarchy, they use mixins a bunch: https://ccbv.co.uk/

indigo/analysis/work_detail/base.py
```diff
@@ -45,3 +45,18 @@ def work_friendly_type(self, work):
             return _(uri.subtype.upper())
 
         return _('Act')
+
+
+class ChapterWorkDetail(BaseWorkDetail):
```
longhotsummer, line 50: Suggest we change this to a mixin `ChapterMixin`.

indigo/analysis/work_detail/base.py
```diff
@@ -45,3 +45,18 @@ def work_friendly_type(self, work):
...
+    """ Takes Chapter numbers into account when working out numbered title.
+    """
+    def work_numbered_title(self, work):
+        work_type = super().work_friendly_type(work)
+        if work.properties.get('cap'):
```
longhotsummer, line 55: This is the only thing this method needs to do, otherwise it just calls the superclass's method.

indigo/analysis/work_detail/base.py
```diff
@@ -45,3 +45,18 @@ def work_friendly_type(self, work):
...
+            # eg. Chapter 2
+            return f"Chapter {work.properties['cap']}"
```
longhotsummer, line 57: This should be localisable: `return _('Chapter %(cap)s') % {'cap': work.properties['cap']}`

indigo/bulk_creator.py
```diff
@@ -51,15 +52,35 @@ def clean_title(self):
         return re.sub('[\u2028 ]+', ' ', title)
 
 
+class ChapterMixin(forms.Form):
```
longhotsummer, line 55:
```suggestion
class ChapterMixin:
```

longhotsummer: Or does the form not pick up this field without this?

Issue #374, opened 2019-02-14
Indigo uses AKN3 namespace


longhotsummer:
* [x] use new cobalt
* [x] XSLT must change
* [x] Migration to change namespace in all documents (incl. deleted and versions)

longhotsummer:
https://docs.google.com/document/d/1auvOz9juOM2MfVKMRd77Dl6PCtxN6f-nqYvQv6LreQY/edit#

longhotsummer:
I'm working on using the new Cobalt

goose-life:
I'll work on the migration

longhotsummer:
The AKN3 branch now uses the new cobalt.

Issue #364, opened 2019-02-12
Document API shouldn't be able to create documents


longhotsummer:
Document creation is partially split across the indigo_app views, and the indigo_api DocumentViewSet view. We should get rid of the latter, it's not used and is overly complicated.

PR #1066, opened 2020-05-08
Use new cobalt functionality, and more

longhotsummer:
* use new cobalt classes
* a document must always have a work associated with it, so that the work's URI can be used (in future the document's frbr_uri field will be removed -- it's duplicated and unused)
* improve parsing and rendering to take document into account
* lookup XSL on the server, rather than the client, making lookup more flexible

Note: existing XSL files in other indigo packages need to be moved! 

Ref: #374, #364 

Requires https://github.com/laws-africa/cobalt/pull/28


longhotsummer:
Fixing tests, and have created https://github.com/laws-africa/cobalt/pull/30 to ensure we can parse components with bangs.

goose-life:
The build should be passing now that https://github.com/laws-africa/cobalt/pull/30 has been merged, is there a way to trigger a re-try?

longhotsummer:
There was one more test to fix.

PR #1072, opened 2020-05-13
namespace migration

goose-life:
ref #374


indigo_api/migrations/0130_migrate_namespaces.py
```diff
@@ -0,0 +1,55 @@
...
+    migration = UpdateAKNNamespace()
+
+    for document in Document.objects.using(db_alias).all():
+        update = False
+        xml = migration.update_namespace(document.document_xml)
+        if xml != document.document_xml:
+            document.reset_xml(xml, from_model=True)
+            update = True
+
+        if update:
```
longhotsummer, line 28: This implies we can roll this change back. We can't. So we don't need any of this.

indigo_api/data_migrations/__init__.py
```diff
@@ -72,3 +72,25 @@ def strip_whitespace(self, xml):
...
+
+class UpdateAKNNamespace(AKNMigration):
+    """ Update all instances of:
+    <akomaNtoso xmlns="http://www.akomantoso.org/2.0" xsi:schemaLocation="http://www.akomantoso.org/2.0 akomantoso20.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    to
+    <akomaNtoso xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0" xsi:schemaLocation="http://docs.oasis-open.org/legaldocml/ns/akn/3.0 ../akomantoso30.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    """
+    # TODO: Sanity check use and value of xsi:schemaLocation
+    def update_namespace(self, xml):
+        from cobalt import Act
```
longhotsummer, line 84: You could probably just do this with a simple regex replacement on the raw XML. No need to parse it with Cobalt.

longhotsummer: I suggest three steps:

1. remove ` xsi:schemaLocation="http://www.akomantoso.org/2.0 akomantoso20.xsd"`
2. remove ` xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"`
3. change `xmlns="..."`

(note the leading space on the first two)

indigo_api/data_migrations/__init__.py
```diff
@@ -72,3 +72,26 @@ def strip_whitespace(self, xml):
...
 
+
+class UpdateAKNNamespace(AKNMigration):
+    """ Update all instances of:
+    <akomaNtoso xmlns="http://www.akomantoso.org/2.0" xsi:schemaLocation="http://www.akomantoso.org/2.0 akomantoso20.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    to
+    <akomaNtoso xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0">
+    """
+    def update_namespace(self, xml):
+        if "http://www.akomantoso.org/2.0" in xml[:300]:
```
longhotsummer, line 83: this isn't strictly necessary since the replace is a no-op if it can't find the string.

indigo_api/migrations/0130_migrate_namespaces.py
```diff
@@ -0,0 +1,41 @@
...
+    db_alias = schema_editor.connection.alias
+    Document = apps.get_model("indigo_api", "Document")
+    ct_doc = ContentType.objects.get_for_model(Document)
+    migration = UpdateAKNNamespace()
+
+    for document in Document.objects.using(db_alias).all():
+        xml = migration.update_namespace(document.document_xml)
+        if xml != document.document_xml:
+            document.reset_xml(xml, from_model=True)
+            document.save()
```
longhotsummer, line 22:
```suggestion
            document.document_xml = xml
            document.save()
```

indigo_api/data_migrations/__init__.py
goose-life: I know it's _highly_ unlikely, but don't we want to guard against replacing it outside of the first line?

goose-life:
> This PR also needs to tell slaw to use AKN3 namespaces

the Gemfile already specifies gem 'slaw', git: 'https://github.com/longhotsummer/slaw', branch: 'akn3'
is there something else that needs to happen?

indigo_api/data_migrations/__init__.py
longhotsummer: > > This PR also needs to tell slaw to use AKN3 namespaces
> 
> the Gemfile already specifies gem 'slaw', git: 'https://github.com/longhotsummer/slaw', branch: 'akn3'
> is there something else that needs to happen?

Slaw now does AKN3 by default.

longhotsummer:
> > This PR also needs to tell slaw to use AKN3 namespaces
> 
> the Gemfile already specifies gem 'slaw', git: 'https://github.com/longhotsummer/slaw', branch: 'akn3'
> is there something else that needs to happen?

Slaw now does AKN3 by default.

Issue #955, opened 2020-02-21
Indigo must handle new eId AKN3 attributes


goose-life:
## data changes

* [x] Migration to update and move id to eId for historical and deleted documents
* [x] Migration for annotation anchors to use the new IDs
* [x] Migration for internal document refs to use the new IDs
* [x] Migration to create correct term eId attributes

## application logic changes
* [x] XSLT must render id attributes using eId
* [x] annotations etc. must be attached based on eId
* [x] linters must use eId
* [x] internal section refs must use eId
* [x] term tags must use eId and generate appropriate ids
* [x] TLCTerm tags must use eId (but their ids don't need to change)

longhotsummer:
See https://docs.google.com/document/d/1auvOz9juOM2MfVKMRd77Dl6PCtxN6f-nqYvQv6LreQY/edit#

longhotsummer:
## Rendering

* When rendering AKN to HTML, the HTML `id` attribute must come from the new `eId` attribute. The HTML `data-eId` attribute is the value of the `eId` attribute.
* When rendering AKN to HTML for schedules, the scope for the ID prefix will change to the id of the attachment element, rather than the schedule name. for example, `att_1/sec_1`. The renderer in JS and Python must pass in an updated id prefix

longhotsummer:
## Annotations

* When attaching an annotation to element in JS, it must use `eId`. 
* In python, do the same thing for the issue page.

longhotsummer:
# Linters, TOC and issues

* Check that TOC and linter/issue JS uses eId not id

longhotsummer:
# Plugins for finding internal section refs

* use eId not id when looking for referenced provisions

longhotsummer:
For refs: https://github.com/laws-africa/indigo/blob/master/indigo/analysis/refs/base.py#L209

longhotsummer:
Also for refs: https://github.com/laws-africa/indigo/blob/master/indigo/analysis/refs/base.py#L300

longhotsummer:
For terms: https://github.com/laws-africa/indigo/blob/master/indigo/analysis/terms/base.py#L239

PR #1078, opened 2020-05-19
Re-format schedules for AKN3

longhotsummer:
* move under `act`
* component[s] -> attachment[s]
* move headings
* remove hcontainer

PR #1079, opened 2020-05-19
eId migration

goose-life:


indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            old_id = node.get("id")
+            if "." in old_id:
+                node.set("id", re.sub(r"\.", "__", old_id))
+
+        # replace all `id`s with `eId`s
+        for node in doc.root.xpath("//a:*[@id]", namespaces={"a": self.namespace}):
+            node.set("eId", node.get("id"))
+            del node.attrib["id"]
+
+        return doc.to_xml().decode("utf-8")
```
longhotsummer, line 175: I suggest keeping the AKNMigration as the subclass, and not serialising/deserialising. Simply work with `document.act`. We're going to be running multiple data migrations on the same document, they can all work with the parsed form of the XML.

longhotsummer: See https://github.com/laws-africa/indigo/pull/1078 as an example.

indigo_api/migrations/0132_migrate_ids_eIds.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('indigo_api', '0131_migrate_namespaces'),
+    ]
+
+    operations = [
+        migrations.RunPython(forward, migrations.RunPython.noop),
```
longhotsummer, line 40: I think remove the backwards migration, to indicate to Django that it can't go backwards.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            - and then by replacing all `id`s with `eId`s
+        """
+        doc = Act(xml)
+        self.namespace = doc.namespace
+
+        # basic replacements, e.g. "chapter-" to "chap_"
+        for element, patterns in self.basic_replacements.items():
+            for node in doc.root.xpath(f"//a:{element}", namespaces={"a": self.namespace}):
+                old_id = node.get("id")
+                new_id = re.sub(re.compile(patterns[0]), patterns[1], old_id)
```
longhotsummer, line 131: re.sub can take a regex string as its first argument, no need to compile it. Compiling is only useful if you want to re-use the compiled regex.
```suggestion
                new_id = re.sub(patterns[0], patterns[1], old_id)
```

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+    def update_xml(self, xml):
+        """ Update the xml,
+            - first by updating the values of `id` attributes as necessary,
+            - and then by replacing all `id`s with `eId`s
+        """
+        doc = Act(xml)
+        self.namespace = doc.namespace
+
+        # basic replacements, e.g. "chapter-" to "chap_"
+        for element, patterns in self.basic_replacements.items():
```
longhotsummer, line 128: You can be extra fun here:
```suggestion
        for element, (pattern, replacement) in self.basic_replacements.items():
```
and then not have to refer to patterns[0] etc.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            for node in doc.root.xpath(f"//a:{element}", namespaces={"a": self.namespace}):
+                old_id = node.get("id")
+                new_id = re.sub(re.compile(patterns[0]), patterns[1], old_id)
+                node.set("id", new_id)
```
longhotsummer, line 132: You should be able to delegate both these lines to `indigo.xmlutils.rewrite_ids()`

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+                self.update_prefixes(node, old_id, new_id)
+
+        # more complex replacements
+        # TODO: simplify the two below (they're very similar)
+        # subsections
+        for node in doc.root.xpath("//a:subsection", namespaces={"a": self.namespace}):
+            old_id = node.get("id")
+            # e.g. sec_3a.2A.4.1 --> sec_3a.subsec_2A-4-1
+            old_pattern = re.compile(r"^(?P<pre>[^.]+)(?P<post>(\.\d[\dA-Za-z]*)+)")
```
longhotsummer, line 141: Would this be simpler?

```python
pre, post = old_id.split('.', 1)
```

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            match = old_pattern.search(old_id)
+            num = (match.group("post").lstrip(".")).replace(".", "-")
```
longhotsummer, line 143: I suggest making a helper method like slaw has, that cleans up a num to be used in an id. It'll be happening a lot.

indigo_api/data_migrations/__init__.py
longhotsummer: Oh, no, that only works if the prefix doesn't have a dot.

Why not re-build the id completely, use the parent element's id as the prefix, and the content of `num` as the num?

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            node.set("id", new_id)
+            self.update_prefixes(node, old_id, new_id)
+
+        # items
+        for node in doc.root.xpath("//a:item", namespaces={"a": self.namespace}):
+            old_id = node.get("id")
+            # e.g. sec_3a.subsec_2-4-1.list_0.b --> sec_3a.subsec_2-4-1.list_0.item_b
+            # e.g. sec_3a.subsec_2-4-1.list_0.3.1 --> sec_3a.subsec_2-4-1.list_0.item_3-1
+            # or sec_4.subsec_4.list_0.item_c.list_0.i --> sec_4.subsec_4.list_0.item_c.list_0.item_i
+            old_pattern = re.compile(r"""
```
longhotsummer, line 154: As above, I think it's going to be simpler just to get the parent element's id and our num, and re-build the whole id completely.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,84 @@ def update_namespace(self, xml):
...
+            new_id = f"{match.group('pre')}.item_{num}"
+            node.set("id", new_id)
+            self.update_prefixes(node, old_id, new_id)
+
+        # finally
+        # "." separators
+        for node in doc.root.xpath("//a:*[@id]", namespaces={"a": self.namespace}):
+            old_id = node.get("id")
+            if "." in old_id:
+                node.set("id", re.sub(r"\.", "__", old_id))
```
longhotsummer, line 168: You could also you simple string replacement, no regex necessary: `old_id.replace('.', '__')`

indigo_api/data_migrations/__init__.py
longhotsummer: We will need to record all the old -> new mappings in a dictionary on the migration object, so that we can the use those to update the annotation and internal references.

indigo_api/data_migrations/__init__.py
goose-life: nice!

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,176 @@ def update_namespace(self, xml):
...
+            if "." in old_id:
+                node.set("id", old_id.replace(".", "__"))
+
+        # replace all `id`s with `eId`s
+        for node in doc.root.xpath("//a:*[@id]", namespaces=nsmap):
+            node.set("eId", node.get("id"))
+            del node.attrib["id"]
+
+    def clean_number(self, num):
+        return num.lstrip(".").replace(".", "-")
```
longhotsummer, line 266: rstrip as well, pls. "1.2.1." in some subsections.

indigo_api/migrations/0132_migrate_ids_eIds.py
```diff
@@ -0,0 +1,50 @@
...
+    db_alias = schema_editor.connection.alias
+    Document = apps.get_model("indigo_api", "Document")
+    ct_doc = ContentType.objects.get_for_model(Document)
+    para_to_hcontainer = UnnumberedParagraphsToHcontainer()
+    component_to_attachment = ComponentSchedulesToAttachments()
+    eid_migration = AKNeId()
+
+    def update_xml(xml):
+        cobalt_doc = Act(xml)
+        para_to_hcontainer.migrate_act(cobalt_doc)
```
longhotsummer, line 24: Before this, we need to correctly id crossheadings to use hcontainer ids.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +96,191 @@ def update_namespace(self, xml):
...
+            if "." in old_id:
+                node.set("id", old_id.replace(".", "__"))
+
+        # replace all `id`s with `eId`s
+        for node in doc.root.xpath("//a:*[@id]", namespaces=nsmap):
+            node.set("eId", node.get("id"))
+            del node.attrib["id"]
+
+    def clean_number(self, num):
+        return num.strip(".").replace(".", "-")
```
longhotsummer, line 281: This is how slaw does it: https://github.com/longhotsummer/slaw/blob/akn3/lib/slaw/grammars/counters.rb#L37

longhotsummer:
Fixes #955

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +113,239 @@ def update_namespace(self, xml):
...
+            new_ref = self.traverse_mappings(mappings, ref)
+            node.set("href", f"#{new_ref}")
+
+
+class AnnotationsMigration(AKNMigration):
+    """ Update all the annotation anchors on a Document
+    """
+    def migrate_act(self, doc, mappings):
+        for annotation in doc.annotations.all():
+            annotation.anchor_id = self.traverse_mappings(mappings, annotation.anchor_id)
```
longhotsummer, line 350: For a comment in a schedule, the old id will have a format such as `schedule/schedule.paragraph0`, but the new id will not have that schedule prefix.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +128,245 @@ def update_namespace(self, xml):
...
+        # add one to n (previously 0-indexed)
+        for element, (pattern, replacement) in self.add_1_replacements.items():
+            for node in doc.root.xpath(f"//a:{element}", namespaces=nsmap):
+                old_id = node.get("id")
+                num = str(int(pattern.search(old_id).group("num")) + 1)
+                new_id = re.sub(pattern, replacement + num, old_id)
+                self.safe_update(node, mappings, old_id, new_id)
+
+        # more complex replacements
+        for element, (pattern, replacement) in self.complex_replacements.items():
```
longhotsummer, line 303: The pattern can now be removed completely.

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +100,414 @@ def update_namespace(self, xml):
...
+        return components
+
+    def migrate_act(self, doc, document=None):
+        """ Update the xml,
+            - first by updating the values of `id` attributes as necessary,
+            - and then by replacing all `id`s with `eId`s
+        """
+        self.nsmap = {"a": doc.namespace}
+        self.document = document
+        mappings = {}
```
longhotsummer, line 163: use defaultdict for mappings :)

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +100,414 @@ def update_namespace(self, xml):
...
+                    self.safe_update(node, mappings, old_id, new_id, name)
+
+    def term_ids(self, doc, mappings):
+        for name, root in self.components(doc).items():
+            counter = Counter()
+            for node in root.xpath('.//a:term', namespaces=self.nsmap):
+                old_id = node.get('id')
+                prefix = self.get_parent_id(node)
+                # the num for this term depends on the number of preceding terms with the same prefix
+                counter[f"{name}-{prefix}"] += 1
```
longhotsummer, line 436: doesn't need the `name`, counter is reset for each component

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +100,414 @@ def update_namespace(self, xml):
...
+            for node in root.xpath(".//a:*[@id]", namespaces=self.nsmap):
+                old_id = node.get("id")
+
+                if old_id.endswith("."):
+                    log.warning(f"Old id doesn't look right: {old_id}")
+
+                if "." in old_id:
+                    new_id = old_id.replace(".", "__")
+                    node.set("id", new_id)
+                    if mappings.get(name):
```
longhotsummer, line 453: defaultdict for mappings should solve this

indigo_api/data_migrations/__init__.py
```diff
@@ -94,3 +100,414 @@ def update_namespace(self, xml):
...
+                        new_post = self.traverse_mappings(post, mappings[pre])
+                        annotation.anchor_id = f"{pre}/{new_post}"
+                    else:
+                        log.warning(f"This annotation's anchor id hasn't been updated: {old_anchor}")
+                        break
+                elif mappings.get("main"):
+                    annotation.anchor_id = self.traverse_mappings(old_anchor, mappings["main"])
+
+                if annotation.anchor_id == old_anchor:
+                    log.warning(f"This annotation's anchor id hasn't been updated: {old_anchor}")
```
longhotsummer, line 509: pls tweak log messages in this method to be "Not updating..." or similar.

PR #1092, opened 2020-06-03
Attachment subheadings as akn-subheading, don't italicise.

### LINKED ISSUE #1091: Render attachment subheadings as subheadings
Currently, we render both `heading` and `subheading` for an attachment (schedule) as `<h2 class="akn-heading">`, which isn't correct.

Additionally, adjust indigo-web so that for attachments, a subheading isn't italicised.

longhotsummer:
The indigo-web changes also need to be applied in that repo.

Fixes #1091

old:

Air Quality Management - Open By-laws South Africa 2020-06-03 10-33-58

new (different doc):

Water and Sanitation Services @ 2015-09-11 2020-06-03 10-33-40

### DIFF
--- a/indigo_api/static/xsl/html_act.xsl
+++ b/indigo_api/static/xsl/html_act.xsl
@@ -188,8 +188,9 @@
 
   <xsl:template match="a:meta" />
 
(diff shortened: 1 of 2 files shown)


longhotsummer:
@goose-life curious about what you think?

PR #1095, opened 2020-06-04
Refactor akn3 migrations so they are reusable

longhotsummer:


indigo_api/data_migrations/akn3.py
```diff
@@ -0,0 +1,446 @@
...
+            <p>text</p>
+          </content>
+        </hcontainer>
+        """
+        for name, root in self.components(doc).items():
+            for para in root.xpath('.//a:paragraph[not(a:num)]', namespaces=self.nsmap):
+                para.tag = f'{{{doc.namespace}}}hcontainer'
+                # new id is based on the number of preceding hcontainer siblings
+                num = len(para.xpath('preceding-sibling::a:hcontainer', namespaces=self.nsmap)) + 1
+                old_id = para.get('id') or 'paragraph-1'
```
longhotsummer, line 209: added this for case when the para doesn't have an id

goose-life: maybe we should add a log message here, in case it happens somewhere other than in stubs?

PR #1127, opened 2020-06-26
Vastly improved diffs

### LINKED ISSUE #870: better differ
Guidelines:

* an inserted definition should show as such, and not force every subsequent line to "change"
* an inserted subsection should show as such, similar to above
* a heavily changed paragraph should mark the whole thing as changed (?)
* inserting a link should not show the 'tail' of the link as changed

longhotsummer:
Penal Code @ 2018-01-01 2020-06-26 11-38-38

Closes #870

### DIFF
--- a/indigo/analysis/differ.py
+++ b/indigo/analysis/differ.py
@@ -1,20 +1,137 @@
 import html
 from difflib import SequenceMatcher
-from itertools import zip_longest
-from copy import deepcopy
 import logging
+import os
+import re
 
 import jsonpatch
 import lxml.html
 import lxml.html.builder
 from lxml import etree
+from xmldiff import main as xmldiff_main, formatting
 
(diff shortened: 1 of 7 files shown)


indigo/analysis/differ.py
```diff
@@ -1,20 +1,137 @@
...
+    xmldiff_options = {
+        'F': 0.75,
+        # using data-refersto helps to xmldiff to handle definitions that move around
+        'uniqueattrs': ['id', 'data-refersto'],
+        'ratio_mode': 'faster',
+        'fast_match': True,
+    }
+
+    def diff_html(self, old_tree, new_tree):
+        """ Compares to trees, and returns new tree with annotated differences.
```
longhotsummer, line 54:
```suggestion
        """ Compares two trees, and returns new tree with annotated differences.
```

PR #1146, opened 2020-07-20
add `doctypes` getter to add certain doctypes to certain places only

goose-life:


longhotsummer:
Is there existing lightweight place settings functionality that we can use instead of a new plugin? Even if it's just a case of looking up a dictionary on the INDIGO settings object.

indigo_app/views/base.py
```diff
@@ -105,3 +105,12 @@ def has_country_permission(self):
                             "but self.country is None.")
 
         return self.request.user.editor.has_country_permission(self.country)
+
+    def doctypes(self):
+        doctypes = settings.INDIGO['DOCTYPES']
+        extras = settings.INDIGO['EXTRA_DOCTYPES'].get(self.country.code)
```
longhotsummer, line 111: I think you'll have to set the default for this to be an empty dict. Otherwise `settings.INDIGO['EXTRA_DOCTYPES']` raises an exception.

longhotsummer:
Actually, please don't merge. let's move the authorialNote out of this PR.

goose-life:
> Actually, please don't merge. let's move the authorialNote out of this PR.

I've moved that change out

PR #1153, opened 2020-07-30
allow for doctypes, actors to be included on bulk import

goose-life:
(without breaking if they aren't)

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -19,15 +19,16 @@ from indigo_api.models import Subtype, Work, PublicationDocument, Task, Amendmen
     VocabularyTopic, TaskLabel
 from indigo_api.signals import work_changed
 
+SUBTYPES = Subtype.objects.all()
+
 
 class RowValidationFormBase(forms.Form):
-    country = forms.CharField()
-    locality = forms.CharField(required=False)
+    country = forms.ChoiceField(required=True)
+    locality = forms.ChoiceField(required=False)
     title = forms.CharField()
     primary_work = forms.CharField(required=False)
-    subtype = forms.CharField(required=False, validators=[
-        RegexValidator(r'^\S+$', 'No spaces allowed.')
-    ])
+    doctype = forms.ChoiceField(required=True)
+    subtype = forms.ChoiceField(required=False)
     number = forms.CharField(validators=[
         RegexValidator(r'^[a-zA-Z0-9-]+$', 'No spaces or punctuation allowed (use \'-\' for spaces).')
     ])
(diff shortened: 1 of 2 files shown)


indigo/bulk_creator.py
```diff
@@ -334,6 +336,11 @@ def validate_row(self, view, row):
         form = self.get_row_validation_form(row_data)
```
longhotsummer, line 336: You could delegate all the validation to the RowValidationFormBase. It would need the country to be passed to the form constructor (which would be useful anyway). You can then make doctype a ChoiceField, and in the form's constructor update the `choices` attribute as appropriate. This is what the form is there to do, after all.

indigo/bulk_creator.py
```diff
@@ -19,15 +19,16 @@
     VocabularyTopic, TaskLabel
 from indigo_api.signals import work_changed
 
+SUBTYPES = Subtype.objects.all()
```
longhotsummer, line 22: Remember, .all() goes to the database. So this will load subtypes once and never pick up any changes. Instead, do this where you need them on line 57.

indigo/bulk_creator.py
```diff
@@ -47,6 +48,19 @@ class RowValidationFormBase(forms.Form):
...
+    def __init__(self, country, locality, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['country'].choices = [(country.code, country.name)]
+        self.fields['locality'].choices = [(locality.code, locality.name)] \
+            if locality else []
+        self.fields['doctype'].choices = self.get_doctypes_for_country(country.code)
+        self.fields['subtype'].choices = [(s.abbreviation, s.name) for s in SUBTYPES]
+
+    def get_doctypes_for_country(self, country_code):
+        return [([d[1], d[0]]) for d in
```
longhotsummer, line 60: Is this correctly nested? It's going to produce, eg: `[(['x', 'y'], ['a', 'b'])]` -- I think you have too many brackets.

longhotsummer: Oh, no, it actually works because the parentheses don't land up making a tuple. But they can be removed completely.

```suggestion
        return [[d[1], d[0]] for d in
```

indigo_app/templates/indigo_api/work_new_batch.html
```diff
@@ -138,7 +138,7 @@ <h4 class="card-header">
                 <td>
                   {% if work.work.subtype %}
                     {{ work.work.subtype }}
-                  {% else %}
+                  {% elif work.status != 'error' %}
                     Act
```
longhotsummer, line 142: Does Act not need to change?

indigo/bulk_creator.py
```diff
@@ -47,6 +48,19 @@ class RowValidationFormBase(forms.Form):
     repealed_by = forms.CharField(required=False)
     taxonomy = forms.CharField(required=False)
 
+    def __init__(self, country, locality, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['country'].choices = [(country.code, country.name)]
+        self.fields['locality'].choices = [(locality.code, locality.name)] \
+            if locality else []
+        self.fields['doctype'].choices = self.get_doctypes_for_country(country.code)
+        self.fields['subtype'].choices = [(s.abbreviation, s.name) for s in SUBTYPES]
```
longhotsummer, line 57: Hmm, if we load the subtypes here we're going to go to the db once for every row, which is lame.

indigo/bulk_creator.py
goose-life: I was trying to avoid going to the db for each row in the spreadsheet given that subtypes are universal and relatively stable. But if a subtype is added and the page is refreshed, we'd definitely want that to be picked up :)

goose-life: Maybe I can get the subtypes once per spreadsheet and pass them along? Do you think that'll be worth the saving?

longhotsummer: Yes, good idea.

indigo_app/templates/indigo_api/work_new_batch.html
goose-life: quite right, thanks!

PR #1155, opened 2020-08-05
improve English subtypes refs finder, add tests

goose-life:


indigo/analysis/refs/base.py
```diff
@@ -91,44 +92,45 @@ def setup(self, root):
...
 
     def setup_subtypes(self):
-        self.subtypes = [s for s in Subtype.objects.all()]
-        self.subtype_names = [s.name for s in self.subtypes]
-        self.subtype_abbreviations = [s.abbreviation for s in self.subtypes]
-
-        self.subtypes_string = '|'.join([re.escape(s) for s in self.subtype_names + self.subtype_abbreviations])
+        self.full_subtypes = Subtype.objects.all()
+        self.subtypes = [s.name for s in self.full_subtypes] + \
+                        [s.abbreviation for s in self.full_subtypes]
```
longhotsummer, line 97: Given what we've learnt from term matching and italics, does it make sense to sort these longest first?

indigo/analysis/refs/base.py
```diff
@@ -91,44 +92,45 @@ def setup(self, root):
...
         self.pattern_re = re.compile(
             fr'''
                 (?P<ref>
-                    (?P<subtype>{self.subtypes_string})\s*
+                    (?P<subtype>{subtypes_for_re})\s*
                     (No\.?\s*)?
                     (?P<num>\d+)
-                    (\s+of\s+|/)
-                    (?P<year>\d{{4}})
+                    .{{0,60}}(\s|/)
```
longhotsummer, line 114: This makes me a bit nervous, it's pretty broad.

indigo/tests/test_refs.py
```diff
@@ -814,9 +818,91 @@ def test_find_simple(self):
...
+            work=self.work,
+            document_xml=document_fixture(
+                xml="""
+        <section eId="sec_1">
+          <num>1.</num>
+          <heading>Tester</heading>
+          <paragraph eId="sec_1.paragraph-0">
+            <content>
+              <p>Something to do with <ref href="/akn/za/act/gn/2012/102">GN no 102 published on 5 March 2012</ref>.</p>
+              <p>And another thing about <ref href="/akn/za/act/si/1998/4">SI 4, published in Government Gazette 12345 of 31 January 1998</ref>.</p>
```
longhotsummer, line 853: Maybe this should only markup the `GN no 102` and `SI 4`?

indigo/analysis/refs/base.py
longhotsummer: Neil had an interesting idea which maybe we need to start exploring: start with a broad regex, and then filter down edge cases with supplemental regexes. His argument is it helps to keep the regexes simple and understandable.

longhotsummer: eg. this would be a false positive: `as published in GN No 102. In section 4378 of the Act...`

This stuff is tricky.

PR #1156, opened 2020-08-06
Kenya law

goose-life:


setup.py
```diff
@@ -40,7 +40,7 @@
 
     python_requires='~=3.6',
     install_requires=[
-        'django>=2.2.12,<3',
+        'django==2.2.12',
```
longhotsummer, line 43: I think we should ignore this change. 2.2.13 was broken, but 2.2.14 was released shortly thereafter to fix what 2.2.13 broke. Pinning this means consumers can't pull in the latest security fixes, which is bad.

longhotsummer, line 43:
```suggestion
        'django>=2.2,<3',
```

longhotsummer: I've adjusted it to `>=2.2,<3`

PR #1157, opened 2020-08-12
Use attachments as components

### LINKED ISSUE #1129: TOC should include ids for attachments
Also, should the element be the attachment (including the heading), or the doc? Most likely the attachment object, which has an eId.

longhotsummer:
Use `attachment`/`component` as component elements, rather than the `doc` element.

Closes #1129.

### DIFF
--- a/indigo_api/tests/fixtures.py
+++ b/indigo_api/tests/fixtures.py
@@ -84,15 +84,15 @@ COMPONENT_FIXTURE = """<?xml version="1.0" encoding="UTF-8"?>
     </body>
     <attachments>
       <attachment eId="att_1">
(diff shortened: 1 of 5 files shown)


indigo_api/tests/fixtures.py
```diff
@@ -85,25 +85,27 @@
     <attachments>
       <attachment eId="att_1">
         <doc name="schedule">
+          <heading>Schedule 1</heading>
+          <subheading>Subheading</subheading>
```
goose-life, line 89: I believe these should be above (and siblings of) `doc`? see e.g. https://edit.laws.africa/documents/2460/

PR #1172, opened 2020-09-14
version 11.0.1 changelog

goose-life:


longhotsummer (commented):
We roughly follow [semver](https://semver.org/), which would make this a minor release, rather than a patch release. ie. 11.1.0 vs 11.0.1.

longhotsummer (approved):

PR #1175, opened 2020-09-30
No footnotes in toc, styles

goose-life:
closes https://github.com/laws-africa/indigo-lawsafrica/issues/254

### DIFF
--- a/indigo_app/static/javascript/indigo/views/document_toc.js
+++ b/indigo_app/static/javascript/indigo/views/document_toc.js
@@ -94,9 +94,26 @@
           qualified_id = $component.attr('eId') + '/' + qualified_id;
         }
 
+        function getHeadingText(node) {
+          let heading = node.children('heading')[0];
+          // if the heading includes an authorial note / footnote, only use the text nodes (if any)
+          if (heading && heading.getElementsByTagName('authorialNote')) {
+            let children = heading.childNodes,
+                headingText = '';
+            children.forEach(function(child){
+              if (child.nodeType === Node.TEXT_NODE) {
+                headingText += child.nodeValue;
+              }
+            });
+            return headingText;
+          } else {
+            return node.children('heading').text();
+          }
+        }
+
         var item = {
           'num': $node.children('num').text(),
-          'heading': $node.children('heading').text(),
+          'heading': getHeadingText($node),
           'element': node,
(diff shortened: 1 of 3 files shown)


indigo_app/static/javascript/indigo/views/document_toc.js
```diff
@@ -94,9 +94,26 @@
           qualified_id = $component.attr('eId') + '/' + qualified_id;
         }
 
+        function getHeadingText(node) {
+          var heading = node.children('heading')[0];
```
longhotsummer, line 98: I don't think `ParentNode.children()` accepts parameters: https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/children - or is my understanding of `node` incorrect?

I still think xpath is the best way to do this...

indigo_app/static/javascript/indigo/views/document_toc.js
```diff
@@ -94,9 +94,26 @@
...
+            var children = heading.childNodes,
+                headingText = '';
+            children.forEach(function(child){
+              if (child.nodeType === Node.TEXT_NODE) {
+                headingText += child.nodeValue;
+              }
+            });
+            return headingText;
+          } else {
+            return node.children('heading').text();
```
longhotsummer, line 110: I don't think `.text()` is a valid method...

indigo_app/static/javascript/indigo/views/document_toc.js
```diff
@@ -94,9 +94,26 @@
...
+          }
+        }
+
         var item = {
           'num': $node.children('num').text(),
-          'heading': $node.children('heading').text(),
+          'heading': getHeadingText($node),
```
longhotsummer, line 116: Oooh, it's a jquery object.

indigo_app/static/javascript/indigo/views/document_toc.js
longhotsummer: Oh, node is a jquery object.

longhotsummer: Perhaps pass in a plain node object to this method, and then locally wrap it in a jquery object so that it's clearer what's going on.

longhotsummer (approved):
Awesome.

PR #1177, opened 2020-10-02
check that URI isn't this document's URI in BaseRefsFinder

goose-life:


longhotsummer (dismissed):
Do you want to add a test?

longhotsummer (approved):

PR #1178, opened 2020-10-02
No numbered title numbers

goose-life:


indigo/analysis/work_detail/base.py
```diff
@@ -22,7 +22,8 @@ def work_numbered_title(self, work):
         """
         # these don't have numbered titles
         number = work.number
-        if number in self.no_numbered_title_numbers:
+        # also check partials, e.g. 'constitution-amendment' shouldn't get one
+        if any([partial in self.no_numbered_title_numbers for partial in number.split("-")]):
```
longhotsummer, line 26:
```suggestion
        if any(partial in self.no_numbered_title_numbers for partial in number.split("-")):
```

Any takes an iterable, and a list comprehension is an iterable. It doesn't need to be forced to a list first.

longhotsummer: This makes it efficient, because it the entire list doesn't need to be calculated if one of the values evaluates to True. It'll only consume up to that point, and the remainder is never evaluated.

longhotsummer (approved):

PR #1187, opened 2020-10-23
No need to call old resize on monaco editor, it resizes automatically

longhotsummer:
Fixes https://sentry.io/organizations/lawsafrica/issues/1975500423/


goose-life (approved):
nice!

PR #1191, opened 2020-10-26
Watch dom for changes; resize; prevent unnecessary updates to xml

longhotsummer:
* add a resize observer to re-layout the xml editor when the window changes
* add a guard to prevent the onDidChangeModelContent from firing when we're force updating the content when rendering the XML from the dom tree
* watch the dom model and ensure what is rendered is kept up to date with changes


goose-life (dismissed):

longhotsummer:
The `this.updating` guard is now used only to prevent the onDidChangeModelContent event handler from firing when calling `setValue()`.

We guard against updating the editor content unnecessarily by checking if the content has changed.

goose-life (approved):

PR #1194, opened 2020-10-30
Squash migrations

goose-life:
Questions:
1. I'm not sure whether all the changes I made before squashing (elidable=True and the renaming of dependencies in files that ended up squashed) need to be committed?
2. It feels a bit clunky to have chunked the squashes the way I have; the big squashed migration from 73 to 135 includes lots of Alters that could (maybe?) have been included in the Creation. Is it worthwhile to manually separate out the circular dependencies as recommended in https://coderbook.com/@marcus/how-to-squash-and-merge-django-migrations/ (and https://docs.djangoproject.com/en/2.2/topics/migrations/#migration-squashing)? (I think it might be!)


indigo_api/migrations/0032_search.py
```diff
@@ -43,7 +43,8 @@ class Migration(migrations.Migration):
         ),
         migrations.RunSQL(
             "CREATE INDEX indigo_api_document_search_vector_ix ON indigo_api_document USING gin(search_vector)",
-            "DROP INDEX indigo_api_document_search_vector_ix"
+            "DROP INDEX indigo_api_document_search_vector_ix",
+            elidable=True
```
longhotsummer, line 47: These migrations aren't elidable. They setup state for search in the DB which cannot be done any other way.

longhotsummer: I have partial code to split this out into a separate package, but it's not ready yet. I can take a look at it this afternoon.

longhotsummer: In master, this has all been pulled out of indigo (into indigo-search-psql) so, assuming we'll be merging master into this branch soon, it doesn't matter that we've removed these operations. So this is now fine.

longhotsummer:
PR #1199 should fix the build and bring up to date with master.

However, before we merge, I'd like to see if we can really squash away the remaining migrations. The weird interdependency is from when I moved models from indigo_app to indigo_api (or vv). In theory, we should be able to magic that away in the squashed migration, because it'll effectively create the model in the correct up in the first place. I haven't looked into how we can get it to do that correctly.

longhotsummer (approved):

longhotsummer (approved):

PR #1202, opened 2020-11-06
Set version to 13.1.0

longhotsummer:
This allows us to ensure that dependencies can be explicit about which version they're depending on. I have already updated the v12.0.0 and v13.0.0 tags to have matching versions.


goose-life:
Thanks for this!

PR #1206, opened 2020-11-12
Spreadsheet export

goose-life:
This PR includes a new export of all works in a place, including all active and passive relationships, for the purpose of exporting an up-to-date legislation spreadsheet that can be maintained going forward.


longhotsummer:
Can you provide a quick overview of what this does? Does it just refactor and test the existing export?

setup.py
```diff
@@ -75,13 +75,15 @@
...
         'mammoth>=1.4.4',
+        'pandas>=1.1.4',
         'psycopg2==2.7.3.2',  # 2.7.4 began using psycopg2-binary
         'requests>=2',
         'unicodecsv>=0.14.1',
         'whitenoise<2,>=1.0.6',
         'django-contrib-comments>=1.9.1',
         'XlsxWriter>=1.2.6',
         'xmldiff>=2.4',
+        'xlrd>=1.2.0',
```
longhotsummer, line 86: Is this used?

setup.py
```diff
@@ -75,13 +75,15 @@
         'libsass==0.14.2',  # 0.14.3 upwards changes imports for .css
         'lxml>=3.4.1',
         'mammoth>=1.4.4',
+        'pandas>=1.1.4',
```
longhotsummer, line 78: If this is only required for tests, it should go under the `test` section of `extras_require` below. Pandas is chunky to install.

indigo_app/templates/place/works.html
```diff
@@ -161,6 +161,14 @@
             </div>
           </div>
         {% endfor %}
+
+        {% if perms.indigo_api.bulk_export_work %}
```
longhotsummer, line 165: As discussed, let's move this into the settings view and use the settings permissions to guard it.

indigo_app/views/places.py
```diff
@@ -301,11 +301,14 @@ def get(self, request, *args, **kwargs):
         self.form.is_valid()
 
         if params.get('format') == 'xlsx':
-            return self.generate_xlsx()
-        
+            self.full_index = params.get('full-index')
```
longhotsummer, line 304: Once the button has moved into the settings view, the logic for the full export could move to a separate view completely. There's really no link between that export and this view/our existing xlsx export.

setup.py
goose-life: tests failed to run without it?

longhotsummer: Huh. If you can determine it's a test dependency, then it should go under that section. Otherwise, it's a bit odd that it has showed up now (maybe pandas depends on it, although I think not?)

indigo_app/urls.py
```diff
@@ -25,6 +25,7 @@
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/activity$', places.PlaceActivityView.as_view(), name='place_activity'),
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/metrics$', places.PlaceMetricsView.as_view(), name='place_metrics'),
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/settings', places.PlaceSettingsView.as_view(), name='place_settings'),
+    re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/works_index', places.PlaceWorksIndexView.as_view(), name='works_index'),
```
longhotsummer, line 28: An underscore in a URL can usually semantically be replaced by a slash, since it's effectively scoping it. Perhaps `.../works/index.xlsx`?

goose-life: I like it!

longhotsummer (approved):

PR #1208, opened 2020-11-13
Fix taxonomy import

goose-life:
Our latest topic, 'Freedom of Association', doesn't import from the spreadsheet because of a previously undiscovered bug

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -530,7 +530,7 @@ class BaseBulkCreator(LocaleBasedMatcher):
                                  amendment=amendment)
 
     def link_taxonomy(self, row):
-        topics = [x.strip(',') for x in row.taxonomy.split(', ')]
+        topics = [x.strip() for x in row.taxonomy.split(',') if x.strip()]
         unlinked_topics = []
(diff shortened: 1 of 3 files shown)


indigo/bulk_creator.py
```diff
@@ -530,7 +530,7 @@ def link_amendment_active(self, row):
                                  amendment=amendment)
 
     def link_taxonomy(self, row):
-        topics = [x.strip(',') for x in row.taxonomy.split()]
+        topics = [x.strip(',') for x in row.taxonomy.split(', ')]
```
longhotsummer, line 533: Split on `,` only so that `foo,bar` isn't ignored, and then trim. After splitting on a comma, the comma won't appear anywhere in the string.

```
>>> "foo,bar ,  baz  , boom,,, a".split(",")
['foo', 'bar ', '  baz  ', ' boom', '', '', ' a']
>>> [x.strip() for x in "foo,bar ,  baz  , boom,,, a".split(",") if x.strip()]
['foo', 'bar', 'baz', 'boom', 'a']
```

longhotsummer (approved):

PR #1209, opened 2020-11-13
add Taxonomist badge

goose-life:
Note: The user will still need to be 'staff' to work with taxonomies

image


longhotsummer (approved):
Can't we name this Taxidermist?

PR #1214, opened 2020-12-01
Filter tasks by type

### LINKED ISSUE #1213: As a reviewer I want to see all signoff tasks across the platform in one view
Proposal:
- Add a permission?
- At https://edit.laws.africa/tasks/ add a 'Signoff' tab, only visible to users with the right permission (include count in tab if possible)
- Add `tasks/signoff/` URL and corresponding view, similar to AvailableTasksView (`AvailableSignoffTasksView`)
- Queryset will filter all tasks by title (this will change once tasks are modelled by type): 'Review and publish changes' and 'Review and sign-off changes'
- Template will look similar to https://edit.laws.africa/tasks/available/ but filter isn't needed (title probably isn't either; we could just link the task number instead?)

goose-life:
closes #1213 

image

Select
image

Clear
image

### DIFF
--- a/indigo_app/templates/indigo_app/tasks/available_tasks.html
+++ b/indigo_app/templates/indigo_app/tasks/available_tasks.html
@@ -69,7 +69,12 @@
               <i class="fas fa-sm fa-fw task-icon-pending_review"></i> Pending review
             </label>
           </div>
-
+          <select class="form-control selectpicker" id="{{ form.type.id_for_label }}" name="{{ form.type.html_name }}"
+                  data-live-search="true" data-size="7">
+            {% for code, title in form.fields.type.choices %}
+              <option value="{{ code }}" {% if code in form.type.value %}selected{% endif %}>{{ title }}</option>
+            {% endfor %}
+          </select>
         </div>
       </form>
     </div>
--- a/indigo_app/templates/indigo_app/tasks/available_tasks.html
+++ b/indigo_app/templates/indigo_app/tasks/available_tasks.html
@@ -69,7 +69,9 @@
               <i class="fas fa-sm fa-fw task-icon-pending_review"></i> Pending review
             </label>
           </div>
-          <select class="form-control selectpicker col-2 ml-2" id="{{ form.type.id_for_label }}" name="{{ form.type.html_name }}"
+          <select class="form-control selectpicker notooltip col-2 ml-2"
+                  id="{{ form.type.id_for_label }}" name="{{ form.type.html_name }}"
+                  data-style="btn-outline-success" title="Task type"
                   data-live-search="true" data-size="7" multiple>
             {% for opt in form.type %}
               {{ opt }}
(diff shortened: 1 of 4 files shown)


indigo_app/templates/indigo_app/tasks/available_tasks.html
```diff
@@ -69,7 +69,12 @@ <h4>Available tasks <span class="badge badge-light">{{ paginator.count }}</span>
               <i class="fas fa-sm fa-fw task-icon-pending_review"></i> Pending review
             </label>
           </div>
-
+          <select class="form-control selectpicker col-2 ml-2" id="{{ form.type.id_for_label }}" name="{{ form.type.html_name }}"
```
longhotsummer, line 72: Perhaps make this a multiple select field (eg. a user probably doesn't care if the commencement linking is active or passive).

Will require the `multiple` keyword here and a change to the form type in Django.

indigo_app/templates/indigo_app/tasks/available_tasks.html
```diff
@@ -69,7 +69,12 @@ <h4>Available tasks <span class="badge badge-light">{{ paginator.count }}</span>
...
+                  data-live-search="true" data-size="7">
+            {% for code, title in form.fields.type.choices %}
```
longhotsummer, line 74: Try this, which is simpler:

```
{% for opt in form.type %}
  {{ opt }}
{% endfor %}
```

indigo_app/templates/indigo_app/tasks/available_tasks.html
goose-life: hmm. that does make it a bit less neat in terms of clearing it, but you're right. maybe the codes should be fewer and we can differentiate between passive / active ones some other way?

goose-life: but let's just get this out for the time being, we can improve on it once people use it :grin:

goose-life:
image

indigo_app/templates/indigo_app/tasks/available_tasks.html
longhotsummer: See "Select/Deselect All" option here: https://developer.snapappointments.com/bootstrap-select/examples/

longhotsummer: If you set the `title` attribute then it will show a decent placeholder, and you don't need the blank/all option (which doesn't make sense when you can select multiple options). (see Placeholder here https://developer.snapappointments.com/bootstrap-select/examples/)

goose-life:
image

indigo_app/templates/indigo_app/tasks/available_tasks.html
goose-life: Thanks so much! This is done now

longhotsummer (approved):

PR #1218, opened 2020-12-17
Current commenceable provisions only

goose-life:


indigo_api/models/works.py
```diff
@@ -234,11 +234,14 @@ def first_commencement_date(self):
         if first:
             return first.date
 
-    def commenceable_provisions(self):
+    def commenceable_provisions(self, current=False, date=None):
```
longhotsummer, line 237: I think we can drop current and just use date

longhotsummer: Perhaps also adjust the docstring:
```
If `date` is provided, only provisions in expressions up to and including that date are included.
```

indigo_api/exporters.py
```diff
@@ -87,9 +87,15 @@ def render_coverpage(self, document):
     def get_context_data(self, **kwargs):
         """ Get the context data passed to the HTML template.
         """
+        # only commencements up to the current point in time should be considered for the coverpage
+        commencements = None
+        if kwargs.get('document'):
+            document = kwargs.get('document')
+            commencements = document.work.commencements.filter(date__lte=document.expression_date)
```
longhotsummer, line 94: This doesn't do quite what we want. Instead, I think we need a helper  method on the work which (optionally) filters commencements to only those with provisions at or before a particular date.

indigo_api/models/works.py
```diff
@@ -262,6 +266,18 @@ def uncommenced_provisions(self):
 
         return [p for p in provisions if p.id not in commenced]
 
+    def commencements_relevant_to_current_expression(self, date):
```
longhotsummer, line 269: If this is on a work, the word `current` is confusing, since the work doesn't have a current expression. Perhaps just `commencements_relevant_at_date`

indigo_api/models/works.py
```diff
@@ -262,6 +266,18 @@ def uncommenced_provisions(self):
...
+        """ Return a list of Commencement objects that have to do with the provisions that exist on this work at the given date.
+        """
+        # common case: one commencement that covers all provisions
+        for commencement in self.commencements.all():
+            if commencement.all_provisions:
+                return [commencement]
+
+        commenceable_provisions = [p.id for p in self.commenceable_provisions(date)]
+        # include commencement if any of its `provisions` are found in `commenceable_provisons`
+        return [c for c in self.commencements.all() if any(p for p in c.provisions if p in commenceable_provisions)]
```
longhotsummer, line 279: Nice!

indigo_api/templatetags/indigo.py
```diff
@@ -63,12 +63,18 @@ def end_current():
...
         provisions = commencement.provisions
 
-    commenceable_provisions = work.commenceable_provisions()
+    commenceable_provisions = work.commenceable_provisions(date=document.expression_date)
 
     return make_beautiful(provisions, commenceable_provisions)
+
+
+@register.simple_tag
+def commencements_relevant_to_current_expression(document):
```
longhotsummer, line 79: Again, I think `current` is confusing, because there's no implicit current expression -- there's only the one passed in.

indigo_api/models/works.py
goose-life: good point, thanks! refactored throughout

longhotsummer (approved):
Great tests.

PR #1219, opened 2020-12-18
Bulk creator more

goose-life:
This PR would allow us to import both sides of the relationships exported in the new spreadsheet export.
It will also enable users to make a spreadsheet from a consolidation or maintain a spreadsheet going forward, without having to change spreadsheets.


goose-life:
I give up! :sweat_smile: I'll look at this next year

longhotsummer (approved):

PR #1227, opened 2021-01-12
current commenceable provisions again

goose-life:


longhotsummer (dismissed):

longhotsummer:
This looks good, I just want to confirm we can tie into it correctly.

indigo_api/models/works.py
```diff
@@ -282,15 +282,15 @@ def commencements_relevant_at_date(self, date):
             if commencement.all_provisions:
```
longhotsummer, line 282: It looks like the only code that calls `commencements_relevant_at_date()` is the template tag. I think it would be simpler if this method moved onto the `DocumentMixin` class rather than going document -> work -> document. After all, this is all to do with commencements for a particular document -- the date is actually an implementation detail.

goose-life:
image

image

longhotsummer (approved):

PR #1229, opened 2021-01-18
Catch and log annotation binding errors

longhotsummer:
Fixes an issue preventing https://edit.laws.africa/documents/5023/ from loading cleanly because an annotation can't be attached.


longhotsummer:
The root cause is somewhere textQuoteToRange which is 3rd party.

goose-life (approved):

PR #1235, opened 2021-01-20
Per-place setting for 'no publication document' text

### LINKED ISSUE #1233: Change wording of unverified notice for law reform compiled consolidations
Change the wording of the notice on acts digitized from official consolidations: "[Note: The original publication document is not available and this content could not be verified.]" to wording that reflects the fact that this is an official and authorised (hence verified by the Law Reform Commission) version of legislation

goose-life:
closes #1233 

TODO: will `blocktrans` work here or always display the saved string? UPDATE: the latter, using `trans` instead

Form:
image

Coverpage English:
image

Coverpage Sotho:
image


indigo_api/templates/indigo_api/akn/coverpage_act.html
```diff
@@ -192,7 +192,7 @@ <h2>{{ subt }}</h2>
         {% if not document.work.publication_document %}
           {% block verification_notice %}
             <li class="verification-notice">
-              [{% blocktrans %}Note: The original publication document is not available and this content could not be verified.{% endblocktrans %}]
+              [{% blocktrans with no_publication_document_text=document.work.place.settings.no_publication_document_text %}{{ no_publication_document_text }}{% endblocktrans %}]
```
longhotsummer, line 195: I don't think blocktrans is quite right, see here: https://stackoverflow.com/questions/6546933/django-translate-variable-content-in-template. We want it to translate the contents of the variable, not treat it as a translatable string containing a variable.

We should also check if there is no message and not include this notice at all - ie. provide a way to turn it off.

indigo_api/models/places.py
```diff
@@ -173,6 +173,7 @@ class PlaceSettings(models.Model):
     spreadsheet_url = models.URLField(null=True, blank=True)
     as_at_date = models.DateField(null=True, blank=True)
     styleguide_url = models.URLField(null=True, blank=True)
+    no_publication_document_text = models.CharField(max_length=1024, default='Note: The original publication document is not available and this content could not be verified.')
```
longhotsummer, line 176: `null=False, blank=True` ? So we always want a string, but it can be blank.

goose-life:
Coverpage Sotho, after adding the following to st/LC_MESSAGES/django.po:
```
msgid "Note: The original publication document is not available and this content could not be verified."
msgstr "S Sotho translation"
```

image

longhotsummer (approved):

longhotsummer (changes requested):
One more thing: in order for Django to realise that that string is translatable (and therefore present it for translation in CrowdIn), we need to mark it as translatable and [run makemessages](https://github.com/laws-africa/indigo#adding-translation-strings).

See details at https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#lazy-translation.

Basically, import `ugettext_lazy` as `_` and apply it to the default string in the model definition `_('Note: blah blah')`.

Then run makemessages as documented in Indigo's readme and check that it shows up.

longhotsummer (approved):

PR #1242, opened 2021-02-03
Block tasks

### LINKED ISSUE #1140: as a senior editor I want to mark a task as 'blocked' until another task has been done
e.g. on https://edit.laws.africa/works/akn/za/act/1957/33/, 
the first amendment task is blocked by the import task,
the second amendment task is blocked by the first,
etc.

### Suggestion
- Add a 'blocked by' field on a task, that takes one or more task numbers as its input
- Disable such a task from being picked up and automatically add a `blocked` label
- Once the blocking tasks have been done, make the task pick up able and remove the label

goose-life:
closes #1140 

TODO:
- [ ] add date to task edit form (CANCELLED)
- [x] disable Unblock in edit form while a task is blocked_by other tasks
- [x] move Blocked tasks to first column
- [ ] filter potential blocking tasks to tasks on this work and related works (currently all in place) (CANCELLED)
- [ ] do magic during bulk import (DO SEPARATELY)
- [x] update blocked tasks when a blocking task is closed
- [ ] enable bulk update of task states

A blocked task:
image
image
image

A task blocked by another task:
image
image
Deselect blocking task from Blocked by list to unblock:
image

A blocking task:
image

Blocked tasks in lists:
image
image
image

Blocked task stats in work list:
image


indigo_api/models/tasks.py
```diff
@@ -280,6 +287,26 @@ def close(self, user, **kwargs):
         # send task_closed signal
         task_closed.send(sender=self.__class__, task=self)
 
+    # block
+    def may_block(self, view):
+        return view.request.user.is_authenticated and \
+               view.request.user.editor.has_country_permission(view.country) and \
+               view.request.user.has_perm('indigo_api.block_task')
+
+    @transition(field=state, source=['open'], target='blocked', permission=may_block)
```
longhotsummer, line 296: Just noting that this doesn't allow blocking a task that is pending review.

indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -3,6 +3,23 @@
 <div class="container mt-3 mb-5">
     <div class="row">
       <div class="mb-3 col-9">
+        {% if task.state == 'blocked' %}
+          <div class="alert alert-danger">
+            This task is blocked.
+            {% with task.blocked_by.all as blocked_by %}
+              {% if blocked_by %}
+                These tasks must be completed first:
+                <ul class="mb-2">
```
longhotsummer, line 12:
```suggestion
                <ul class="mt-2">
```

The alert has inner padding, so rather give some space between the intro sentence and the list.

indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -242,6 +259,41 @@ <h6>Labels</h6>
           </ul>
         {% endwith %}
 
+        {% with task.blocked_by.all as blocked_by %}
```
longhotsummer, line 262: You're loading blocked_by twice, rather do it once (saves a DB trip)

indigo_app/urls.py
```diff
@@ -36,11 +36,14 @@
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/assign', tasks.TaskAssignView.as_view(), name='assign_task'),
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/unassign', tasks.TaskAssignView.as_view(unassign=True), name='unassign_task'),
     re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/projects', tasks.TaskChangeWorkflowsView.as_view(), name='task_workflows'),
+    re_path(r'^places/(?P<place>[a-z]{2}(-[^/]+)?)/tasks/(?P<pk>\d+)/block_by_task', tasks.TaskChangeBlockingTasksView.as_view(), name='task_blocked_by'),
```
longhotsummer, line 39: Perhaps `.../blocking-tasks`? I prefer dashes over underscores in URLs.

indigo_app/views/tasks.py
```diff
@@ -136,6 +136,13 @@ def get_context_data(self, **kwargs):
 
         context['possible_workflows'] = Workflow.objects.unclosed().filter(country=task.country, locality=task.locality).all()
 
+        # TODO: other tasks on this and related works (if the task has a work, otherwise none)
```
longhotsummer, line 139: I think safest to be any task in this place.

indigo_app/views/tasks.py
```diff
@@ -398,6 +411,38 @@ def get_redirect_url(self):
...
+    http_method_names = ['post']
+
+    def post(self, request, *args, **kwargs):
+        task = self.get_object()
+        user = self.request.user
+        task.updated_by_user = user
+        ids = self.request.POST.getlist('blocked_by')
+
+        if ids:
+            blocked_by = Task.objects.filter(id__in=ids).all()
```
longhotsummer, line 427: Also filter by the current country and locality.

indigo_app/views/tasks.py
```diff
@@ -398,6 +411,38 @@ def get_redirect_url(self):
...
+        else:
+            blocked_by = []
+
+        task.blocked_by.set(blocked_by)
+        if blocked_by and has_transition_perm(task.block, self):
```
longhotsummer, line 432: We're only checking the perm *after* the task has been blocked. Instead, check the perm first and then pass the user and tasks to block on into task.block(), like we do with task.close(), then that method does the business logic.

indigo_app/urls.py
goose-life: I used a verb to follow the other patterns;  what do you think of `change-blocking-tasks`? that would also correspond with the view name

indigo_api/models/tasks.py
```diff
@@ -280,6 +287,26 @@ def close(self, user, **kwargs):
...
+    def block(self, user, **kwargs):
+        self.assigned_to = None
```
longhotsummer, line 298: Use `self.assign_to(None)`?

goose-life: Nice

indigo_app/urls.py
longhotsummer: It's a mixed bag. If we're doing REST (which we vaguely are, eg. projects), then the endpoints should actually be nouns and the HTTP verb determines the action. This doesn't work well for assign, unassign, etc. But it does kinda work for blocking tasks -- you're POSTing to the blocking-tasks collection.

If you're strictly using RPC-style naming, like a method name, then it would be a verb as you say.

indigo_app/views/tasks.py
goose-life: Is this a performance thing? because tasks have unique ids across places, right? Happy to do it of course, just curious

indigo_app/views/tasks.py
goose-life: If I'm understanding you right, you're suggesting that the block transition method on task set the blocked_by tasks on our task in the process of blocking it (and unblock would clear them). 

We also need to make provision for the blocked_by list just being updated without there being a change in state; that can only happen in the view unless I'm mistaken.

So `task.blocked_by.set(blocked_by)` would only be called in the case of there being blocked_by tasks but the transition permission being False (because you can't move from blocked into blocked).

indigo_app/views/tasks.py
longhotsummer: It's about being defensive about what you accept and validating the inputs. The user should only be able to specify task ids of tasks in the same place as the current task.

indigo_app/views/tasks.py
longhotsummer: Hmm, good point - maybe get around that by allowing blocked -> blocked as a state transition? Otherwise, set the blocked tasks outside of task.blocked(). Otherwise it's duplicate code.

indigo_api/models/tasks.py
```diff
@@ -280,6 +287,30 @@ def close(self, user, **kwargs):
...
+    def may_block(self, view):
+        return view.request.user.is_authenticated and \
+               view.request.user.editor.has_country_permission(view.country) and \
+               view.request.user.has_perm('indigo_api.block_task')
+
+    @transition(field=state, source=['open'], target='blocked', permission=may_block)
+    def block(self, user, **kwargs):
+        if self.assigned_to:
+            self.assign_to(None, user)
+        blocked_by = kwargs['blocked_by'] or None
```
longhotsummer, line 300: Perhaps make this an explicit parameter with a default of None -- python will do the hard work for you.

indigo_app/templates/indigo_api/task_form.html
```diff
@@ -118,14 +118,26 @@
...
+          {% if task.state == 'open' or task.state == 'pending_review' or task.state == 'blocked' %}
             <form method="POST">
             {% csrf_token %}
               {% if cancel_task_permission %}
                 <button class="btn btn-danger float-left" type="submit" data-confirm="Are you sure you want to cancel this task?" formaction="{% url 'cancel_task' place=place.place_code pk=task.pk %}">Cancel task</button>
               {% endif %}
             </form>
           {% endif %}
+          {% if task.state == 'open' and block_task_permission %}
+            <form method="POST">
```
longhotsummer, line 130: Rather than duplicating the form, just wrap them all in the same form, since the `formaction` attribute does the necessary.

indigo_app/views/tasks.py
goose-life: I don't love a blocked-to-blocked transition, and we'd have to then special-case the actions on that so that we don't end up with multiple 'user blocked task' actions in a row if they're just updating the list, so I've changed it to set the blocking tasks outside of the block and unblock methods

longhotsummer: Just sanity check the if tree... what if the task is blocked by the user doesn't have necessary perms to make this change?

indigo_app/templates/indigo_api/task_form.html
```diff
@@ -118,14 +118,17 @@
...
-            <form method="POST">
-            {% csrf_token %}
-              {% if cancel_task_permission %}
-                <button class="btn btn-danger float-left" type="submit" data-confirm="Are you sure you want to cancel this task?" formaction="{% url 'cancel_task' place=place.place_code pk=task.pk %}">Cancel task</button>
-              {% endif %}
-            </form>
-          {% endif %}
+          <form method="POST">
+          {% csrf_token %}
+            {% if task.state == 'open' or task.state == 'pending_review' or task.state == 'blocked' and cancel_task_permission %}
```
longhotsummer, line 123: This lands up being `if A or B or C and D` == `if A or B or (C and D)` because AND his higher precedence than OR. (AND is like multiply, OR is like add).

As far as I know, you can't use brackets in expressions in django templates. So you may need to move this into the view.

indigo_app/views/workflows.py
```diff
@@ -57,7 +57,7 @@ def get(self, request, *args, **kwargs):
 
         # initial state
         if not params.get('state'):
-            params.setlist('state', ['open', 'assigned', 'pending_review', 'done', 'cancelled'])
+            params.setlist('state', ['open', 'assigned', 'pending_review', 'done', 'cancelled', 'blocked'])
```
longhotsummer, line 60: There's an opportunity here (and maybe elsewhere) to do something like `... list(Task.STATES) + ['pending_review']` so that there's less typing of magic strings.

indigo_app/views/tasks.py
goose-life: Ah. I thought the permission_required as well as disabling the button if the user doesn't have permissions would cover us, but I see the country permission isn't checked (or it's not obvious if it is, maybe I've missed it now).

So because indigo_api.change_task is listed under permission_required, is the only other thing I need to check for the country permission? or should I explicitly check for indigo_api.change_task at the point where I'm setting the blocked_by list?

I'm not super clear on how permission_required works, but I see there's a has_permission on PlaceViewBase which checks for country permission. So should I just say `elif blocked_by and self.has_permission()`, and either raise a permission denied or go ahead and set the blocked_by list at that point?

Once I'm clear on this I can figure out how not to repeat the `set` line three times :grimacing:

longhotsummer: `permission_required` will be checked right at the start of the request, but assume that you don't know how `has_transition_perm` works, so you can't assume that because we've got this far the user can make this change (since has_transition_perm checks both validity and authorization, right?).

Change the `if` tree to be like this, which I think will make it clearer:

```
if blocked_by:
  # user is trying to block this task, is that transition allowed?
  if has_transition_perm(...):
    # can proceed
    ...
  else:
    raise permission denied
else:
  # user is trying to unblock, can that transition proceed?
  if has_transition_perm(..):
    # cool
    ...
  else:
    raise permission denied
```

indigo_app/views/tasks.py
```diff
@@ -327,13 +336,25 @@ def post(self, request, *args, **kwargs):
...
+            self.update_blocked_tasks(task, user)
+
         return redirect(self.get_redirect_url())
 
     def get_redirect_url(self):
         if self.request.GET.get('next'):
             return self.request.GET.get('next')
         return reverse('task_detail', kwargs={'place': self.kwargs['place'], 'pk': self.kwargs['pk']})
 
+    def update_blocked_tasks(self, task, user):
```
longhotsummer, line 349: This is a good candidate to move onto the task object, since it doesn't really have anything to do with the view. It could either be done based on the signals that django-fsm fires when transitions happen, or the `cancel` and `close` methods on the task could call it explicitly (which is probably the best).

indigo_app/views/tasks.py
```diff
@@ -327,13 +336,25 @@ def post(self, request, *args, **kwargs):
...
+        previously_blocked_tasks = list(task.blocking.all())
+        # this task is no longer blocking other tasks
+        task.blocking.clear()
+        for blocked_task in previously_blocked_tasks:
+            if not blocked_task.blocked_by.all() and has_transition_perm(blocked_task.unblock, self):
```
longhotsummer, line 354: use `.exist()` rather than loading all the data from the db and not using it.

should `self` at the end be `user`?

indigo_app/views/tasks.py
longhotsummer: This improves the abstraction: the view is responsible for translating the user's browser request into some action: eg. blocking or unblocking, etc. The task object itself is responsible for managing the impact of that action, not the view. The view should focus on the request, validating permissions, validating input and parameters, and then delegate to the object (or something else) to do the actual business logic. It makes testing and code re-use much easier.

indigo_app/views/tasks.py
goose-life: I was thinking of refactoring the may_{verb} code, but ya they all use the view, not the user

indigo_app/views/tasks.py
goose-life: I was blocked by only having access to the view, not the user, but since refactoring that code this is now possible. I did a little hoop jumping to get the messages to also say which tasks have been updated and unblocked by closing a task, le tmeknow what you think

indigo_api/models/tasks.py
goose-life: this has now been changed

indigo_app/views/tasks.py
goose-life: I've updated this to exclude closed tasks; we might need to filter it further in future so that not too many tasks are loaded.

goose-life:
image

image

indigo_api/models/tasks.py
```diff
@@ -181,13 +187,13 @@ def decorate_potential_assignees(cls, tasks, country):
...
     @classmethod
-    def decorate_permissions(cls, tasks, view):
+    def decorate_permissions(cls, tasks, user):
         for task in tasks:
-            task.change_task_permission = view.request.user.has_perm('indigo_api.change_task')
-            task.submit_task_permission = has_transition_perm(task.submit, view)
-            task.reopen_task_permission = has_transition_perm(task.reopen, view)
-            task.unsubmit_task_permission = has_transition_perm(task.unsubmit, view)
-            task.close_task_permission = has_transition_perm(task.close, view)
+            task.change_task_permission = user.has_perm('indigo_api.change_task')
```
longhotsummer, line 192: passing in the user and not the view is much nicer!

longhotsummer (approved):

PR #1243, opened 2021-02-08
Fix block tasks

goose-life:
Selecting task from dropdown works
image

Searching and hitting Enter works
image

Searching and then clicking now also works
image
image

### DIFF
--- a/indigo_app/templates/indigo_api/_task_detail.html
+++ b/indigo_app/templates/indigo_api/_task_detail.html
@@ -265,7 +265,9 @@
 
           <select class="selectpicker" multiple name="blocked_by"
                   data-dropdown-align-right="true" data-style="btn-link pl-0 text-dark"
-                  {% if not perms.indigo_api.change_task %}disabled{% endif %}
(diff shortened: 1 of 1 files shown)


indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -260,12 +260,14 @@ <h6>Labels</h6>
         {% endwith %}
 
         <form action="{% url 'task_blocked_by' place=place.place_code pk=task.pk %}" method="POST"
-              id="task-blocked_by-form" onchange="this.submit();">
+              id="task-blocked_by-form">
           {% csrf_token %}
 
-          <select class="selectpicker" multiple name="blocked_by"
+          <select class="selectpicker" multiple name="blocked_by" onchange="$(this.form).trigger('submit')"
```
longhotsummer, line 266: Any reason to use `trigger('submit')` vs calling `this.form.submit()`?

goose-life: no good reason; I was just copying what we had before :bow:

longhotsummer (approved):

PR #1245, opened 2021-02-11
don't automatically unblock a task when its blocked_by list is empty

goose-life:
User feedback indicates that it's better (sometimes) for a task to remain blocked because the work should ideally be signed off before the next task is done.

It'll be fairly tricky and I don't think failsafe to transfer the blockedness of a task from one (now closed) blocking task to a (created separately) signoff task, so for now let's err on the side of caution: The blocked task is still blocked, but not by the blocking task.

The new 'partially unblocked' actstream verb should also help to give users more insight into what happened.

image

After cancelling the two blocking tasks:
image

image

### DIFF
--- a/indigo_api/models/tasks.py
+++ b/indigo_api/models/tasks.py
@@ -317,12 +317,8 @@ class Task(models.Model):
             # this task is no longer blocking other tasks
             task.blocking.clear()
             for blocked_task in blocked_tasks:
-                if has_transition_perm(blocked_task.unblock, user):
-                    # the other task no longer has blocking tasks; unblock it
-                    blocked_task.unblock(user)
-                else:
-                    action.send(user, verb='updated', action_object=blocked_task,
-                                place_code=blocked_task.place.place_code)
+                action.send(user, verb='partially unblocked', action_object=blocked_task,
+                            place_code=blocked_task.place.place_code)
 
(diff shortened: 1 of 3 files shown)


indigo_api/models/tasks.py
```diff
@@ -317,12 +317,8 @@ def update_blocked_tasks(self, task, user):
             # this task is no longer blocking other tasks
             task.blocking.clear()
             for blocked_task in blocked_tasks:
-                if has_transition_perm(blocked_task.unblock, user):
-                    # the other task no longer has blocking tasks; unblock it
-                    blocked_task.unblock(user)
-                else:
-                    action.send(user, verb='updated', action_object=blocked_task,
-                                place_code=blocked_task.place.place_code)
+                action.send(user, verb='partially unblocked', action_object=blocked_task,
```
longhotsummer, line 320: I don't mind what the verb used internally is, but when we display it, maybe phrasing like "resolved a task that was blocking this one" would be clearer? Also, include the `task` as the action's target, so we have that backward reference.

longhotsummer: OR `... resolved {task} that was blocking this task`.

goose-life:
image

longhotsummer (approved):

PR #1247, opened 2021-02-11
Show all extra properties on bulk import 

goose-life:


goose-life:
Added some extra (unused) work properties to ZW for testing:
image

longhotsummer (approved):
I think the columns may be getting out of hand, but let's see how this goes :)

PR #1254, opened 2021-02-26
Include 'amends_on_date' and other '_on_date's on bulk import

### LINKED ISSUE #1253: Bulk importer: 'amends_on_date' not behaving as expected
See e.g. https://docs.google.com/spreadsheets/d/1ej8B5kITwzZu6m6SdvjTEtYuHMakcDrdsbe6bNYLPuA/edit#gid=1433517502

For the first amendment, the amends_on_date is given as July. This should supersede the commencement date (January) of the amending work as the date on which the amendment came into force, but this hasn't happened on import.

goose-life:
These are exported but have been ignored on import so far (other than for active commencements).

`repealed_on_date` and `repeals_on_date` would still be ignored on import after this PR; they should eventually be pulled in on import too.

closes #1253

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -55,6 +55,7 @@ class RowValidationFormBase(forms.Form):
     commences = forms.CharField(required=False)
     commences_on_date = forms.DateField(required=False, error_messages={'invalid': 'Date format should be yyyy-mm-dd.'})
     amends = forms.CharField(required=False)
+    amends_on_date = forms.DateField(required=False, error_messages={'invalid': 'Date format should be yyyy-mm-dd.'})
     amended_by = forms.CharField(required=False)
     repealed_by = forms.CharField(required=False)
     repeals = forms.CharField(required=False)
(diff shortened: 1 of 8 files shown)


indigo/bulk_creator.py
```diff
@@ -437,7 +441,7 @@ def check_preview_duplicates(self):
     def link_commencement_passive(self, row):
         # if the work has commencement details, try linking it
         # make a task if a `commenced_by` FRBR URI is given but not found
-        date = row.commencement_date
+        date = row.commenced_on_date or row.commencement_date
```
longhotsummer, line 444: What's the difference between `commenced_on_date` and `commencement_date`?

indigo/bulk_creator.py
```diff
@@ -52,10 +52,13 @@ class RowValidationFormBase(forms.Form):
     commencement_date = forms.DateField(required=False, error_messages={'invalid': 'Date format should be yyyy-mm-dd.'})
```
longhotsummer, line 52: Maybe it's time we documented the spreadsheet columns on docs.laws.africa, link to that here, and also add some brief comments here. Without context, there are not easy to understand.

goose-life: done

longhotsummer (approved):

PR #1260, opened 2021-03-18
Commencement note

goose-life:
will close https://github.com/laws-africa/indigo-kenyalaw/issues/166

To do:
- [x] Don't lose commencement note when reloading form
- [x] Only show 'Commencement note' section of form if 'Commencement date is unknown' is checked, and other clever things (Cancelled)
- [x] Remove warning about commencement date being discarded
- [x] Include commencement note on auto-generated header
- [x] Add commencement note to each commencement object on the Commencements tab of a work
- [x] Include note on each commencement in the commencements table


goose-life:
https://user-images.githubusercontent.com/32566441/111671423-99c42900-8821-11eb-87da-5213ff0cc7b5.mp4

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -164,6 +164,7 @@ <h5 class="card-header">Classification</h5>
 
       </div>
     </div>
+    {% block additional-information %}{% endblock %}
```
longhotsummer, line 167: This isn't necessary, because other templates can simply do this:

```
{% block work-form-details %}
  {{ block.super }}
  extra custom stuff
{% endblock %}
```

That's the reason for work-form-details being a block.

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -274,7 +275,18 @@ <h5 class="card-header">Commencement and assent</h5>
...
-                <p class="form-text text-muted">Check this box when a (usually old) work has come into force, but the exact commencement date is not known.</p>
+                <p class="form-text text-muted">Check this box when a work has come into force, but the exact commencement date is unknown or open to interpretation.</p>
+
+                <div id="commencement_note">
+                  <div class="form-row">
+                    <div class="form-group">
+                      <label for="{{ form.commencement_note.id_for_label }}">Commencement note</label>
+                      <input type="text" class="form-control" id="{{ form.commencement_note.id_for_label }}" name="{{ form.commencement_note.html_name }}" placeholder="See section 1." value="{{ form.commencement_note.value|default:'' }}">
+                      <p class="form-text text-muted">When the commencement date is open to interpretation, give a reference to a provision of this or a commencing work for added context.</p>
+                      <div class="alert alert-warning">If you add a commencement note, the commencement date (if there is one) will be deleted when you save.</div>
```
longhotsummer, line 286: I don't see the logic that does this? Also, I don't think we should do this -- if the user wants to have a date and a note, that seems reasonable to me.

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -274,7 +275,18 @@ <h5 class="card-header">Commencement and assent</h5>
                   <input type="checkbox" class="form-check-input" id="commencement_date_unknown">
                   <label for="commencement_date_unknown">Commencement date is unknown</label>
                 </div>
-                <p class="form-text text-muted">Check this box when a (usually old) work has come into force, but the exact commencement date is not known.</p>
+                <p class="form-text text-muted">Check this box when a work has come into force, but the exact commencement date is unknown or open to interpretation.</p>
+
+                <div id="commencement_note">
+                  <div class="form-row">
```
longhotsummer, line 281: The `from-row` is only necessary if it has a `col-*` as an immediate child. (I know further up it appears erroneously)

longhotsummer:
I think we should always allow a commencement note, rather than just when it's unclear. The user may want to add a date and indicate some caveat. It's more flexible if we're less restrictive, and being restrictive doesn't (AFAICT) make the implementation easier.

longhotsummer:
In which case, we could swap the two columns in the row it's in, so that it always appears.

goose-life:
image

goose-life:
image

indigo_app/static/lib/indigo-web/scss/_coverpage.scss
```diff
@@ -55,6 +55,7 @@
       font-style: italic;
     }
 
+    .commencement-note,
```
longhotsummer, line 58: this will need to happen in indigo-web too, otherwise the LIIs etc. won't pick it up.

indigo_app/templates/indigo_api/work_commencements.html
```diff
@@ -100,6 +101,12 @@ <h6>
                 </div>
               </div>
 
+              <div class="form-row form-group">
+                <div class="col-md-6 col-lg-4">
+                  <input type="text" class="form-control mr-2" placeholder="See section 1 (of the commencing work)." name="note" value="{{ commencement.note|default:'' }}" autocomplete="off">
```
longhotsummer, line 106: I think the placeholder is confusing. Perhaps use "e.g. See section 1 ..." ?

goose-life:
Updated commencements table
image

indigo_api/templates/indigo_api/akn/coverpage_act.html
```diff
@@ -88,93 +88,106 @@ <h2>{{ subt }}</h2>
...
+                              {% if numbered_title %}
+                                {% blocktrans %}by <a href="{{ resolver_uri }}">{{ numbered_title }}</a>{% endblocktrans %}.
+                              {% else %}
+                                {% blocktrans %}by <a href="{{ resolver_uri }}">{{ title }}</a>{% endblocktrans %}.
+                              {% endif %}
+                            {% endwith %}
+                          {% endif %}
+                        {% endwith %}
+                        {% if commencement.note %}
+                          {% blocktrans %}<br><i>Note:{% endblocktrans %}
```
longhotsummer, line 172:
```suggestion
                          <br><i>{% blocktrans %}Note:{% endblocktrans %}
```

longhotsummer (dismissed):

longhotsummer (approved):

PR #1275, opened 2021-05-24
Update perms

### LINKED ISSUE #1274: Clean up permission enforcement
Standardise permissions to make reasoning and enforcement easier

* [x] remove option to make authentication optional, it is too complex
* [x] for app api (not content api), use normal django model permissions where possible, and still enforce document-based view and change permissions
* [x] enforce view permissions for viewing documents and works
* [x] make badges awarded to new users configurable
* [x] add view permissions to ContributorBadge

longhotsummer:
Closes #1274 

* add and enforce view permissions
* enforce permissions on some views that didn't have them
* use model permissions in APIs and views


indigo_api/tests/test_analysis_api.py
```diff
@@ -70,3 +74,62 @@ def test_link_terms(self):
...
+        self.assertEqual(response.status_code, 403)
+
+        # user has both view and change perms, but not publication permissions
+        user.user_permissions.add(Permission.objects.get(
+            codename='change_document',
+            content_type=ContentType.objects.get_for_model(Document)
+        ))
+
+        response = self.client.post('/api/documents/1/analysis/link-terms', data)
+        self.assertEqual(response.status_code, 403)
```
goose-life, line 120: shouldn't a user be able to link terms without publication permissions?

goose-life (approved):

indigo_api/tests/test_analysis_api.py
longhotsummer: In this case they're trying to manipulate a published (non-draft) document, so any modification requires publication perms. Yes, you could argue that they should still be able to run it and not save, but I haven't made it that granular yet.

goose-life: makes sense to me, thanks!

PR #1283, opened 2021-06-18
Sub section commencements

### LINKED ISSUE #1252: as an editor I want to record commencements below the section level
@longhotsummer I don't think we actually have an open issue for this?

# The problem

While we support multiple commencements, we only support them down to the 'section' level, but often subsections, paragraphs, or subparagraphs commence on different dates (or remain uncommenced).

# The desired user experience

1. On the Commencements page of a work, when selecting commenced provisions, instead of being shown a list of sections from which to choose, have a tree view instead, where one can e.g. select all provisions, and then drill down into e.g. Chapter 2, section 4, subsection (2), paragraph (dA), subparagraph (i), subsubparagraph (C), and unselect it.
2. When only some provisions are uncommenced, only show their ancestors, not their siblings; only allow the uncommenced provision itself to be selected (or remain unselected).

# The necessary changes

1. Nest the ToC down to all hierarchical elements, plus `item`. The ToC itself is already hierarchical, which shouldn't change.
2. Change the UI to reflect the ToC tree, and automatically select all children of a selected element by default.
3. Instead of showing the flattened 'commenced provisions' above, show a decorated ToC instead, where each element in the ToC knows whether it's commenced or not.
3. Record every commenced provision (probably no change needed here), i.e. if section 1 and its children are selected also add sec_1__para_a etc to 'commenced provisions'. How 'uncommenced provisions' are worked out should therefore probably not change either.
4. Potentially change the WorkExpression serializer to produce a more descriptive ToC summary, because an API user will get not a lot of value from long lists of eIds, or provide the summary in addition to the serialized ToC?

goose-life:
will close #1252 

To do:

- [x] ~Migration~ Management command
- [x] Tests
- [x] Adjust work.all_uncommenced_provisions()
- [ ] last tweaks to commencements view

The provisions and other commencement-related stuff gets worked out here:
https://github.com/laws-africa/indigo/blob/1928ec57a8b6a2a7583f0370d64587f71346e8e7/indigo_app/views/works.py#L303

The current view for editing commencements is here:
https://github.com/laws-africa/indigo/blob/36abecbfb84ae71ae9a12b3268750883dd33f61e/indigo_app/templates/indigo_api/work_commencements.html#L134

A Commencement object gets updated here (from the post in the form):
https://github.com/laws-africa/indigo/blob/1928ec57a8b6a2a7583f0370d64587f71346e8e7/indigo_app/views/works.py#L325

Hope this makes sense, shout if you need dummy data or anything else!

## Front end implemenations:
- Implemented Tree checkbox view for provisions in a commencement

Demo:
https://www.loom.com/share/3dbee34e8253484d95e49e80b74f7e44


goose-life:
I'll update the tests once we're sure that this is how tocs should look :)

indigo_app/templates/indigo_api/template_parts/provisions_tree_view.html
```diff
@@ -0,0 +1,18 @@
+{% for prov in provisions %}
+  <li class="list-group-item">
+    <div class="d-flex align-items-center">
+      <input id="{{ prov.id }}"
```
musangowope, line 4: need to figure out how to create unique ids otherwise we have duplicate ids

indigo_app/templates/indigo_api/work_commencements.html
```diff
@@ -130,6 +130,13 @@ <h6>
               <div class="form-group provisions-commenced {% if commencement.all_provisions %}d-none{% endif %}">
                 <label>Provisions commenced:</label>
                 <small class="form-text text-muted">Use SHIFT and Ctrl (or ⌘) to select multiple items.</small>
+
+                <ul class="list-group" data-id="provisions">
+                  {% include "indigo_api/template_parts/provisions_tree_view.html" with provisions=commencement.possible_provisions %}
```
musangowope, line 135: what do you think of this implementation for nested lists. do you have a better way?

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -7,6 +7,31 @@
   /**
    * Handle the work commencements display.
    */
+
+  //TODO: Adapt Logic to backbone
```
musangowope, line 11: @longhotsummer can we chat about how to adapt this to backbone standards?

indigo_app/templates/indigo_api/template_parts/provisions_tree_view.html
goose-life: can you prepend the parent id? that should ensure uniqueness (I think)

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -7,18 +7,24 @@
...
+      'click .commencement-form .expand-collapse-button' : 'onExpandCollapseClick',
+      'change .commencement-form input[name="select-all"]' : "onSelectAll",
     },
 
     initialize: function() {
       this.model = new Indigo.Work(Indigo.Preloads.work);
-      this.$('.commencement-form').on('show.bs.collapse', _.bind(this.formShow, this));
-      this.$('.commencement-form').on('hide.bs.collapse', _.bind(this.formHide, this));
+
+      // this.$('.commencement-form').on('show.bs.collapse', _.bind(this.formShow, this));
```
musangowope, line 26: @longhotsummer can we chat about these functions tomorrow? are they needed?
it causing this issue
https://africanliiteam.slack.com/files/UF69ATL6S/F025J79LDAT/work-commencement.mp4?origin_team=TB3H1VDGD&origin_channel=CK3QE8BU1

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -68,6 +92,35 @@
         // we can't iterate over this in-place, so do this instead
         while (form.elements.provisions.selectedOptions.length > 0) form.elements.provisions.selectedOptions[0].selected = false;
       }
+    },
+
+    handleCheckboxesChange(e) {
+      function toggleSelectionByCheckSelected(currentCheckbox) {
+        const parentNodeListItem = currentCheckbox.closest("li");
```
longhotsummer, line 99: Once we've got the LI that this checkbox belongs to, can't we just find all checkboxes beneath it and toggle their state? Something like:

```js
for (const checkbox of currentCheckbox.closest('li').querySelectorAll('ul input[type=checkbox]')) {
  checkbox.checked = currentCheckbox.checked;
}
```

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -7,18 +7,24 @@
   /**
    * Handle the work commencements display.
    */
+
```
longhotsummer, line 10: The class comment should be adjacent to the class it documents.

indigo_app/static/javascript/indigo/views/work_commencements.js
musangowope: done

indigo_app/static/javascript/indigo/views/work_commencements.js
musangowope: done

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -60,6 +64,24 @@
       $(e.target).closest('.card').find('.provisions-commenced').toggleClass('d-none', e.target.checked);
     },
 
+    onExpandCollapseClick: function(e) {
+      e.stopPropagation();
+      const button = e.currentTarget;
+      const parent = button.closest('li');
+      if(button.classList.contains('expanded')) {
```
longhotsummer, line 71: `button.classList.toggle('expanded')`

Please also put a space in `if(` (and below, as well).

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -60,6 +64,24 @@
...
+        button.classList.remove('expanded');
+      } else {
+        button.classList.add('expanded');
+      }
+
+      if(parent) {
```
longhotsummer, line 77: Surely the HTML guarantees that `parent` is not null?

You can also rely on jQuery to do all this in one go (what it's good at - it handles zero, one and many matches gracefully): 

`$(parent).find('.collapse').collapse('toggle');`

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -68,6 +90,22 @@
         // we can't iterate over this in-place, so do this instead
         while (form.elements.provisions.selectedOptions.length > 0) form.elements.provisions.selectedOptions[0].selected = false;
       }
+    },
+
+    handleCheckboxesChange(e) {
+      const parentElement = e.target.closest("li");
+      const nestedListElement = parentElement.querySelector('ul');
+      if(nestedListElement) {
+        for(const checkBox of nestedListElement.querySelectorAll("input[name='provisions']")) {
```
longhotsummer, line 99: You can eliminate the if and nestedListElement completely and just iterate directly over the inputs, eliminating two variables which are used only once AND the if statement:

```
for (const checkBox of e.target.closest('li').querySelectorAll("ul input[name='provisions']")) {
```

Also space in `for(` -- for isn't a function call.

indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -68,6 +90,22 @@
...
+          checkBox.checked = e.target.checked;
+        }
+      }
+    },
+
+    onSelectAll: function (e) {
+      for(const checkbox of document.querySelectorAll(".commencement-form input[name='provisions']")) {
```
longhotsummer, line 106:
```suggestion
      for (const checkbox of document.querySelectorAll(".commencement-form input[name='provisions']")) {
```

indigo_app/static/javascript/indigo/views/work_commencements.js
musangowope: done

indigo_app/static/javascript/indigo/views/work_commencements.js
musangowope: `.find` is selects all the nodes in the parent. lets rather use
`$(parent).find('.collapse').first().collapse('toggle');`

indigo_app/static/javascript/indigo/views/work_commencements.js
musangowope: done

goose-life:
@musangowope these are now working but unbeautiful:

Commenced provisions per commencement:
image


Uncommenced provisions:
image


To do:

- [ ] Migration
- [ ] Tests

indigo/tests/test_beautiful_provisions.py
```diff
@@ -1,94 +1,444 @@
...
+            'chp_1__part_B', 'sec_4',
+        ]
+        self.beautifier.commenced = True
+        self.assertEqual('Chapter 1 (in part), Part A (section 1–3); Part B (in part); section 4', self.run_nested(provision_ids))
+        self.beautifier.commenced = False
+        self.assertEqual('Chapter 1 (in part), Part A (section 1–3); Part B (in part); section 4', self.run_nested(provision_ids))
+
+        provision_ids = ['chp_1', 'sec_2', 'chp_1__part_B']
+        self.beautifier.commenced = True
+        self.assertEqual('Chapter 1 (in part), section 2; Part B (in part)', self.run_nested(provision_ids))
```
longhotsummer, line 240: Is `Chapter 1 (in part), section 2` the equivalent of `Chapter 1 (in part); Chapter 1, section 2` ? I'm not quite following when the Chapter is included because it's explicitly commenced, and when it's included because it's a parent of the section (ie. to scope the section).

indigo/tests/test_beautiful_provisions.py
```diff
@@ -1,94 +1,444 @@
...
+        self.beautifier.commenced = False
+        self.assertEqual('Chapter 1 (in part), Part A (section 1–3)', self.run_nested(provision_ids))
+
+        provision_ids = [
+            'sec_2', 'sec_3',
+            'chp_1__part_B', 'sec_4', 'sec_5',
+            'chp_2', 'sec_6', 'sec_7'
+        ]
+        self.beautifier.commenced = True
+        self.assertEqual('Chapter 1, section 2–3, Part B (section 4–5); Chapter 2 (section 6–7)', self.run_nested(provision_ids))
```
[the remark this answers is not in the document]
goose-life: Ah, good point. Sections 2 and 3 are in Part A, which isn't mentioned at all.
-->
`'Chapter 1, Part A, section 2–3, Part B (section 4–5); Chapter 2 (section 6–7)'`?

indigo/tests/test_beautiful_provisions.py
goose-life: I'm starting to think it should be the second option. I had previously thought differently but that was before giving `Chapter 1 (in part)` if it was explicitly marked as commenced.

goose-life:
@longhotsummer test outputs updated again, what do you think? I initially thought something like this would be too much repetition but should we just be explicit about everything?

longhotsummer:
I think this is a great starting point. I am happy to deploy and use this, and see what it looks like in the real world.

Later, if we want to reduce the repetition based on our knowledge of the basic unit (which doesn't need scoping to be unique), we can do so in the cases where the parent elements are there purely for context (and not because they have commencement details at that point).

goose-life:
Here are some screenshots:
image
image
image
Same set of provisions, later point in time:
image

indigo_api/models/documents.py
```diff
@@ -183,21 +183,38 @@ def add_ids(toc):
     def commenceable_provisions(self):
```
longhotsummer, line 183: I think an outcome of our discussion is that all these methods are removed in favour of being run on the work object, since logically they belong there and not here.

indigo/analysis/toc/base.py
```diff
@@ -245,10 +257,7 @@ def commenceable_items(self, toc):
         """
         def process(item):
             if item.component == 'main':
-                if item.children:
-                    for kid in item.children:
-                        process(kid)
-                elif item.type not in self.non_commenceable_types and item.num:
+                if item.type not in self.non_commenceable_types and item.num:
```
longhotsummer, line 260: Just noting that this no longer recurses into children, so it's returning a tree structure rather than a list.

indigo_api/models/works.py
```diff
@@ -269,6 +269,26 @@ def all_commenceable_provisions(self, date=None):
 
         return provisions
 
+    def uncommenced_provision_ids(self, date=None):
```
longhotsummer, line 272: I'm trying to work out why we can't use something like `[p.id for p in self.uncommenced_provisions(date)]`. It seems odd that we have `uncommenced_provisions()` and `commenced_provisions()` and yet the id-based versions don't make use of them, and that `has_uncommenced_provisions()` isn't just `bool(self.uncommenced_provisions())` or similar.

Basically, we have four functions that do very similar work in very slightly different ways.

indigo_content_api/tests/v2/test_content_api.py
```diff
@@ -221,10 +221,27 @@ def test_published_toc_sections_with_headings(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.accepted_media_type, 'application/json')
         self.assertEqual(response.data['toc'], [
-            {'id': 'sec_1', 'type': 'section', 'num': '1.',
-             'component': 'main', 'subcomponent': 'section/1',
-             'url': 'http://' + self.api_host + self.api_path + '/akn/za/act/2010/1/eng/!main/section/1',
-             'heading': 'Foo', 'title': '1. Foo'}])
+            {
```
longhotsummer, line 224: This implies that the TOC for all existing works will now always go below the section level. We need to work out what the impact of this for our existing websites.

indigo_app/views/works.py
```diff
@@ -303,31 +303,93 @@ def get_top_contributors(self):
...
             # possible options
-            commencement.possible_provisions = self.get_possible_provisions(commencement, commencements, provisions)
+            commencement.possible_provisions = self.get_possible_provisions(commencement, commencements)
 
         return context
 
-    def get_possible_provisions(self, commencement, commencements, provisions):
+    def add_info(self, p, commenced_provisions, upc, tpc):
+        for c in p.children:
+            upc, tpc = self.add_info(c, commenced_provisions, upc, tpc)
```
longhotsummer, line 338: Perhaps `n_commenced, n_uncommenced` are less opaque variable names.

indigo_app/views/works.py
```diff
@@ -303,31 +303,93 @@ def get_top_contributors(self):
...
+
+        p.commenced = p.id in commenced_provisions
+        p.uncommenced_descendants = (any(not c.commenced for c in p.children)) or any(c.uncommenced_descendants for c in p.children)
```
longhotsummer, line 341:
```suggestion
        p.uncommenced_descendants = any(not c.commenced for c in p.children) or any(c.uncommenced_descendants for c in p.children)
```

indigo_app/views/works.py
longhotsummer: Or, re-use the variable names from `add_commencement_info`

indigo_app/views/works.py
```diff
@@ -303,31 +303,93 @@ def get_top_contributors(self):
...
+        commenced_provisions = [p for c in commencements for p in c.provisions]
+        uncommenced_provisions_count = 0
+        total_provisions_count = 0
+
+        for p in provisions:
+            uncommenced_provisions_count, total_provisions_count = self.add_info(p, commenced_provisions, uncommenced_provisions_count, total_provisions_count)
+
+        return uncommenced_provisions_count, total_provisions_count
+
+    def get_commenced_provisions(self, commencement):
```
longhotsummer, line 359: This isn't actually getting all the commencements, it's getting all the commenceable provisions at this date, and marking those that are commenced by this commencement.

Since both this and `get_possible_provisions` actually operate on `all_commencable_provisions`, maybe call `all_commenceable_provisions` once and pass the provisions in. Then this (and the next fn) can be more accurately named `mark_commenced` and `mark_visible`.

indigo_app/views/works.py
```diff
@@ -303,31 +303,93 @@ def get_top_contributors(self):
...
-            commencement.provision_items = [provision_set.get(p, p) for p in commencement.provisions]
+            commencement.commenced_provisions = self.get_commenced_provisions(commencement)
             # possible options
-            commencement.possible_provisions = self.get_possible_provisions(commencement, commencements, provisions)
+            commencement.possible_provisions = self.get_possible_provisions(commencement, commencements)
 
         return context
 
-    def get_possible_provisions(self, commencement, commencements, provisions):
+    def add_info(self, p, commenced_provisions, upc, tpc):
```
longhotsummer, line 336: This seems really similar to `get_commenced_provisions`... can the action of setting attributes be merged?

longhotsummer:
@goose-life I've added some comments, very happy to walk through them.

Additionally, the recursion pattern you follow is used many times. It may make your life easier, and the code easier to read, if you defined a helper function (it can live with the TOC code since it's directly linked):

```python
def descend_toc_pre_order(items):
  # yields each TOC element and then its ancestors
  for item in items:
    yield item
    for descendant in descend_toc_pre_order(item.children):
      yield descendant
```

You can then change your recursive code to something like:

```python
collection = [p for p in descend_toc_pre_order(provisions) if p.thingy]
# OR
for provision in descend_toc_pre_order(provisions):
  if some_condition(provision):
    provision.foo = bar
```

And there's an equivalent `descend_toc_post_order(items)` of course.

indigo_api/templatetags/indigo.py
```diff
@@ -4,6 +4,203 @@
 register = template.Library()
 
 
+class Beautifier:
```
longhotsummer, line 7: I think a better home for this would be `toc/base.py`. Also maybe `CommencementsBeautifier` is a less generic name?

indigo_api/models/works.py
goose-life: Agreed. I'm thinking to nuke `uncommenced_provisions()` entirely, because all we ever need is a list of the ids.
The reason we can't use `[p.id for p in self.uncommenced_provisions(date)]` is `uncommenced_provisions()` would return a tree structure if we were to update it to work correctly with `all_commenceable_provisions()`, so we'd have to recurse into that.
For `has_uncommenced_provisions()`, I thought it might simplify things by stopping at the first uncommenced provision rather than going through the whole commencement tree multiple times. I'll have another look, maybe we should just `use all_uncommenced_provision_ids()` as a proxy instead.

longhotsummer: Cool. I suspect with a helper like `descend_toc_pre_order` a lot of this becomes much easier, and so having dedicated functions to do things like extract ids, isn't required.

goose-life:
@longhotsummer all comments taken in except the recursion helper, which I don't 100% understand

goose-life:
@musangowope this is 98% of the way there now, just missing the beauty on the commencements listing view

goose-life:
Commencements view:
image
image

Commencements table on coverpage:
image
image

indigo_app/views/works.py
```diff
@@ -349,7 +371,7 @@ def get_permission_required(self):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs['work'] = self.work
-        kwargs['provisions'] = self.work.all_commenceable_provisions()
+        kwargs['provisions'] = [p for p in descend_toc_pre_order(self.work.all_commenceable_provisions())]
```
longhotsummer, line 374:
```suggestion
        kwargs['provisions'] = list(descend_toc_pre_order(self.work.all_commenceable_provisions()))
```

indigo/analysis/toc/base.py
```diff
@@ -333,3 +369,179 @@ def as_dict(self):
...
+            # do this here for all provisions
+            p.num = p.num.strip('.')
+
+            # when self.commenced is True, assess_against is the list of commenced provision ids
+            # when self.commenced is False, assess_against is the list of uncommenced provision ids
+            p.commenced = self.commenced if p.id in assess_against else not self.commenced
+
+            # Do ALL descendants share the same commencement status as the current p?
+            # empty list passed to all() returns True
+            p.all_descendants_same = all(c.commenced == p.commenced for c in descend_toc_pre_order(p.children)) \
```
longhotsummer, line 391:
```suggestion
            p.all_descendants_same = all(c.commenced == p.commenced and (c.all_descendants_same or not c.children) for c in p.children)) \
```

indigo_api/templatetags/indigo.py
```diff
@@ -21,59 +25,27 @@ def work_resolver_url(context, work):
...
 
-    return make_beautiful(provisions, commenceable_provisions)
+    beautifier = CommencementsBeautifier(commenced=not uncommenced)
+    # decorate the ToC with useful information
+    provisions = beautifier.decorate_provisions(provisions, provision_ids)
+    return beautifier.make_beautiful(provisions)
 
 
 @register.simple_tag
 def commencements_relevant_at_date(document):
```
longhotsummer, line 46: I guess this doesn't need to be a template tag since we can call this directly on the document object. Maybe in the past we passed in a date...

longhotsummer: I raise this simply because I'm trying to understand the semantic methods that we have on the Document object, and `commencements_relevant_at_expression_date` is really pretty decent.

indigo_api/management/commands/update_commencements.py
```diff
@@ -0,0 +1,102 @@
...
+                for c in p.children:
+                    add_containers(c)
+
+                if p.container and not p.commenced and (all(c.commenced for c in p.children)):
+                    # this container is uncommenced but all its children are commenced; fix that
+                    commencement.provisions.append(p.id)
+                    # update decoration
+                    p.commenced = True
+
+            link = f"https://edit.laws.africa/works{commencement.commenced_work.frbr_uri}/commencements/#commencement-{commencement.pk}"
```
longhotsummer, line 63: Use `indigo.settings.INDIGO_URL` to build this URL.

longhotsummer:
Here's how vuetify shows a tree. Notice how the chevrons are (almost exactly) aligned with the text of the parent, and when there are no children there is no icon, but the same indentation.

image

docs/changelog.rst
```diff
@@ -3,13 +3,19 @@ Changelog
 =========
 
 14.1.0 (?)
```
goose-life, line 5: should this be a new major version?

longhotsummer: Definitely.

longhotsummer (dismissed):
🔥

longhotsummer (approved):

longhotsummer (approved):

PR #1288, opened 2021-06-28
Vue gutter and annotations

longhotsummer:
This is the start of using Vue in Indigo. It adds a new Gutter component which pulls annotation threads out of the body of the document and into a gutter alongside it, with dynamic positioning.

* webpack config for new main.js, alongside existing ES6 imports (these will, in time, be merged together)
* add a new ES6-based IndigoApp that is responsible for creating vue-based components
* add a new global event to allow the new IndigoApp to initialise itself at the right time
* styling for annotations in gutter

## Outstanding

* [x] insert the "new annotation" floating button into the gutter


longhotsummer:
@musangowope here's the starting point for Vue in Indigo. Please can you take a look at doing the final step for the "new annotation" floating button. This will help me check that everything for Vue is documented and in place.

PR #1290, opened 2021-07-08
Add missing compiled indigo-app.js

longhotsummer:


longhotsummer:
In future we can explore compiling this automatically, but for now manually is sufficient.

PR #1292, opened 2021-07-14
Styling for commencements list

musangowope:
add required styling for commencement list tree views
https://www.loom.com/share/23ea868be7b54f68ba0e21461f829694

### DIFF
--- a/indigo_app/static/stylesheets/_treeview.scss
+++ b/indigo_app/static/stylesheets/_treeview.scss
@@ -1,0 +1,19 @@
+.treelist-wrapper {
+  ul {
+    padding-left: 0;
+  }
+  li {
+    display: flex;
+  }
+
+  .spacer, input[type="checkbox"] {
+    width: 22px;
+  }
+  input[type="checkbox"]  {
+    margin-top: 3px;
+  }
+
+  .content {
+    flex-grow: 1;
+  }
+}
--- a/indigo_app/static/stylesheets/_treeview.scss
+++ b/indigo_app/static/stylesheets/_treeview.scss
@@ -4,13 +4,17 @@
   }
   li {
     display: flex;
+    align-items: flex-start;
   }
 
-  .spacer, input[type="checkbox"] {
+  .spacer {
     width: 22px;
   }
+
   input[type="checkbox"]  {
     margin-top: 3px;
+    width: 12px;
+    margin-right: 3px;
   }
 
   .content {
--- a/indigo_app/static/stylesheets/_treeview.scss
+++ b/indigo_app/static/stylesheets/_treeview.scss
@@ -7,14 +7,8 @@
     align-items: flex-start;
   }
 
-  .spacer {
-    width: 22px;
-  }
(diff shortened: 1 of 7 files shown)


goose-life:
the checkboxes look weirdly big on my end? 
image

also the vertical alignment of the checkboxes should be at the top here:
image

indigo_app/static/stylesheets/_treeview.scss
```diff
@@ -0,0 +1,23 @@
...
+    align-items: flex-start;
+  }
+
+  .spacer {
+    width: 22px;
+  }
+
+  input[type="checkbox"]  {
+    margin-top: 3px;
+    width: 12px;
```
longhotsummer, line 16: This isn't a good idea. Checkboxes are weird: the widgets differ between platforms so setting the actual widget size is asking for trouble (as you discovered with Firefox on linux).

Rather than setting the widget width, if you put it inside a wrapping div (eg `.marker` to match `.content`) then the div size can be set consistently and the div content is unstyled. That would also mean you don't need to differentiate between having content and having a spacer.

indigo_app/static/stylesheets/app.scss
```diff
@@ -39,6 +39,7 @@ $fa-font-path: "../lib/fontawesome-5.2/webfonts";
 @import 'utilities';
 @import 'works';
 @import 'provisions-list';
+@import "treeview";
```
longhotsummer, line 42: Why create a new set of styles, instead of adjusting the provisions-list styles, which you added just for this use case IIRC? Don't we now run the risk that `provisions-list` is never used?

indigo_app/templates/indigo_api/_provisions_tree_commenced.html
```diff
@@ -1,20 +1,23 @@
...
-      </span>
-
-      {% if prov.commenced and prov.all_descendants_same %}
-        (Commenced in full)
-      {% elif prov.children %}
-        <ul class="provisions-list">
-          {% include "indigo_api/_provisions_tree_commenced.html" with provisions=prov.children %}
-        </ul>
+      {% if prov.commenced %}
+        <input type="checkbox" disabled checked/>
```
longhotsummer, line 5: Self-closing tags aren't supported in HTML5.
```suggestion
        <input type="checkbox" disabled checked>
```

indigo_app/static/stylesheets/app.scss
longhotsummer: To answer my own question: because they're different things!

indigo_app/static/stylesheets/_treeview.scss
musangowope: refactored

musangowope:
updated screenshot:

Uncommenced list view
latest-treeview

Provisions list with long label text
provision-list-fixed

longhotsummer (approved):
LGTM

goose-life (approved):

PR #1293, opened 2021-07-15
Fix as dict

goose-life:


indigo/analysis/toc/base.py
```diff
@@ -368,17 +368,13 @@ def as_dict(self):
             'title': self.title,
             'children': [c.as_dict() for c in self.children],
             'basic_unit': self.basic_unit,
+            'num': self.num,
```
goose-life, line 371: should this be self.num or ''?

longhotsummer: no, this is fine, it must be None if not present not `''`

longhotsummer: I think add heading here too.

longhotsummer (approved):

longhotsummer (approved):

PR #1295, opened 2021-07-16
fix sentry error e.target.closest(...).querySelector(...) is null

musangowope:
- resolve sentry issue
https://sentry.io/organizations/lawsafrica/issues/2513849430/?project=5229057&referrer=slack
-For for provision list in a list item, there might not always be a nested ul element, so I added a check to guard against that

### DIFF
--- a/indigo_app/static/javascript/indigo/views/work_commencements.js
+++ b/indigo_app/static/javascript/indigo/views/work_commencements.js
@@ -91,7 +91,7 @@
     handleCheckboxesChange(e) {
       const ulElement = e.target.closest("li").querySelector('ul');
       //There may not always be a nested ul
-      if(ulElement) {
+      if (ulElement) {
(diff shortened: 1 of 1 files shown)


indigo_app/static/javascript/indigo/views/work_commencements.js
```diff
@@ -89,10 +89,12 @@
     },
 
     handleCheckboxesChange(e) {
-      for (const checkBox of e.target.closest("li")
-          .querySelector('ul')
-          .querySelectorAll("input[name='provisions']")) {
-        checkBox.checked = e.target.checked;
+      const ulElement = e.target.closest("li").querySelector('ul');
+      //There may not always be a nested ul
+      if(ulElement) {
```
longhotsummer, line 94:
```suggestion
      if (ulElement) {
```

longhotsummer (approved):

PR #1296, opened 2021-07-16
As at override

### LINKED ISSUE #1289: As a senior editor I want the option to override a place's as-at date per work (or sub-collection)
ref #1263 

User experience:
On the Work form, fill in the as-at date.

Back end:
Add 'as at date' to Work model, null=True, blank=True.

Front end:
Add 'as-at date' field to Work form, null if blank.
On the coverpage (and in the API?), use the work's as-at date, falling back to the place's as-at date.

In future: 
Introduce 'collections' which have as-at dates?

goose-life:
closes #1289 

Field added to end of Work form:
image

Displayed on overview:
image

Used on coverpage:
image


indigo_app/templates/indigo_api/work_form.html
```diff
@@ -338,6 +338,20 @@ <h5 class="card-header">Repeal</h5>
       </div>
     </div>
     {% endblock work-form-repeal %}
+    <div class="form-row">
+      <div class="form-group col-md-4">
+        <label for="{{ form.as_at_date.id_for_label }}">{{ form.as_at_date.label }}</label>
+        <input type="text" class="form-control" data-provide="datepicker" id="{{ form.as_at_date.id_for_label }}" placeholder="yyyy-mm-dd" pattern="\d{4}-\d\d-\d\d" name="{{ form.as_at_date.html_name }}" value="{{ form.as_at_date.value|default_if_none:''|stringformat:'s' }}">
+        <p class="form-text text-muted">Date up to which this work was last checked for updates.</p>
```
longhotsummer, line 345: I think this needs to refer to the fact that if it's blank, this will be computed.

`Date up to which this work was last checked for updates. If unset, uses the greater of {{ place.name }}'s as-at date, and this work's latest expression date.`

indigo_api/models/works.py
```diff
@@ -455,9 +456,13 @@ def versions(self):
         """
         return Version.objects.get_for_object(self).select_related('revision', 'revision__user')
 
-    def as_at_date(self):
+    def computed_as_at_date(self):
```
longhotsummer, line 459: I would do this the other way -- callers shouldn't have to care that this is computed, they call `as_at_date` as before (ie. don't change the object's API). Just use a field name that doesn't clash, eg. `as_at_date_override`? That means you need to change less of the code that makes use of this existing method.

indigo_app/templates/indigo_api/work_overview.html
```diff
@@ -19,6 +19,7 @@ <h3>
     {% if not work.commenced %} <span class="badge badge-info">not commenced</span>
     {% elif not work.commencement_date %} <span class="badge badge-info">commencement date unknown</span>
     {% endif %}
+    {% if work.as_at_date %}<span class="badge badge-info" title="{{ work.as_at_date|date:'Y-m-d' }}">as-at-overridden</span>{% endif %}
```
longhotsummer, line 22: I think this looks a bit clumsy.

How about putting it here:

image

```
<whatever the heading tag is>Up-to-date as at</...>

{{ calculated as at date }}
{% if work.as_at_date_override %}
<span class="text-muted" title="Set explicitly on this work"><i class="fa fa-info-circle"></i> Overidden</span>
{% endif %}

File: indigo_content_api/v2/serializers.py [hunk 4]
```diff
@@ -68,7 +68,8 @@ class PublishedDocumentSerializer(DocumentSerializer, PublishedDocUrlMixin):
     points_in_time = serializers.SerializerMethodField()
     publication_document = serializers.SerializerMethodField()
     taxonomies = serializers.SerializerMethodField()
-    as_at_date = serializers.DateField(source='work.as_at_date')
+    # TODO: double check this
+    as_at_date = serializers.DateField(source='work.computed_as_at_date')
```
longhotsummer, line 72: You don't need to worry about this if you don't change the existing `as_at_date` API on the work object.

goose-life:
updated screenshots:

form
image

overview
image

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -338,6 +338,21 @@ <h5 class="card-header">Repeal</h5>
       </div>
     </div>
     {% endblock work-form-repeal %}
+    <div class="form-row">
```
longhotsummer, line 341: Please put this inside a `.card` with an appropriate header, so that it matches the layout of the rest of the page.

An idea would be to put this together with the Assent date, since that's combined with commencement which is pretty complex. So Commencement would be on its own, and "Key dates" would have Assent an As-at date. Thoughts?

longhotsummer (changes requested):
Nice, less code has to change :)

Just a change to where the widget is placed in the form, pls.

goose-life:
form change:
image

indigo_app/templates/indigo_api/work_form.html
goose-life: yeah that makes sense! lmk what you think of the screenshot

longhotsummer (approved):

PR #1305, opened 2021-07-23
deepcopy cached toc before manipulating it

goose-life:


longhotsummer (approved):
I think this should solve all of the issues, given that we now know that `insert_provisions` changes children on elements. Do we still need to deepcopy in the commencements detail view?

longhotsummer (approved):

PR #1306, opened 2021-07-24
Make it possible to create vue components on demand

longhotsummer:
This means that if some code loads HTML dynamically, we can get Indigo to setup vue components on that HTML.


musangowope (approved):

musangowope:
@longhotsummer since you have this. We will have to refactor the way we load existing vue components. I.e Gutter and annotations

longhotsummer:
> @longhotsummer since you have this. We will have to refactor the way we load existing vue components. I.e Gutter and annotations

This still calls `createVueComponents(document)` when the app starts, so nothing has to change (unless I've missed something?)

musangowope:
> > @longhotsummer since you have this. We will have to refactor the way we load existing vue components. I.e Gutter and annotations
> 
> This still calls `createVueComponents(document)` when the app starts, so nothing has to change (unless I've missed something?)

right nevermind.

PR #1310, opened 2021-07-28
Change work place

goose-life:


longhotsummer:
I suspect there's a simpler way to do this that adjusts some form elements on the page, with a searchable dropdown of all places (that is normally hidden). Let's work with Musango to do it.

longhotsummer:
On the server, `WorkForm` must be updated to accept `country` and `locality`.

goose-life:
The current thing definitely doesn't work because `TypeError: Indigo.PlaceChooserView is not a constructor`

longhotsummer (approved):

PR #1311, opened 2021-08-05
Support pdf override

goose-life:
Makes PDFExporter a plugin 
Accompanies https://github.com/laws-africa/indigo-lawsafrica/pull/461


indigo_api/renderers.py
```diff
@@ -147,7 +148,7 @@ def render(self, data, media_type=None, renderer_context=None):
         filename = self.get_filename(data, view)
         renderer_context['response']['Content-Disposition'] = 'inline; filename=%s' % filename
         request = renderer_context['request']
-        exporter = self.get_exporter()
```
longhotsummer, line 150: Doesn't it make more sense for `get_exporter()` to lookup the plugin? The point of `get_exporter` is that the internals of how that exporter is fetched can be modified if necessary.

PR #1314, opened 2021-08-13
Country / locality

longhotsummer:
This makes the javascript and form work.

This was annoyingly complicated. A Vue.js implementation would make it a bit easier.

We still need to update the country and locality of all tasks linked to the work.


longhotsummer:
@goose-life ready for review, I think this is now complete

When button is clicked, it expands:

image

goose-life (approved):

PR #1315, opened 2021-08-16
Docx imports

longhotsummer:
Use indigo html to text conversion for docx files rather than using slaw's. Also delete broken documents that fail
imports, and correct `text` to `.txt`.


goose-life:
lgtm, but I do remember us adding some logic here for compatibility with other plugins, bluebell, etc. 
Did you check how it interacts with all the other moving parts? not to second-guess you, I just remember there being something messy here :sweat_smile:

longhotsummer:
Let me double check.

longhotsummer:
I've double checked -- everything here is good to go.

goose-life (approved):

PR #1316, opened 2021-08-19
Better commenceable provisions

goose-life:
To do:

- [x] beautifier builds up a shadow tree that maps the actual TOCElement tree, and adds decorations there, rather than modifying TOCElement
- [ ] Update templates to take shadow tree into account
- [ ] all_commenceable_provisions incrementally builds and caches the combined TOCs for previous points in time, allowing them to be re-used
- [ ] all_commenceable_provisions strips .element from each TOCElement

### DIFF
--- a/indigo/analysis/toc/base.py
+++ b/indigo/analysis/toc/base.py
@@ -36,7 +36,7 @@ def type_title(typ, language):
 
 
 def descend_toc_pre_order(items):
-    # yields each TOC element and then its children, recursively
+    # yields each item and then its children, recursively
     for item in items:
         yield item
         for descendant in descend_toc_pre_order(item.children):
@@ -44,7 +44,7 @@ def descend_toc_pre_order(items):
 
 
 def descend_toc_post_order(items):
-    # yields each TOC element's children, recursively, ending with itself
+    # yields each item's children, recursively, ending with itself
     for item in items:
         for descendant in descend_toc_post_order(item.children):
             yield descendant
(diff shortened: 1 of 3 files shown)


indigo/analysis/toc/base.py
```diff
@@ -384,6 +384,23 @@ def as_dict(self):
         }
 
 
+class BeautifulElement:
+    def __init__(self, toc_element):
+        self.toc_element = toc_element
+        self.num = toc_element.num.strip('.') if toc_element.num else ''
+        # TODO: this will recurse all the way down, is that okay?
```
longhotsummer, line 391: yup

indigo/analysis/toc/base.py
```diff
@@ -401,13 +418,12 @@ def __init__(self):
...
     def decorate_provisions(self, provisions, assess_against):
+        provisions = [BeautifulElement(p) for p in provisions]
         for p in descend_toc_post_order(provisions):
-            # do this here for all provisions
-            p.num = p.num.strip('.') if p.num else ''
 
             # when self.commenced is True, assess_against is the list of commenced provision ids
             # when self.commenced is False, assess_against is the list of uncommenced provision ids
-            p.commenced = self.commenced if p.id in assess_against else not self.commenced
+            p.commenced = self.commenced if p.toc_element.id in assess_against else not self.commenced
```
longhotsummer, line 426: I'll explain above how you can set it up to avoid having to redo all of this.

indigo/analysis/toc/base.py
```diff
@@ -384,6 +384,23 @@ def as_dict(self):
         }
 
 
+class BeautifulElement:
```
longhotsummer, line 387: We can make each instance of this class act as a proxy for the TOCElement object it wraps. That way, callers don't have to know that some information comes from this object, and some from the proxy.

The idea is to use `__getattr__` to intercept calls to missing attributes, and pass them on to the proxied object. See https://blog.rmotr.com/python-magic-methods-and-getattr-75cf896b3f88

```
class BeautifulElement:
  ... (init as before)...

  def __getattr__(self, attr):
    return getattr(self.toc_element, attr)
```

Then you can do both `beautiful.num` (defined on the beautiful element) and `beautiful.id` (defined on the wrapped `toc_element`).

indigo_app/views/works.py
```diff
@@ -326,24 +326,27 @@ def get_context_data(self, **kwargs):
 
         # decorate all provisions on the work
         commenced_provision_ids = [p_id for c in commencements for p_id in c.provisions]
-        self.beautifier.decorate_provisions(provisions, commenced_provision_ids)
+        provisions = self.beautifier.decorate_provisions(provisions, commenced_provision_ids)
+        # TODO: update template to use p.toc_element, not p, where needed
```
longhotsummer, line 330: With the above suggestion, you don't need to change anything.

indigo/analysis/toc/base.py
goose-life: incredible! thank you

longhotsummer (approved):

PR #1322, opened 2021-09-14
Subtype bug

goose-life:
Resolves import issues where e.g. the subtype is `LN ` instead of `LN`


longhotsummer (approved):

longhotsummer:
Ah yes, stripping blanks from your input. You never think of it until it's a problem!

Issue #1347, opened 2021-10-29
Use crossHeadings and correctly prefixed eIds in attachments


longhotsummer:
This is a series of changes to improve our use of AKN.

* use crossHeadings rather than "fake" crossHeadings via hcontainer
* ensure that eId attributes in attachments include the attachment prefix

## Steps to complete

* [x] update slaw to generate crossHeading
* [x] update slaw to prefix attachment eIds
* [x] migration to migrate existing crossHeadings
* [x] migration to migrate existing attachment eIds
* [x] migration to update annotations bound to attachment eIds
* [x] stop prefixing eIds when rendering to HTML

longhotsummer:
https://github.com/laws-africa/indigo/pull/1348 for Indigo

PR #1348, opened 2021-10-29
Crossheadings and attachment eIds

longhotsummer:
Implements the data migration parts of #1347


indigo_api/data_migrations/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        self.ns = document.doc.namespace
+        if self.migrate_xml(document.doc.root):
+            return True
+
+    def migrate_xml(self, xml):
+        changed = False
+        self.eid_mappings = {}
+
+        for att in xml.xpath('//a:attachment', namespaces={'a': self.ns}):
+            changed = changed or self.migrate_element(att)
```
goose-life, line 66: does this mean only the first att will have migrate_element() called on it?

indigo_api/migrations/0009_migrate_attachment_eids.py
```diff
@@ -0,0 +1,54 @@
...
+                # update sec_1 -> att_1__sec_1
+                Annotation.objects.using(db_alias)\
+                    .filter(document_id=doc.id, anchor_id=old)\
+                    .update(anchor_id=new)
+
+                # update prefixed anchors: att_1/sec_1 -> att_1__sec_1
+                prefix = new.split('__', 1)[0] + '/'
+                assert(prefix.startswith('att_'))
+                old = prefix + old
+                new = prefix + new
```
goose-life, line 31: prefix ends in /, and we don't want that in `new`, right?

indigo_api/migrations/0009_migrate_attachment_eids.py
```diff
@@ -0,0 +1,54 @@
...
+
+    dependencies = [
+        ('indigo_api', '0008_taxonomy_vocabulary'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, migrations.RunPython.noop),
+
+        # strip att_1/ from att_1/att_1__sec_1 in anchor ids
+        migrations.RunSQL(
```
goose-life, line 47: just out of interest, why do we need this if we're updating the Annotation objects above?

indigo_api/data_migrations/__init__.py
longhotsummer: Corrected.

indigo_api/migrations/0009_migrate_attachment_eids.py
longhotsummer: I've updated the comment. We preserve the `att_1/att_1__foo` here, and then strip them everywhere with SQL in the next step.

The SQL ensures that we update all annotations, not just those for re-written documents.

indigo_api/migrations/0009_migrate_attachment_eids.py
longhotsummer: The above code keeps the `att_1/` prefix. This code then strips it from everything.

goose-life (approved):

PR #1351, opened 2021-11-05
Prune deleted documents updated more than 90 days ago

longhotsummer:
* Prune deleted documents updated more than 90 days ago
* Prune document versions older than 90 days, except for the 5 most recent


longhotsummer:
Turns out django_reversion has a command which does exactly this.

goose-life (approved):
nice!

PR #1356, opened 2021-11-18
Unique eids

goose-life:
- [ ] TODO: update package-lock once new indigo-akn is released


indigo_app/static/javascript/indigo/models.js
```diff
@@ -110,27 +110,10 @@
         }
       }
 
-      // ensure attachment ids are correct, we could have deleted an attachment
-      if (oldNode.tagName === 'attachment') {
```
longhotsummer, line 114: I don't think we can remove this. Slaw doesn't know if an attachment has been deleted during a quick edit, which requires eIds of subsequent attachments to be re-numbered. The only place that has sufficient context to do that is here.

goose-life:
Also closing this for the time being (see https://github.com/laws-africa/indigo-akn/pull/5)

PR #1363, opened 2021-11-25
La toc

musangowope:
- Add LaTOCController to DocumentView.vue
- Wire up data to component so it can mimic existing vue component functionality

DEMO:
https://www.loom.com/share/ae278a38da184b3f9d9f6de1ca66aeda


indigo_app/static/stylesheets/_documents.scss
```diff
@@ -227,30 +227,31 @@ button.btn-link.help {
...
+      border: 1px solid $blue;
+      color: $blue;
+    }
+  }
 
-  ol {
-    width: 100%;
+  la-toc-item {
+    .content__action__title:hover {
+      background-color: #e9ecef;
```
longhotsummer, line 249:
```suggestion
      background-color: $gray-200;
```

indigo_app/static/stylesheets/_documents.scss
```diff
@@ -227,30 +227,31 @@ button.btn-link.help {
...
-        margin-right: 2px;
-        i {
-          font-size: 8px;
-          padding: 2px;
-          border: 1px solid $blue;
-          color: $blue;
-        }
-      }
+    &.selected > .content > .content__action > .content__action__title {
+      background-color: #2d7ad4;
```
longhotsummer, line 253:
```suggestion
      background-color: $blue;
```

indigo_app/js/components/DocumentTOCView.vue
```diff
@@ -1,69 +1,29 @@
...
-        ref="tocController"
-        @on-title-click="onTitleClick"
+    <la-table-of-contents-controller
+        ref="la-toc-controller"
+        :items.prop="roots"
+        :title-filter="titleQuery"
+        expand-all-btn-classes="btn btn-primary btn-sm mr-1"
+        collapse-all-btn-classes="btn btn-primary btn-sm mr-1"
+        search-filter-input-classes="form-control form-control-sm"
+        title-filter-placeholder="Search by title"
```
longhotsummer, line 10: Can we now rely on the default placeholder, since that has been updated?

indigo_app/js/components/DocumentTOCView.vue
```diff
@@ -239,8 +214,21 @@ export default {
...
+      for (const tocItem of tocItems) {
+        tocItem.classList.remove('selected');
+        if (tocItem.item.index === i) {
+          tocItem.classList.add('selected');
+        }
+      }
+
       const index = this.selection.get('index');
 
       i = Math.min(this.toc.length - 1, i);
```
longhotsummer, line 234: Does this old logic of setting `.selected` on the TOC entry still apply?

Does the selection get preserved when the TOC is re-rendered after an edit?

longhotsummer: Maybe it would be  easier to respond to the on item rendered event to add the selection class, and re-render the entire tree when the selection changes?

indigo_app/js/components/DocumentTOCView.vue
musangowope: done

longhotsummer (changes requested):
Looking good. Some functionality that still needs to be preserved from the old experience:

* font size for TOC must be marginally smaller so more can fit
* don't wrap long lines, truncate with ellipses instead
* no underline on hover
* give the toc titles a little bit of padding
* remove the 'flash-target' class

indigo_app/js/components/DocumentTOCView.vue
longhotsummer: It looks like this works, I'm not sure why :D

longhotsummer:
Actually, keep `flash-target` but cancel the default action of the TOC item click so that the anchor of the page doesn't update. That would mean that clicking on an internal ref gets the flashed target, but not clicking on the TOC (which is what we want).

indigo_app/js/components/toc-controller/TOCItem.vue
```diff
@@ -1,159 +0,0 @@
...
-          this.$refs[`child_item_${index}`][0].collapseItemAndDescendants();
-        });
-      }
-    }
-  }
-};
-</script>
-
-<style scoped>
-.toc-item {
```
longhotsummer, line 109: Check what you need to preserve of all of this styling, so we don't lose it.

indigo_app/js/components/DocumentTOCView.vue
musangowope: added listener 
```
this.selection.on('change', this.updateLaToc);
```
so listens for selection change that it will update the web component

indigo_app/js/components/toc-controller/TOCItem.vue
musangowope: done: 
note though that I put margin-left: 5px on `content__action__append` instead of the `content__action__title` element. the reason being is that `.content__action__append` conditionally renders so I would rather assign a margin gap to the element that dissapear rather that having a permanent margin-right on `content__action__title`

musangowope:
> Actually, keep `flash-target` but cancel the default action of the TOC item click so that the anchor of the page doesn't update. That would mean that clicking on an internal ref gets the flashed target, but not clicking on the TOC (which is what we want).

I set on the hrefs  of title links to `#` to prevent flash targets. Let me know if you think there is a better way to resolves your request

longhotsummer:
> > Actually, keep `flash-target` but cancel the default action of the TOC item click so that the anchor of the page doesn't update. That would mean that clicking on an internal ref gets the flashed target, but not clicking on the TOC (which is what we want).
> 
> I set on the hrefs of title links to `#` to prevent flash targets. Let me know if you think there is a better way to resolves your request

You could `e.preventDefault()` to stop it doing anything -- we don't actually want to navigate anywhere when the TOC entry is clicked.

indigo_app/js/components/DocumentTOCView.vue
longhotsummer: This seems like only part of the picture to me. `this.selection` is currently how the component communicates the selection to the outside world -- it doesn't actually respond to changes in that (otherwise we'd have had an event handler on that before). So it's a bit of a red herring.

The bigger issue for me is the `.selected` attribute on the TOC entries. That is currently being set but not used. What I was talking about this morning was actually using that to set the `selected` class in `onItemRendered`. If we're not using `.selected` then it should be removed. If we ARE going to drive selections  based on it, then we don`t need to walk over the `la-toc-item` elements and set classes through that mechanism.

Basically: chose one mechanism to declare how we know which item is selected, and use it throughout.

musangowope:
> > > Actually, keep `flash-target` but cancel the default action of the TOC item click so that the anchor of the page doesn't update. That would mean that clicking on an internal ref gets the flashed target, but not clicking on the TOC (which is what we want).
> > 
> > 
> > I set on the hrefs of title links to `#` to prevent flash targets. Let me know if you think there is a better way to resolves your request
> 
> You could `e.preventDefault()` to stop it doing anything -- we don't actually want to navigate anywhere when the TOC entry is clicked.

the option is not available since we're emitting a custom event
https://github.com/laws-africa/la-web-components/blob/cee17004e216d62021f931f3964376bb540cc876/src/components/toc-item/toc-item.tsx#L73

Essentially, if we want to use preventDefault here, we need to emit a `click event`
In vanilla js you would go 
```
 const event = new MouseEvent('click', {
    view: window,
    bubbles: true,
    cancelable: true
  });
cb.dispatchEvent(event)
```
will need to see how to do this in stencil

indigo_app/js/components/DocumentTOCView.vue
musangowope: ok, I have chosen to go with using the `index` to determine which toc item gets the `selected` CSS class. the `attribute` `selected` has been removed and its associated logic

longhotsummer (approved):

longhotsummer (approved):

PR #1367, opened 2021-12-02
Prevent too many tasks from being assigned.

### LINKED ISSUE #1362: As a team manager I want to be able to limit the number of tasks assigned to a person at a time
We have cases where we don't want users to have more than X tasks on their plate at a time.

## Proposal

1. set a global max-assigned-tasks (eg. 3); we could later make this country-specific
2. in the dropdown menus where users can assign tasks:
  a. move current user to the top (convenience)
  b. users who are at the threshold: disabled with reason, eg. “Greg Kempe - too many tasks”
  c. users with a certain permission (?) can assign to anyone, despite being over the threshold
3. this applies to both “open” and “pending review” tasks, counted separately
4. tasks automatically assigned due to “request changes” or similar always get assigned
5. country-based permissions still apply and don’t change
6. add a new "assign to me" button on the task detail page
  a. this is disabled if the user can't be assigned the task, with a short description of why (eg. permissions vs max tasks), to help users learn

## Thoughts

Dropdowns are available in three places:

1. the task detail page
2. the task listing page, underneath the task
3. the task listing page, in the area above all tasks

The latter is a bit more complicated because we may need to limit assignment options between both "open" and "pending review" tasks.

longhotsummer:
Closes #1362

image

image

image

image


indigo_api/models/tasks.py
```diff
@@ -183,9 +184,35 @@ def decorate_potential_assignees(cls, tasks, country):
...
+        # mark users that have too many tasks
+        cls.decorate_user_task_limits(set(u for task in tasks for u in task.potential_assignees), tasks)
 
         return tasks
 
+    @classmethod
+    def decorate_user_task_limits(cls, users, tasks):
+        """ Count assigned tasks for these users and decorate users that have exceeded their threshold.
+        """
+        states = set(t.state for t in tasks)
```
goose-life, line 203: I'm not sure about this. Wouldn't it change `counts` to allow e.g. a reviewer to pick up more tasks if they filter the tasks listing view to only include `pending_review` tasks? or is this decoration all happening before the tasks are filtered? (seems odd if true) 
I guess regardless, I'm a little puzzled by what `states` have to do with `counts`?

indigo_app/templates/indigo_api/_task_assign_to_menu.html
```diff
@@ -0,0 +1,19 @@
...
+  {% endif %}
+
+  {% for potential_assignee in potential_assignees %}
+    <button class="dropdown-item" type="submit" name="assigned_to"
+            {% if potential_assignee.too_many_tasks %}disabled{% endif %}
+            {% if task %}formaction="{% url 'assign_task' place=place.place_code pk=task.pk %}"{% endif %}
+            value="{{ potential_assignee.id }}">
+      {% user_display potential_assignee %}
+      {% if potential_assignee.too_many_tasks %}
+        - too many tasks ({{ potential_assignee.assigned_tasks_count }})
```
goose-life, line 15:
```suggestion
        – too many tasks ({{ potential_assignee.assigned_tasks_count }})
```

indigo_api/models/tasks.py
longhotsummer: The two assignable states we care about are "open" and "pending review". We consider them to have separate task counters. So if the limit of open tasks is 5, they can be assigned 5 open and 5 pending review. That's part of why we take states into account. (I'm open to changing this, I don't have a particularly strong argument for keeping it this way).

In most cases, this function only actually gets a single task and it's simple. The tricky situation is when a user is trying to re-assign multiple tasks through the task listing view, and they choose tasks across both "open" and "pending review". To avoid dodging the limits, we need to effectively count all assigned tasks together.

indigo_api/models/tasks.py
```diff
@@ -183,9 +184,35 @@ def decorate_potential_assignees(cls, tasks, country):
...
+        counts = Task.objects \
+            .filter(assigned_to__in=users, state__in=states) \
+            .values('assigned_to') \
+            .annotate(tasks=Count(1))
+        counts = {c['assigned_to']: c['tasks'] for c in counts}
```
goose-life, line 208: I might not be understanding what `tasks=Count(1)` is doing above. 
From my reading, line 204-207 will produce something like
QuerySet[{'assigned_to': 37, 'tasks': 1}, {'assigned_to': 42, 'tasks': 1}, {'assigned_to': 37, 'tasks': 1}]
where `tasks` will always be `1`? 

and line 208 something like 
{37: 1, 42: 1}
because when `37` is encountered for the second time it'll overwrite it?

longhotsummer: The `tasks=Count(1)` effectively tells SQL to count every row, grouping by `assigned_to`.

```sql
SELECT
  assigned_to,
  count(1)
FROM xxx
GROUP BY assigned_to;
```

https://learnsql.com/blog/difference-between-count-distinct/

longhotsummer: `count(1)` counts every row. It looks a bit weird, I know. I thought it was a bug when I first saw it used :)

goose-life: ah thanks! I looked at the docs online and they only give field names in the examples :upside_down_face:

PR #1377, opened 2022-01-13
new indigo-web, web-components

goose-life:
a few things other than indigo-web and web-components were updated in package-lock with `npm install` – do these look right to you?


longhotsummer (approved):
LGTM.

I suspect we should class web-components as a devDependency, but we can do that next time. It won't change much.

PR #1388, opened 2022-01-20
Date and consolidation phrasing on coverpage

### LINKED ISSUE #1263: As a senior editor I want the option to override a place's "no gazette" note per work
e.g. in Uganda the "No Gazette text" is `Note: This version of the Act was revised and consolidated by the Law Reform Commission of Uganda.` (https://edit.laws.africa/places/ug/settings) but there are a handful of Regulations for which the word "Act" is unsuitable.

A user with a Place Admin badge should be able to override the "No Gazette text" for Uganda for that handful of works.

### Proposal

On the work form, at the bottom under a 'Miscellaneous / Additional information' heading, include a "No Gazette" text field.
Always display the place's "No Gazette" text below it, as a guide and for reference. 
This field should only be visible to a Place Admin.
We don't want to inadvertently save an empty string if there's no intention of overriding the note to be blank on this work; but we do also want the option of doing just that. Maybe a checkbox might be best to make it explicit?

On the work model, add a `no gazette text` field, which can be blank and null.
Don't add anything during the migration.

When generating the coverpage, use the work's "No Gazette" text first (if relevant) (this may be an empty string, in which case don't make an empty block), and default to the place's if the work's is null.

goose-life:
Closes #1263 

See https://docs.google.com/document/d/1HwOTe3tt0tgi9g75ZIVQvPq1X485QUc7ytAlKHmxcHM/edit?usp=sharing for all scenarios

Consolidation (note set on work), latest, with as-at date set for place:
image

Initial publication, no as-at date set for place:
image

Consolidation, as-at date for place is before expression date:
image


indigo_api/templates/indigo_api/akn/coverpage_act.html
```diff
@@ -192,24 +192,48 @@ <h2>{{ subt }}</h2>
...
             [{% trans no_publication_document_text %}]
           </li>
         {% endif %}
         {% endwith %}
       {% endblock %}
+
+      {% block consolidation_note %}
+        {% with document.consolidation_note as consolidation_note %}
+          {% if document.is_consolidation and consolidation_note %}
+            <li class="verification-notice">
```
longhotsummer, line 231: new class?

indigo_api/models/documents.py
```diff
@@ -222,6 +223,89 @@ def element_to_pdf(self, element):
...
+        """
+        latest = self.is_latest()
+        consolidation = self.is_consolidation()
+        expression_date = self.expression_date.strftime('%-d %B %Y') if self.expression_date else None
+        end_date = self.valid_until()
+        if end_date:
+            end_date = end_date.strftime('%-d %B %Y')
+        # Scenario 1: Latest, non-arbitrary; with or without end date
+        if latest and not consolidation:
+            return _(f'[This is the version of this document from {expression_date}' +
```
longhotsummer, line 295: From django manual:

```
    output = _('Today is %(month)s %(day)s.') % {'month': m, 'day': d}
```

PR #1389, opened 2022-01-23
Patch 1

buff0k:


indigo/settings.py
```diff
@@ -337,6 +337,7 @@
 EMAIL_HOST_USER = os.environ.get('DJANGO_EMAIL_HOST_USER')
 EMAIL_HOST_PASSWORD = os.environ.get('DJANGO_EMAIL_HOST_PASSWORD')
 EMAIL_PORT = int(os.environ.get('DJANGO_EMAIL_PORT', 25))
+EMAIL_USE_TLS = os.environ.get('DJANGO_EMAIL_USE_TLS', 'false')
```
longhotsummer, line 340: I don't think this will work as you expect. Django expects `EMAIL_USE_TLS` to be a boolean, not a string. Even if the string is `"false"`, it will be treated as a string, and a non-empty string is always considered `True`.

You probably want this instead:

```python
EMAIL_USE_TLS = os.environ.get('DJANGO_EMAIL_USE_TLS', 'false') == 'true'
```

PR #1394, opened 2022-01-30
eol to br

longhotsummer:
* slaw 12.0.0
* parse and render eol as br
* migration to change eol to br
* update activity view to check if document changed on server, and tell user they must update if it has

Requires https://github.com/laws-africa/slaw/pull/62

* [x] release and use updated slaw
* [x] release and use updated indigo-akn


indigo_api/serializers.py
```diff
@@ -554,6 +556,11 @@ def create(self, validated_data):
         validated_data['document'] = self.context['document']
         return super(DocumentActivitySerializer, self).create(validated_data)
 
+    def get_document_updated_at(self, obj):
+        # load only the field we need
+        doc = Document.objects.undeleted().only('updated_at').filter(pk=obj.document_id).first()
+        return doc and doc.updated_at
```
goose-life, line 562: won't this return True / False?

longhotsummer: Good question. Because `doc.updated_at` is not None, it's considered "truthy" and will be the result of the expression. If it's None, then None will be the result. If `doc` is None (or anything false-y), then None will be the result (and the doc.updated_at won't get evaluated at all).

```python
>>> print(False and 12)
False
>>> print(True and 12)
12
>>> print(False and None)
False
>>> print(True and None)
None
```

longhotsummer: https://realpython.com/python-boolean/ has a pretty good overview.

goose-life: oh awesome thanks!

goose-life (dismissed):

goose-life (approved):

PR #1397, opened 2022-02-02
La gutter replacement

musangowope:
-  Replace gutter vue component with la-gutter and la-gutter-items and refactor annotations logic.

dependent PR
https://github.com/laws-africa/la-web-components/pull/69#issue-841835707


indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -256,7 +257,8 @@
 
       if (range && this.mark(range)) {
         // gutter uses this for positioning
-        this.anchorElement = this.marks[0];
+        // this.anchorElement = this.marks[0];
```
longhotsummer, line 260:
```suggestion
```

longhotsummer: The anchor should be set to null if we can't determine a range (the else part of this if block).

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -474,16 +476,17 @@
...
-      var ix = this.gutter.items.indexOf(view);
-      if (ix > -1) {
-        this.gutter.items.splice(ix, 1);
-      }
+      // var ix = this.gutter.items.indexOf(view);
+      // if (ix > -1) {
+      //   this.gutter.items.splice(ix, 1);
+      // }
+      view.el.remove();
       this.layout();
```
longhotsummer, line 490: I suspect we can remove `this.layout()` both here and everywhere. The gutter should know when to layout.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -541,28 +543,26 @@
...
           // once it's rendered in the gutter, focus it
-          this.gutter.$nextTick(() => {
-            v.focus();
-            v.$el
-             .find('textarea')
-             .first()
-             .focus();
-          });
+
+          // this.gutter.$nextTick(() => {
```
longhotsummer, line 549: This functionality would be nice to preserve -- it means when you create a new comment the textarea is focused automatically.

longhotsummer (commented):
Looking good!

indigo_app/static/javascript/indigo/views/annotations.js
musangowope: done

indigo_app/static/javascript/indigo/views/annotations.js
musangowope: done

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -443,8 +442,11 @@
       this.threadViews = [];
       this.visibleThreads = [];
       this.counts.set({threads: 0});
-      this.threads.forEach(t => this.makeView(t));
-      this.renderAnnotations();
+      //Wait for page to load implicitly
+      window.setTimeout(() => {
```
longhotsummer, line 446: This makes me nervous because it looks like we have a race condition, and you never know if 500ms will be enough. Let's talk through it.

musangowope: Agreed.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -541,28 +532,19 @@
...
-            v.focus();
-            v.$el
-             .find('textarea')
-             .first()
-             .focus();
-          });
-        } else {
-          v.blur();
+          v.el.active = true;
+          window.setTimeout(() => {
```
longhotsummer, line 536: As discussed, let's add an event handler here for a new event from the gutter, something like 'layoutdone', and then set the textarea active once that is fired.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -535,76 +512,58 @@
...
+      this.newButtonTimeout = null;
     },
 
     nextAnnotation: function(e) {
       e.preventDefault();
       e.stopPropagation();
-
-      this.scrollSelected(true);
+      this.gutter.activateNextItem().then((activeItem) => {
+        activeItem.scrollIntoView({ behavior: "smooth" });
```
musangowope, line 557: use an alternative method instead of scrollIntoView. `scrollIntoView` does not have `offsetTop` capabilities. I have tried 
these techniques
https://matthiasott.com/notes/scrolling-elements-into-view
no luck yet.
I have to pop right now, this is on my todo list to change

longhotsummer: Hmm. Double check it's even necessary -- the new annotation should already be visible because we've just highlighted the text that forms the anchor for it.

longhotsummer: > Hmm. Double check it's even necessary -- the new annotation should already be visible because we've just highlighted the text that forms the anchor for it.

Ignore this -- wrong place :)

What does scrollSelected do? Can we just do that?

musangowope: used logic for scrollSelected
https://www.loom.com/share/7a9d56b812f34c3482841ccabef17c24

longhotsummer: Smooth like butter 🧈

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -473,23 +469,23 @@
...
-      var ix = this.gutter.items.indexOf(view);
-      if (ix > -1) {
-        this.gutter.items.splice(ix, 1);
-      }
-      this.layout();
+
+      document.querySelector('.active-thread-number-container .active-number').innerText = "";
+      document.querySelector('.active-thread-number-container .active-divider').innerText = "";
+      this.counts.set({
+        threads: this.visibleThreads.length,
```
musangowope, line 479: this is to fix I bug I discovered. the total thread count does not update when deleting a thread. Checked on `master` does the same thing

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -457,10 +447,16 @@
 
       this.listenTo(view, 'deleted', this.threadDeleted);
       this.listenTo(view, 'closed', this.threadClosed);
-      this.listenTo(view, 'resized', this.layout);
+
+      view.laItemChangedCb = () => {
```
musangowope, line 451: my aim here is to setup a listener for `la-gutter-items` so when it goes `active`  the `html` `action-thread-number-container` can be updated, the problem I am facing at the moment is where to setup this up. the view DocumentView renders twice from what I can see

longhotsummer (dismissed):
Great.

Does this handle all the other places in indigo where we use the gutter (if any)?

musangowope:
> Great.
> 
> Does this handle all the other places in indigo where we use the gutter (if any)?

As far I as know, yes
Indigo signoff (Which covers the points in time page) has a pr for the gutter usage.

longhotsummer (approved):

PR #1398, opened 2022-02-03
Pipelines

longhotsummer:
* Use pipelines for parsing and importing documents. This makes it a lot easier to customise this behaviour.
* Adjust `importer` plugin methods to distinguish between importing from a document, and parsing text.
* smarter docx import via smarter html import (cleaning tables, merging lists, cleaning whitespace, merging adjacent inlines, removing empty inlines)


indigo/pipelines/html.py
```diff
@@ -0,0 +1,318 @@
...
+    <xsl:copy>
+      <xsl:apply-templates select="@*|node()"/>
+    </xsl:copy>
+  </xsl:template>
+
+</xsl:stylesheet>
+"""
+
+    def __call__(self, context):
+        pass
```
goose-life, line 133: should this line come out?

also should the XSL live in a separate file eventually?

indigo/pipelines/html.py
```diff
@@ -0,0 +1,318 @@
...
+
+class NormaliseHtmlTextWhitespace(Stage):
+    """ Strip and normalise whitespace in HTML text.
+
+    Reads: context.html_text
+    Writes: context.html_text
+    """
+    def __call__(self, context):
+        # &nbsp; to space
+        context.html_text = context.html_text.replace('&nbsp;', ' ')
```
goose-life, line 146: Side topic: I'd love for us to keep and somehow support non-breaking spaces at some point

indigo/pipelines/html.py
```diff
@@ -0,0 +1,318 @@
...
+                for cell in row:
+                    style = cssutils.parseStyle(cell.attrib.get('style'))
+
+                    # normalize cell width to % based on total row width
+                    width = cell.attrib.get('width')
+                    if width:
+                        width = int(width.replace('%', ''))
+                        w_pct = math.floor(width / total_row_width * 100)
+                        style['width'] = f'{w_pct}%'
+                        cell.attrib['style'] = style.cssText.replace('\n', ' ')
```
goose-life, line 192: if there are other styles already, will they be overwritten by this?

goose-life: oh wait, I think I see that's the intention :sweat_smile:

goose-life (approved):

PR #1423, opened 2022-02-23
Fix scrolling to active gutter items by increasing scroll top offset

musangowope:
demo for fixing the scroll problem
https://www.loom.com/share/dc190ab4502147f9a7fb12fc03e78a1f

### DIFF
--- a/indigo_app/static/javascript/indigo/views/annotations.js
+++ b/indigo_app/static/javascript/indigo/views/annotations.js
@@ -407,7 +407,6 @@
 
       this.counts = new Backbone.Model();
       this.listenTo(this.counts, 'change', this.renderCounts);
-      this.lastActiveItem = null;
 
       this.threadViews = [];
       this.visibleThreads = [];
--- a/indigo_app/static/javascript/indigo/views/annotations.js
+++ b/indigo_app/static/javascript/indigo/views/annotations.js
@@ -406,6 +406,7 @@
 
       this.counts = new Backbone.Model();
       this.listenTo(this.counts, 'change', this.renderCounts);
+      this.lastActiveItem = null;
 
       this.threadViews = [];
       this.visibleThreads = [];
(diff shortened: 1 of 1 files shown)


indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -543,13 +544,19 @@
     nextAnnotation: function(e) {
       e.preventDefault();
       e.stopPropagation();
-      this.gutter.activateNextItem().then((activeItem) => this.handleScrollForActiveItem(activeItem));
+      this.removeNewButton();
```
longhotsummer, line 547: This seems a bit odd to me. When the selection clears (because the user clicks elsewhere, etc.) then the button should remove itself.

longhotsummer: Clicking the prev/next button isn't particularly special and shouldn't have to do this, AFAIK.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -562,10 +570,9 @@
 
     scrollToElement: function(element) {
       const container = element.closest('.document-sheet-container');
-      const top = element.getBoundingClientRect().top;
       if (container) {
-        container.scrollBy({
-          top: top - container.getBoundingClientRect().top - 70,
+        container.scrollTo({
+          top: parseFloat(element.style.top.replace('px', '')) - 50,
```
musangowope, line 575: I found that this yields the most accurate results

longhotsummer: Nice, takes advantage of what the gutter has already calculated.

indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -406,6 +407,7 @@
 
       this.counts = new Backbone.Model();
       this.listenTo(this.counts, 'change', this.renderCounts);
+      this.lastActiveItem = null;
```
longhotsummer, line 410:
```suggestion
```

longhotsummer (approved):
I think this is fine for a first fix, but I suspect we're going to have to work with the editors to find a better mechanism than starting at the topmost annotation when clicking "next". Too many interactions de-activate the currently active annotation.

PR #1434, opened 2022-03-14
Only set akoma-ntoso property for gutter if akoma-ntoso is not defined

musangowope:
Atm, we are setting the attribute `akoma-ntoso` for `la-gutter` in two ways, either via `annotations.js` or physically setting the attribute I.e `<la-gutter akoma-ntoso="#some-id" />`

As it stands there are cases where a `la-gutter` element will already have an attribute `akoma-ntoso` set, but `annotation.js` will still run and set the gutter's attribute again

We should generally rely on the dom to set the akomaNtoso for la-gutter. If the `akoma-ntoso` attribute is not set to an `la-gutter` element , then run js to find the nearest la-akoma-ntoso  for la-gutter.

It just makes it a bit more explicit


indigo_app/static/javascript/indigo/views/annotations.js
```diff
@@ -402,8 +402,15 @@
...
 
-      this.gutter = this.el.querySelector('.gutter la-gutter');
-      this.gutter.akomaNtoso = this.el.querySelector('.content-with-gutter > .content');
+      this.gutter = this.el.querySelector('la-gutter');
+
+      /*
+      * la-gutter components should have the akoma-ntoso as a attribute set. In the case that it is not dont. Find
+      * the the nearest akomo-ntoso element
+      * */
+      if(!this.gutter.akomaNtoso) {
```
longhotsummer, line 411: I'm trying to work out when this would happen - surely it would be a bug? If a gutter isn't linked to its la-akoma-ntoso, that's a problem.

longhotsummer: ooking in la-web-components: it already defaults to the first la-akoma-ntoso on the page: https://github.com/laws-africa/la-web-components/blob/master/src/utils/linking.ts#L11

so we could drop this completely?

longhotsummer (approved):

PR #1435, opened 2022-03-16
in get_exporter, get the right exporter based on country, language, locality

goose-life:


indigo_api/renderers.py
```diff
@@ -168,8 +168,8 @@ def render(self, data, media_type=None, renderer_context=None):
 
         return pdf
 
-    def get_exporter(self, *args, **kwargs):
-        return plugins.for_locale('pdf-exporter')
+    def get_exporter(self, data, *args, **kwargs):
+        return plugins.for_locale('pdf-exporter', country=data.country, language=data.language.code, locality=data.locality)
```
longhotsummer, line 172: `data` may be a list, possibly covering multiple localities and multiple languages. It's not actually as useful as it sounds, but legacy code expects it.

Perhaps just pass in a country, either of `data` or `data[0]`?

Alternatively override `get_exporter` in a subclass somewhere to force a particular exporter or locale.

PR #1440, opened 2022-04-08
Use updated eid rewriter and re-write all attachments

longhotsummer:
JS must be recompiled before this is merged, once  https://github.com/laws-africa/indigo-akn/pull/14 is merged


longhotsummer:
This has been updated to preserve the old mechanism of rewriting eIds.

goose-life:
Should indigo-akn be compiled in again?

goose-life (approved):

PR #1486, opened 2022-06-06
Page counts

goose-life:
Country cards
image

Locality cards
image

Place overview
image

Works list (with filter applied)
image

Taxonomy
image


indigo_app/views/base.py
```diff
@@ -6,6 +9,16 @@
 from indigo_api.models import Country
 
 
+def page_count(works):
+    page_count = 0
+    for w in works:
+        for d in w.document_set.undeleted():
+            xml = etree.fromstring(d.document_xml)
+            text = '\n'.join(x.strip() for x in xml.xpath('//a:*//text()', namespaces={'a': xml.nsmap[None]}))
+            page_count += ceil(len(text.split()) / 250)
```
longhotsummer, line 18: i'm not following the logic here -- why join with newlines?

Are you estimating 250 words per page?

longhotsummer:
Nice start. My major concern is the cost of calculating this. We have to load, parse and then process the XML for every single document, across all places. That's super expensive and will make the page load times very slow.

Some thoughts to address that:

1. make it asynchronous and only load it on certain pages, eg the place overview page. We could then load it via JS once the page has loaded, and put a 30 minute cache on the value.
2. we could calculate it daily, say, and store it in the db on a per-place basis. Perhaps along with other per-place stats (word count? byte count?). Then it's cheap to load and we can show it anywhere.

longhotsummer:
We already have background metrics tasks which count the number of works in a place, etc. We could build on that, see https://github.com/laws-africa/indigo/blob/master/indigo_metrics/models.py#L172

indigo_app/views/base.py
goose-life: Joining with newlines was easier for debugging when I was checking it was doing the right thing; updating to be a little simpler.

And yes, 250 words per page.

goose-life:
From slack: 

> why don’t we create a DocumentMetrics object or similar that stores the number of bytes, words, provisions and pages for a document, and use that to calculate work- and place-based metrics

longhotsummer (approved):
Small suggestion to prevent too much empty space.

PR #1490, opened 2022-06-09
Add french locale translation placeholders

longhotsummer:
These are effectively empty and must be merged to master in order for crowdin to pick them up and allow for translation.

### DIFF
--- a/indigo/pipelines/text.py
+++ b/indigo/pipelines/text.py
@@ -1,7 +1,7 @@
 import re
 import tempfile
 
-from django.utils.translation import ugettext as _
+from django.utils.translation import gettext as _
 from lxml import etree
 from cobalt.akn import AkomaNtosoDocument
 


indigo/bulk_creator.py
```diff
@@ -199,13 +207,17 @@ def get_datatable_gsheets(self, spreadsheet_id, sheet_name):
...
+                    self._gsheets_secret['client_email'],
+                ))
 
         rows = result.get('values', [])
         if not rows or not rows[0]:
-            raise ValidationError("There doesn't appear to be data in sheet {} of {}".format(sheet_name, spreadsheet_id))
+            raise ValidationError(_("There doesn't appear to be data in sheet %(name)s of %(id)s") % {
+                'name': sheet_name,
+                'id': spreadsheet_id
+            })
```
goose-life, line 220:
```suggestion
            }))
```

longhotsummer: It's currently correct: `raise ValidationError(_("...") % { ... })`.

The interpolation happens after the `_(...)` translation.

indigo/pipelines/text.py
```diff
@@ -1,6 +1,7 @@
 import re
 import tempfile
 
+from django.utils.translation import ugettext as _
```
goose-life, line 4: should this be `gettext` as in bulk_creator?

longhotsummer: whoops, yes. `ugettext` is deprecated since python 3 and is now the same as `gettext`

longhotsummer, line 4:
```suggestion
from django.utils.translation import gettext as _
```

indigo_app/templates/account/email.html
```diff
@@ -13,8 +13,11 @@ <h5 class="card-header">Email addresses</h5>
   <div class="card-body">
     {% with user.emailaddress_set.all as emails %}
       {% if emails %}
-        <p>You can use any of these email addresses to log in.</p>
-        <p><strong>Note:</strong> We only send email notifications to your <strong>primary</strong> email address.</p>
+        <p>{% trans 'You can use any of these email addresses to log in.' %}</p>
+        <p>
+          <strong>{% trans 'Note:' %}</strong>
+          {% blocktrans %}We only send email notifications to your <strong>primary</strong> email address.{% endblocktrans %}
```
goose-life, line 19: could the 'Note: ' not also go into the blocktrans? or do we want 'Note: ' separately to reuse?

indigo_app/templates/account/login.html
```diff
@@ -75,12 +75,15 @@ <h1 class="mb-3">Log In</h1>
...
+        {% blocktrans with provider=provider.name %}Log in with {{provider}}{% endblocktrans %}
       </a>
     {% endfor %}
   </div>
 </div>
 
-<div class="mt-3 text-muted text-center">By logging in you agree to our <a href="{% url 'terms_of_use' %}">Terms of Use &amp; Privacy Policy</a>.</div>
+<div class="mt-3 text-muted text-center">
+  {% url 'terms_of_use' as tos_url %}
+  {% blocktrans %}By logging in you agree to our <a href="{{ tos_url }}">Terms of Use and Privacy Policy</a>.{% endblocktrans %}
```
goose-life, line 86: no 'with tos_url=tos_url' needed after `blocktrans`? Would this be because of the use of `as` on 85?

indigo_app/templates/indigo_api/document/_attachments.html
```diff
@@ -58,13 +60,11 @@ <h5>or</h5>
...
-{% verbatim %}
 <form>
   <div class="mb-2"><input type="text" class="form-control form-control-sm" name="filename" required></div>
   <div>
-    <button class="btn btn-sm btn-success save-attachment" type="submit">OK</button>
-    <button class="btn btn-sm btn-outline-secondary cancel">Cancel</button>
+    <button class="btn btn-sm btn-success save-attachment" type="submit">{% trans 'OK' %}</button>
+    <button class="btn btn-sm btn-outline-secondary cancel">{% trans 'Cancel' %}</button>
   </div>
 </form>
```
goose-life, line 69: Sure we didn't need the verbatim around this form?

indigo_app/templates/indigo_api/document/show.html
```diff
@@ -37,18 +36,28 @@ <h5 class="main-header-title">
               {% if request.user.is_authenticated %}
                 {% if user_can_edit %}
                   <div class="btn-group save-btn-group">
-                    <button class="btn btn-success save" disabled><span class="if-published">Save &amp; publish</span><span class="if-draft">Save draft</span></button>
+                    <button class="btn btn-success save" disabled>
+                      <span class="if-published">{% trans 'Save &amp; publish' %}</span>
```
goose-life, line 40: will `&amp;` be fine here?

indigo_app/templates/indigo_api/document/_attachments.html
longhotsummer: No, verbatim is only necessary so that `{{ foo }}` is ignored, because handlebars uses the same syntax. We don't have those here, so it's not necessary.

indigo_app/templates/account/login.html
longhotsummer: Yes. Blocktrans has in the surrounding context available to it, and `tos_url` is now part of the context.

indigo_app/templates/indigo_api/document/show.html
longhotsummer: It seems this is the correct way to do it :/

PR #1493, opened 2022-06-17
use latest indigo-web (8.1.2)

goose-life:


longhotsummer:
Unfortunately we still need to import various things directly from this, so the updated indigo-web must also be copied into 
 `indigo_app/static/lib/indigo-web/scss`

🙄

goose-life:
I was surprised by the extent of the changes to the scss but I'm guessing it's just been a while since we needed the latest indigo-web for indigo, so the restructuring hadn't been pulled in?

longhotsummer:
> Wow, yes, it has been ages. Let me see if now is a good time to get rid of that completely...

Everything except PDF exports now uses la-web-components. I'm looking into what we can do for PDF.

longhotsummer:
Let's merge https://github.com/laws-africa/indigo/pull/1495 first, then this should be partially redundant

longhotsummer (approved):

PR #1543, opened 2022-08-03
no numbered title for Acts with text numbers

goose-life:


goose-life:
I considered nuking `no_numbered_title_numbers = ['constitution']` and the related check, but I think this could still be useful across the board (other doctypes, all subtypes). 
So while it's redundant in the case of an Act with the number 'constitution', I think we should keep it for the time being.

longhotsummer (approved):

PR #1568, opened 2022-09-23
Work frbr fields

longhotsummer:
Store FRBR URI fields explicitly on the work. This makes reporting and analysis from the db simpler.


indigo_api/migrations/0015_work_frbr_fields.py
```diff
@@ -0,0 +1,58 @@
...
+        migrations.AlterField(
+            model_name='task',
+            name='extra_data',
+            field=models.JSONField(blank=True, null=True),
+        ),
+        migrations.AlterField(
+            model_name='work',
+            name='properties',
+            field=models.JSONField(blank=True, default=dict),
+        ),
```
goose-life, line 57: Are these alterations part of this change? or just updates to allow None?

goose-life (approved):
I'm not sure I get the significance of moving the properties off the DocumentMixin and onto the Document; or why the `work_uri`'s properties are being used there rather than the `work`'s.

but lgtm!

indigo_api/migrations/0015_work_frbr_fields.py
longhotsummer: These are old ones only showing up now.

longhotsummer:
> I'm not sure I get the significance of moving the properties off the DocumentMixin and onto the Document; or why the `work_uri`'s properties are being used there rather than the `work`'s.
> 
> but lgtm!

Otherwise they clash with signoff. They do look a bit odd in this PR :)

PR #1575, opened 2022-10-05
Sentry Integration

### LINKED ISSUE #1573: Migrate performance metrics to Sentry
Configure the project to use Sentry for APM and drop Elastic APM support.
* [ ] update indigo-lawsafrica, indigo-unep, indigo-la-tradition, indigo-kenyalaw to reflect this change

ro6ley:
This PR adds Sentry to Indigo.

closes #1573


indigo/apps.py
```diff
@@ -9,3 +16,20 @@ def ready(self):
         # ensure our plugins are pulled in
         import indigo.analysis      # noqa
         import indigo.bulk_creator  # noqa
+
+        # setup sentry error reporting
+        if not settings.DEBUG and os.environ.get('SENTRY_DSN'):
```
longhotsummer, line 21: I think getting the SENTRY_DSN from settings.py, where it is taken from the environ, is better. A single place to store config.

longhotsummer: Also, can't we do all this in settings.py directly?

indigo/signals.py
```diff
@@ -0,0 +1,37 @@
+from django.conf import settings
+from django.dispatch import receiver
+from background_task.signals import task_started, task_successful, task_error, task_finished
+import sentry_sdk
+
+
+# monitor background tasks
+@receiver(task_started)
+def bg_task_started(sender, **kwargs):
+    transaction = sentry_sdk.start_transaction(op="task")
```
longhotsummer, line 10: Does this error at all if sentry isn't configured?

longhotsummer:
I think this should replace all the logic related to `REPORT_JS_ERRORS`?

ro6ley:
> I think this should replace all the logic related to `REPORT_JS_ERRORS`?

Yes

indigo/signals.py
longhotsummer: Since it's pretty standard to pull `signals.py` into an app, I don't think we should rely on that. Rather, assume it'll be imported and check within each signal handler if sentry is expected to be configured (eg. check settings for the DSN).

longhotsummer:
> > I think this should replace all the logic related to `REPORT_JS_ERRORS`?
> 
> Yes

Great -- please clear it out :)

indigo/signals.py
ro6ley: Effectively transfer the DSN availability check from apps.py to the signals?

longhotsummer: Yes. Because otherwise we can't put any other handlers into `signals.py` that may not be related to sentry.

indigo/settings.py
```diff
@@ -482,3 +484,19 @@
...
+    sentry_logging = LoggingIntegration(
+        level=logging.INFO,  # Capture info and above as breadcrumbs
+        event_level=None,  # Don't send errors based on log messages
+    )
+
+    sentry_sdk.init(
+        dsn=SENTRY_DSN,
+        integrations=[DjangoIntegration(), sentry_logging],
+        send_default_pii=True,
+        traces_sample_rate=0.5,  # sample 50% of requests for performance metrics
```
longhotsummer, line 501: Pls make this an env variable.

indigo/settings.py
```diff
@@ -482,3 +484,23 @@
 # Each item in the list should be a tuple of (Full name, email address). Example:
 # [('Indigo Admin', 'indigoadmin@example.com')]
 ADMINS = []
+
+# Sentry configuration
+SENTRY_DSN = os.environ.get('SENTRY_DSN')
+
+# Sample 50% of requests for performance metrics by default or use the configured
+# environment variable which is a number between 0 (0% of errors sent) and 1 (100% of errors sent).
+SENTRY_SAMPLE_RATE = os.environ.get('SENTRY_SAMPLE_RATE', 0.5)
```
longhotsummer, line 493: i think you need to parse the ENV variable value (which will be a string) into a float

PR #1580, opened 2022-10-18
Document Edit Activity

### LINKED ISSUE #1579: As an admin I want to be able to track the amount of time spent making changes to a document
This will help us to drive this number down.

## Proposal

In the document view, we want to track when the user clicks "edit" to when they click "update". If they click "cancel", we ignore the edit. Additionally, we only care about edits if the user is actually saving them. So we should record the time spent editing and POST it to the server when the user clicks the document-wide "Save" button.

### Backend

* A new model in `indigo_metrics`, such as `DocumentEditActivity`
* Model has: document_id, started_at, ended_at, user_id
* Model also has a derived `duration_secs` field which records `ended_at - started_at` in seconds and is updated during `save()`
* A new internal POST-only API to write bulk records; only the started_at and ended_at are necessary, the document ID is derived from the URL and the userid is the currently authenticated user

### Frontend

* Business logic to respond to the "edit", "update" and "cancel" events.
* We may need to add global events for these, or it must be tied into the existing views/classes
* Logic to send records to the server when "Save" is clicked

ro6ley:
This PR adds functionality to track the amount of time spent making edits:

Backend:
* [x] New `DocumentEditActivity` model w/ migration, view and serializer
* [x] Post API to handle list of edit activities

Frontend:
* [ ] _pending_

### Endpoint Details
* The new API endpoint is: `/api/documents/<document_id>/activity/edits`
* Sample payload:
  ```json
  [
      {
          "started_at": "2022-10-18T13:41",
          "ended_at": "2022-10-18T13:49"
      },
      {
          "started_at": "2022-10-18T13:50",
          "ended_at": "2022-10-18T14:01"
      }
  ]
  ```

closes #1579


ro6ley:
@musangowope feel free to branch off or work from this branch

longhotsummer (commented):
LGTM.

Up to Musango to tie into this now.

longhotsummer (approved):

PR #1595, opened 2022-10-21
add bluebell-manaco as a dependency

musangowope:
resolves https://github.com/laws-africa/indigo-la-tradition/issues/163


longhotsummer:
Wrong repo, this must be done in https://github.com/laws-africa/indigo-la-tradition

PR #1599, opened 2022-10-25
Main Commencement Tweak

ro6ley:
Revert `Work.main_commencement()`


longhotsummer:
Why do we need to revert it?

ro6ley:
> Why do we need to revert it?

Sorry, misread the slack message.

PR #1605, opened 2022-10-27
add padding bottom to toc so items are above label

musangowope:
resolves https://github.com/laws-africa/indigo-lawsafrica/issues/609

demo:
https://www.loom.com/share/87806599cca8405db3d47b5ed419f29e


longhotsummer (approved):
nice

PR #1612, opened 2022-11-03
Publication dates optional on country object

longhotsummer:
Moves this setting out of code into a database object

image

Optional:

image

Required:

image


ro6ley:
LGTM, the migrations are out of sync with the master branch

PR #1632, opened 2022-11-21
add work_amendments to PublishedDocumentSerializer

goose-life:


indigo_content_api/v2/serializers.py
```diff
@@ -68,6 +68,19 @@ def get_url(self, instance):
...
 
 
+class AmendmentSerializer(serializers.ModelSerializer):
+    date = serializers.DateField(source='date')
+    amending_title = serializers.CharField(source='amending_work.title')
+    amending_uri = serializers.CharField(source='amending_work.frbr_uri')
+
+    class Meta:
+        model = Amendment
+        fields = (
```
longhotsummer, line 78: I think this must be `('date', 'amending_title', 'amending_uri')` -- all the fields being included in the serializer. The `date` field shouldn't need to be declared explicitly above, because it should be worked out based on the model field.

longhotsummer (approved):

PR #1639, opened 2022-11-28
Error report response.text() and editTimes

musangowope:
Refactor error of edit times in an attempt to get a better sense of what is happening with error could not be saved

The following changes were made:
- Report `editTimes` 
- We report `response.text()`

After testing locally this the error reporting looks like on sentry:
image


goose-life (approved):
lgtm

PR #1645, opened 2022-12-02
Bulk updater

goose-life:
This is for the start of the bulk updater, where update columns can be selected.

The logic for doing the literal updates still needs to be worked out.

https://www.loom.com/share/c50837d76b1d40c8bac564c948f54c7b

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -309,54 +309,51 @@ class BaseBulkCreator(LocaleBasedMatcher):
             work.frbr_uri = frbr_uri
             work.country = self.country
             work.locality = self.locality
+            for attribute in ['title',
+                              'publication_name', 'publication_number',
+                              'assent_date', 'publication_date',
+                              'commenced', 'stub', 'principal']:
+                setattr(work, attribute, getattr(row, attribute, None))
             work.created_by_user = self.user
-            self.set_work_attributes_and_save(work, row)
+            work.updated_by_user = self.user
+            self.add_extra_properties(work, row)
+
+            try:
+                work.full_clean()
+                work.set_frbr_uri_fields()
+                if not self.dry_run:
+                    work.save_with_revision(self.user)
+
+                    # signals
+                    if not self.testing:
+                        work_changed.send(sender=work.__class__, work=work, request=self.request)
+
+                # info for linking publication document
+                row.params = {
+                    'date': work.publication_date,
+                    'number': work.publication_number,
+                    'publication': work.publication_name,
+                    'country': self.country.place_code,
+                    'locality': self.locality.code if self.locality else None,
+                }
+
+                self.link_publication_document(work, row)
+
+                if work.principal:
+                    self.create_task(work, row, task_type='import-content')
+
+                row.work = work
+                row.status = 'success'
+
+            except ValidationError as e:
+                if hasattr(e, 'message_dict'):
+                    row.errors = ' '.join(
+                        ['%s: %s' % (f, '; '.join(errs)) for f, errs in e.message_dict.items()]
+                    )
+                else:
+                    row.errors = str(e)
 
         return row
-
-    def set_work_attributes_and_save(self, work, row):
-        for attribute in ['title',
-                          'publication_name', 'publication_number',
-                          'assent_date', 'publication_date',
-                          'commenced', 'stub', 'principal']:
-            setattr(work, attribute, getattr(row, attribute, None))
-        work.updated_by_user = self.user
-        self.add_extra_properties(work, row)
-
-        try:
-            work.full_clean()
-            work.set_frbr_uri_fields()
-            if not self.dry_run:
-                work.save_with_revision(self.user)
-
-                # signals
-                if not self.testing:
-                    work_changed.send(sender=work.__class__, work=work, request=self.request)
-
-            # info for linking publication document
-            row.params = {
-                'date': work.publication_date,
-                'number': work.publication_number,
-                'publication': work.publication_name,
-                'country': self.country.place_code,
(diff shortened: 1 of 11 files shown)


indigo_app/templates/place/works.html
```diff
@@ -14,7 +14,8 @@
         {% if perms.indigo_api.bulk_add_work %}
           <button type="button" class="btn btn-success dropdown-toggle dropdown-toggle-split" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"><span class="sr-only">Toggle Dropdown</span></button>
           <div class="dropdown-menu">
-            <a class="dropdown-item" href="{% url 'new_batch_work' place=place.place_code %}">New work batch</a>
+            <a class="dropdown-item" href="{% url 'new_batch_work' place=place.place_code %}">Bulk importer</a>
+            <a class="dropdown-item" href="{% url 'update_batch_work' place=place.place_code %}">Bulk updater</a>
```
longhotsummer, line 18: It's better if these are a verb phrase, rather than a noun. Describe the thing that the user will do. eg. "Bulk import" and "Bulk update".

eg. we don't use "Work editor", we use "Edit work".

longhotsummer (commented):
I think the UI is good -- now just for the logic.

indigo/bulk_creator.py
```diff
@@ -1017,3 +1025,108 @@ def update_works_list(self, work):
...
+
+@plugins.register('bulk-updater')
+class BaseBulkUpdater(BaseBulkCreator):
+    """ Update works in bulk from a google sheets spreadsheet.
+    """
+    # TODO: get these core fields from somewhere else? cobalt / FRBR URI fields? add 'actor'?
+    core_fields = ['country', 'locality', 'doctype', 'subtype', 'number', 'year']
+    update_columns = None
+
+    def get_rows_from_table(self, table):
```
longhotsummer, line 1038: I would have thought all of this could be inherited from the importer class?

goose-life:
https://www.loom.com/share/133b02bba0e6490390e48247b1a42e1c

indigo/bulk_creator.py
```diff
@@ -93,10 +96,63 @@ def get_doctypes_for_country(self, country_code):
                 settings.INDIGO['DOCTYPES'] +
                 settings.INDIGO['EXTRA_DOCTYPES'].get(country_code, [])]
 
+    def sanitize_incoming(self, data):
```
longhotsummer, line 99: A more Django-esque approach to this would be to do this in `clean_<field>`, or tweak the form widget: https://stackoverflow.com/questions/24313201/convert-all-charfield-form-field-inputs-to-lowercase-in-django-forms

But it doesn't matter that much.

indigo/bulk_creator.py
```diff
@@ -352,6 +404,18 @@ def create_work(self, row, idx):
                 else:
                     row.errors = str(e)
 
+    def create_work(self, row, idx):
+        # handle spreadsheet that still only uses 'principal'
+        row['stub'] = row.get('stub') if 'stub' in row else not row.get('principal')
```
longhotsummer, line 409: can this be handled by `clean`, like you do with commenced etc.?

indigo/bulk_creator.py
```diff
@@ -352,6 +404,18 @@ def create_work(self, row, idx):
...
+        row = self.validate_row(row)
```
longhotsummer, line 410: `validate_row` takes raw data and gives you something structured back. It's therefore responsible for giving you back something useful. It should also take care of `status` and `row_number`, otherwise you're splitting responsibility for the pristine state of `row` between two different methods.

longhotsummer (commented):
This is looking much cleaner with less duplication.

`create_or_update` still has some duplication. You could break it into helper methods that do the actual work (eg. link the publication document) to reduce duplication of code.

indigo/bulk_creator.py
goose-life: `clean_x` only gets called it if the form is valid?, and because the choices are case-sensitive it gets rejected before we get here.

indigo/bulk_creator.py
goose-life: by the time `clean` gets a hold of it, `False` has already been assigned to 'stub' if it's missing from the actual spreadsheet, so I'm suggesting doing it during `sanitize_incoming` instead, when we still have access to the row data and can check if 'stub' is in the dict.

goose-life:
I can _almost_ get rid of sanitize_incoming; just need to figure out a way of handling spreadsheets that don't have 'stub' in it.
I tried subclassing BooleanField as Stubfield to intercept it there, but it's already False if 'stub' is missing from the dict, plus we don't have access to the rest of the data at that point. 
I'm just moving it into init for now; any  earlier and we're doing it outside the form, and any later and it'll be set to False and we won't know that it's missing from the original row data.

indigo/bulk_creator.py
longhotsummer: `clean` has access to both `self.data` and `self.cleaned_data` -- so I think you could still check the original `data` and adjust as necessary.

indigo/bulk_creator.py
```diff
@@ -27,18 +28,25 @@ def __init__(self, data, errors):
         self.relationships = []
         self.tasks = []
         self.taxonomies = []
+        self.status = None
         for k, v in data.items():
             setattr(self, k, v)
 
 
+class LowerChoiceField(forms.ChoiceField):
```
longhotsummer, line 36: Nice

indigo/bulk_creator.py
goose-life: thanks!

longhotsummer (approved):

PR #1683, opened 2023-03-06
Allow unicode characters in work's number field

### LINKED ISSUE #1682: Work form and bulk importer don't allow unicode characters in `number` field
I think we went with a-zA-Z or similar in the regex for this form validation — which also excludes à, é, è, etc

https://edit.laws.africa/works/akn/aa-ohada/act/2017/comptable-et-a-l-information-financiere-systeme-comptable/edit/

goose-life:
closes #1682 

- [x] Python (bulk importer)
- [x] JavaScript (work form)


goose-life:
I'm not sure how to do the JavaScript, it's here: https://github.com/laws-africa/indigo/blob/master/indigo_app/static/javascript/indigo/views/work.js#L189

using `\w` with the `u` flag doesn't seem to work — at least not on Chrome.

longhotsummer (approved):

PR #1709, opened 2023-05-01
Support all AKN elements in HTML: update XSL

goose-life:
See https://github.com/laws-africa/indigo-lawsafrica/issues/835


indigo_api/static/xsl/html_akn.xsl
```diff
@@ -55,17 +55,41 @@
     </xsl:attribute>
   </xsl:template>
 
-  <xsl:template match="a:article">
+  <!-- containers with headings on the next line -->
+  <xsl:template name="container-1">
```
longhotsummer, line 59: Minor: `container` has a specific meaning in AKN. Perhaps `hier-heading-next-line` or similar is better?

longhotsummer (dismissed):
Approved, but I think the `container` template name could be less confusing.

goose-life:
@longhotsummer I made a further change to the template to cut down in repetition hence the re-request

longhotsummer (approved):
great

PR #1720, opened 2023-05-05
Block or cancel tasks on bulk creation

goose-life:
image

image

image

image

### DIFF
--- a/indigo/bulk_creator.py
+++ b/indigo/bulk_creator.py
@@ -1148,11 +1148,29 @@ Possible reasons:
         task.code = task_type
         task.created_by_user = self.user
 
-        # need to save before assigning workflow because of M2M relation
+        # save the task before updating it
         task.save()
+        self.update_task(task, task_type)
+
+        return task
+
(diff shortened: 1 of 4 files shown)


indigo/bulk_creator.py
```diff
@@ -819,10 +833,14 @@ def link_amendment_passive(self, row):
         if not amending_work:
             return self.create_task(row.work, row, task_type='link-amendment-passive')
 
-
         if self.dry_run:
             row.relationships.append(f'Amended by {amending_work}')
-            row.notes.append("An 'Apply amendment' task will be created on this work")
+            note = "An 'Apply amendment' task will be created on this work"
+            if self.cancel_amendment_tasks:
+                note += ' and CANCELLED'
```
longhotsummer, line 840: Why not simply not create the task? So skip, rather than cancel?

goose-life: We'd rather have a record of the task having been created in the system, so that it can be discussed, reopened if needed, or just commented on.

PR #1725, opened 2023-05-10
Make hierarchicalizer subclassable

### LINKED ISSUE #1722: Allow hierarchicalizer to be chosen based on locale during import
See #1704

### LINKED ISSUE #1728: Catch all paragraphs on import
Even if they have an italicised dash before the opening bracket

goose-life:
closes #1722 
closes #1728 

TODO:
- [x] Add test for lstrip text before matching as there might be (potentially significant) inline whitespace at the start of lines


longhotsummer (approved):

indigo/pipelines/hier.py
```diff
@@ -30,7 +30,8 @@ class IdentifySections(Stage):
 
     def __call__(self, context):
         for p in context.html.xpath('./p'):
-            text = ''.join(p.xpath('.//text()'))
+            # lstrip text in case there's inline whitespace at the start
+            text = ''.join(p.xpath('.//text()')).lstrip()
```
longhotsummer, line 34: Can you tweak one of the existing test fixture files so it covers this?

longhotsummer: ie. it would have failed before but now passes. So that we don't regress later accidentally.

goose-life: good idea, thanks!

goose-life:
for the record the test I'm about to add does fail if we remove the .lstrip() 

image

indigo_api/tests/test_importers.py
```diff
@@ -362,6 +362,31 @@ def test_subparagraphs_1(self):
...
+  PARAGRAPH (c)
+
+    electoral officer.
+""".strip(),
+
+            self.html_to_bluebell("""
+<p>	(<em>a</em>)	chief electoral officer;</p>
+<p>	(<em>b</em>)	deputy chief electoral officer; and</p>
+<p>
+<em>	</em>(<em>c</em>)	electoral officer.</p>
```
longhotsummer, line 389: Can you also add a case with whitespace before the "c" within the element: `<em>  c</em>`.

longhotsummer: actually, more like `<p>  <em>   (c)</em> saoeu</p>`

longhotsummer: Also add one with `<u>` because `RemoveEmptyInlines` removes a, b, i, sup and sub but not u.

goose-life: I can add it; it won't pass as is. We haven't encountered this yet as far as I know?

longhotsummer (approved):

PR #1746, opened 2023-05-22
Paragraphs in schedules

### LINKED ISSUE #1734: Basic units in Schedules should be paragraphs, whether they have headings or not
Let's add an additional stage to the hierarchicalizer pipeline that transforms sections inside attachments into paragraphs.

goose-life:
closes #1734


longhotsummer (approved):
nice

PR #1761, opened 2023-06-15
Multiple amendments at same date

### LINKED ISSUE #1756: As a reader of a document that contains multiple amendments, I want them listed in order on the coverpage
Obviously this can't be in order of amendment date, because they're all on the same date.

Publication date would be the best second-tier ordering, but if not all the amending works have a publication date, we'll need a third tier to fall back to — perhaps year, then number? 
That would help when the publication dates are present but the same in any case.

goose-life:
closes #1756

### DIFF
--- a/indigo_api/models/documents.py
+++ b/indigo_api/models/documents.py
@@ -372,6 +372,13 @@ class Document(DocumentMixin, models.Model):
         else:
             return []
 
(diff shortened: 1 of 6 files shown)


longhotsummer (approved):
nice

PR #1763, opened 2023-06-16
Publication document description

goose-life:
TODO: 
- [x] Translation strings are working inconsistently; but see e.g. https://github.com/laws-africa/indigo/pull/1763/files?diff=split&w=1#diff-5087eb164363eb47b5ac61af1ece5c993c8619561703a6a7536f06ab2831e50dR220-R253
- [x] Update: it seems this will be taken care of once it goes through CrowdIn, once the `fuzzy` label is removed: https://stackoverflow.com/questions/2328185/django-i18n-common-causes-for-translations-not-appearing


indigo_api/locale/fr/LC_MESSAGES/django.mo
longhotsummer, line 1: these .mo files shouldn't be deleted. they keep the other translations working until we have translated these new strings.

longhotsummer (changes requested):
Awesome, so much simpler.

Just the .mo files to keep around in this PR. Copy them in from master if necessary.

longhotsummer (approved):

PR #1773, opened 2023-06-23
update bluebell dependency to use bluebell-akn

longhotsummer:
this is now published on pypi

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
         'django>=3.2,<4',
         'boto3>=1.7',
         'cobalt>=6.1',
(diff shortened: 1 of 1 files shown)


goose-life (approved):

setup.py
```diff
@@ -43,6 +43,7 @@
         'django>=3.2,<4',
         'boto3>=1.7',
         'cobalt>=6.1',
```
goose-life, line 45: should we remove this from indigo, given bluebell has cobalt 7 and higher in its dependencies?

longhotsummer: nice, done.

PR #1777, opened 2023-06-26
Bluebell migration tweaks

### LINKED ISSUE #1775: Migration tweaks (bluebell)
- no argument should run all places in order (currently nothing happens if a place or work isn't given)
- hide eId mapping at the end by default; add --print-eid-map or whatever as an argument
- list all the problem documents at the end, grouped by issue e.g. validation

goose-life:
closes #1775


longhotsummer (approved):
nice

PR #1780, opened 2023-06-27
replace all spaces with underscores in uploaded attachment filenames

goose-life:


longhotsummer (commented):
This is a good start and helps to reduce the chance of spaces. However, this still doesn't support existing files with spaces in their names. Editors can also change filenames to introduce spaces.

Fundamentally, we do support spaces in filenames - they must simply be escaped to %20 when inserted into bluebell text.

goose-life:
I've opened https://github.com/laws-africa/bluebell-monaco/pull/98

longhotsummer (approved):

longhotsummer (approved):

PR #1782, opened 2023-06-30
Don't let empty links break PDF

goose-life:


goose-life:
[bad-links.pdf](https://github.com/laws-africa/indigo/files/11915573/bad-links.pdf)

@longhotsummer

longhotsummer (approved):

PR #1788, opened 2023-07-14
Tree taxonomy 

actlikewill:
- adds tree taxonomy

image


actlikewill:
@longhotsummer @goose-life Still a few things I'd want to add. Particularly, the toc filter needs a way to mark active or inactive states, which I think needs a custom JS component.

indigo_app/static/javascript/indigo-app.js
goose-life, line 1: you can probably roll this back if you rebase from master as we're now using law-widgets 1.5.0

package-lock.json
goose-life, line 1: same as above

package.json
goose-life, line 1: same as above

indigo_api/serializers.py
```diff
@@ -627,6 +628,15 @@ def get_taxonomies(self, instance):
 
         return taxonomies
 
+    def get_taxonomy_topics(self, instance):
+        taxonomies = []
+        for t in instance.taxonomy_topics.all():
```
longhotsummer, line 633: I think we can just make this a list of slugs

indigo_app/templates/indigo_api/_task_filter_form.html
```diff
@@ -80,5 +80,9 @@
         </div>
       </div>
     {% endif %}
+
+  <!-- Hidden field to capture and preserve the taxonomy_topic parameter-->
+  <input type="hidden" name="taxonomy_topic" value="{{ request.GET.taxonomy_topic }}" />
```
longhotsummer, line 85: I think you can also just do `form.taxonomy_topic.as_hidden`

longhotsummer: this is the more form-oriented approach, since `request.GET` bypasses the form completely.

indigo_app/views/tasks.py
```diff
@@ -90,6 +90,8 @@ def get_context_data(self, **kwargs):
         context['frbr_uri'] = self.request.GET.get('frbr_uri')
         context['task_groups'] = Task.task_columns(self.form.cleaned_data['state'], context['tasks'])
 
+        context["taxonomy_toc"] = json.dumps(TaxonomyTopic.get_toc_tree(self.request.GET))
```
longhotsummer, line 93: in places.py you don't use json.dumps -- are the uses different?

indigo_app/js/components/TaxonomyTOC.vue
```diff
@@ -0,0 +1,44 @@
...
+        anchor.classList.add('active');
+      }
+    });
+  }
+};
+</script>
+
+<style>
+.active, .active:hover {
+  color: tomato;
```
longhotsummer, line 40: In the document view we apply this styling, which users are used to:

```css
  background-color: #2d7ad4;
  color: white;
```

indigo_app/templates/indigo_api/_task_filter_form.html
actlikewill: @longhotsummer We need to set the value from request.GET.

indigo_app/templates/place/works.html
```diff
@@ -22,7 +22,11 @@
     </div>
   {% endblock %}
 
-  <div class="card">
+<div class="row">
+  <div data-vue-component="TaxonomyTOC"></div>
```
longhotsummer, line 26: the direct children of `.row` must always be `.col` otherwise the padding breaks.

indigo_app/templates/place/works.html
```diff
@@ -5,7 +5,7 @@
 {% block view-id %}library-view{% endblock %}
 
 {% block content %}
-<div class="container mt-3 mb-5">
+<div class="container-fluid px-5 mt-3 mb-5">
```
longhotsummer, line 8: Once you've fixed the `.row > .col` thing below, the `px-5` can fall away.

indigo_app/templates/place/works.html
```diff
@@ -22,7 +22,11 @@
     </div>
   {% endblock %}
 
-  <div class="card">
+<div class="row">
+  <div data-vue-component="TaxonomyTOC"></div>
+  {{ taxonomy_toc|json_script:"taxonomy_toc" }}
+
+  <div class="card col-md mx-1 p-0">
```
longhotsummer, line 29: wrap the `.card` in a `.col`, you can't combine them. Don't mess with the `.col`'s padding or margin.

indigo_app/templates/place/works.html
```diff
@@ -153,6 +157,7 @@
                             </div>
                           {% endif %}
                         </div>
+  </div>
```
longhotsummer, line 160: this is breaking things

longhotsummer: you need to move the pagination inside the second `.col`

indigo_app/templates/indigo_api/_task_filter_form.html
longhotsummer: shouldn't the form already have it as a value? The form is created based on `request.GET`

actlikewill: Ah, right. Its working. Thanks

indigo_api/serializers.py
actlikewill: Done. :+1:

indigo_app/static/javascript/indigo-app.js
actlikewill: So I can omit this file altogether?

indigo_app/views/tasks.py
actlikewill: Removed `json.dumps`

indigo_app/js/components/TaxonomyTOC.vue
actlikewill: Changed :heavy_check_mark:

indigo_app/templates/place/works.html
actlikewill: changed :white_check_mark:

indigo_app/templates/place/works.html
actlikewill: changed

indigo_app/templates/place/works.html
actlikewill: changed :white_check_mark:

indigo_app/templates/place/works.html
actlikewill: Fixed :white_check_mark:

indigo_app/templates/place/works.html
actlikewill: It looks much better without messing with native padding or margins

indigo_app/static/javascript/indigo-app.js
goose-life: yeah I reckon!

indigo_api/management/commands/taxonomies.py
goose-life, line 1: cool! I didn't know about `type=argparse.FileType` for arguments.

indigo_app/templates/place/works.html
goose-life, line 1: just checking if you've looked at the works list on a narrower / laptop screen with this added?

actlikewill: @goose-life Yeah, I think it looks okay on narrower screens. We use the col-md breakpoint, so the columns stack on smaller screens.

indigo_api/management/commands/migrate_taxonomies.py
```diff
@@ -0,0 +1,21 @@
...
+                    if vocabulary.level_2:
+                        TaxonomyTopic.add_child(child, name=vocabulary.level_2)
+
+        for work in Work.objects.all().iterator():
+            for taxonomy in work.taxonomies.all():
+                topic = taxonomy.title
+                level_1 = taxonomy.level_1
+                level_2 = taxonomy.level_2
+                topics = TaxonomyTopic.objects.filter(name__in=[topic, level_1, level_2])
+                work.taxonomy_topics.add(*topics)
```
actlikewill, line 21: @longhotsummer, @goose-life  Here is the migration command. I still have the two for loops, it seems easier this way. I have the TaxonomyVocabulary object as the root here where I then loop through all related Vocabulary topic objects to build the tree.

indigo_api/management/commands/migrate_taxonomies.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        for t in TaxonomyVocabulary.objects.all().iterator():
+            root = TaxonomyTopic.add_root(name=t.title)
+            for vocabulary in t.topics.all():
+                if vocabulary.level_1:
+                    child = TaxonomyTopic.add_child(root, name=vocabulary.level_1)
+                    if vocabulary.level_2:
+                        TaxonomyTopic.add_child(child, name=vocabulary.level_2)
```
longhotsummer, line 13: Isn't it safer and simpler to, at this point, use `for work in vocabulary.works.all()` to adjust the works to use the new `child` (either at level 1 or level 2).

actlikewill: @longhotsummer Yeah. Forgot about the reverse relationship. Changed.

indigo_api/management/commands/migrate_taxonomies.py
```diff
@@ -0,0 +1,24 @@
...
+class Command(BaseCommand):
+    def handle(self, *args, **kwargs):
+        for t in TaxonomyVocabulary.objects.all().iterator():
+            root = TaxonomyTopic.add_root(name=t.title)
+            log.info(f"Created root node {root}")
+            for vocabulary in t.topics.all():
+                if vocabulary.level_1:
+                    log.info(f"Creating child node {vocabulary.level_1}")
+                    child_1 = TaxonomyTopic.add_child(root, name=vocabulary.level_1)
+                    for work in vocabulary.works.all():
```
longhotsummer, line 16: I think you're over-thinking this. If `vocabulary` has a `level_2`, then the works are tagged at the second level. Otherwise, they're tagged at the first level.

See here how `level_1` is duplicated when there is a `level_2`:

image

So you only need to apply EITHER `level_2` (if it's not blank) OR fall back to `level_1`. Not both.

indigo_app/static/javascript/indigo-app.js
actlikewill: Removed :+1:

indigo_api/management/commands/taxonomies.py
actlikewill: :grinning:

indigo_api/management/commands/migrate_taxonomies.py
actlikewill: Updated :+1:

longhotsummer (approved):

PR #1797, opened 2023-07-26
fixed psycopg-2 error during deps installation

nicksonlangat:


longhotsummer:
Hi @nicksonlangat thanks for the PR.

We've chosen not to include this dependency because of the [recommendation](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary) that psycopg2-binary is only used for development, and that psycopg2 itself is used in production.

You can see in our [example Dockerfile](https://github.com/laws-africa/indigo/blob/master/Dockerfile#L37C17-L37C25) how we pull it in.

However, I see that the [local installation documentation](https://github.com/laws-africa/indigo/blob/master/docs/running/index.rst#installing-indigo-locally) doesn't include a note about this, so I'll fix that.

longhotsummer:
Ref https://github.com/laws-africa/indigo/pull/1798

PR #1800, opened 2023-07-27
ignore generated app css files

nicksonlangat:
These two files were auto generated and can be accidentally pushed in a commit, is it okay to ignore them? 

indigo_app/static/stylesheets/app.css
indigo_app/static/stylesheets/app.css.map

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ venv
 attachments/
 indigo_app/static/lime/develop
 staticfiles/
+static/stylesheets/app.css
+static/stylesheets/app.css.map
 docs/_build
 node_modules/
 .coverage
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,6 +6,8 @@ venv
 attachments/
 indigo_app/static/lime/develop
 staticfiles/
+static/stylesheets/app.css
+static/stylesheets/app.css.map
```
longhotsummer, line 10: I think these both need to be `indigo_app/static/...` ?

nicksonlangat: Indeed.

longhotsummer (approved):

PR #1805, opened 2023-08-03
Fix active class for taxonomy table

actlikewill:
- fixes the active class on the taxonomy table

### DIFF
--- a/indigo_app/js/components/TaxonomyTOC.vue
+++ b/indigo_app/js/components/TaxonomyTOC.vue
@@ -6,7 +6,7 @@
         title-filter-clear-btn-classes="btn btn-sm btn-secondary"
         title-filter-input-classes="form-field"
         title-filter-placeholder="Filter by topic"
-        :class="active"
+        class="taxonomy-sidebar"
     ></la-table-of-contents-controller>
 </template>
 
@@ -33,10 +33,3 @@ export default {
   }
 };
 </script>
-
-<style>
-.active {
-  background-color: #2d7ad4;
-  color: white;
-}
-</style>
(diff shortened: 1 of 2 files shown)


indigo_app/js/components/TaxonomyTOC.vue
```diff
@@ -27,15 +27,15 @@ export default {
...
       if (params.get('taxonomy_topic') === href.get('taxonomy_topic')) {
-        anchor.classList.add('active');
+        anchor.classList.add('active-taxonomy');
       }
     });
   }
 };
 </script>
 
 <style>
```
longhotsummer, line 37:
```suggestion
<style scoped>
```

longhotsummer: that'll scope this only to this component.

alternatively scope the CSS selector to the sidebar and/or `la-table-of-contents-controller`

longhotsummer: @actlikewill my preferred mechanism would be to add a wrapping `taxonomy-sidebar` class to the whole sidebar, and then move this CSS into the app's actual SCSS files (rather than here), and then scope the `.active` styles to just the sidebar.

actlikewill: @longhotsummer This is not working. I think its because the style added dynamically with JS

actlikewill: @longhotsummer I went with the second suggestion. Thanks

goose-life (approved):

longhotsummer (approved):

PR #1813, opened 2023-08-17
Work commencement description

goose-life:


indigo_api/templatetags/indigo.py
```diff
@@ -91,3 +91,75 @@ def publication_document_description(work, placeholder=False):
         }
     elif placeholder:
         return _('Published')
+
+
+@register.simple_tag
+def work_commencement_description(work, friendly_date=True):
+    """ Describes the commencement status of a work.
+        Returns a tuple containing: (code, descriptive string, [optional commencing work tuple])
```
longhotsummer, line 99: I think it'll make reading and writing this code (and the code that uses the result) much cleaner if you return a dict instead:

```python
{
  "type": "uncommenced" # or "single" or "multiple" -- avoid magic numbers wherever possible :)
  "title": ...
  "note": ...
}

File: indigo_api/templatetags/indigo.py [hunk 2]
```diff
@@ -91,3 +91,75 @@ def publication_document_description(work, placeholder=False):
...
+    """
+    n_commencements = work.commencements.all().count()
+
+    if n_commencements > 1:
+        return 2, _('There are multiple commencements')
+
+    if n_commencements == 0:
+        return 0, _('Not commenced')
+
+    else:
```
longhotsummer, line 117: Minor: you don't need the `else` because the if uses `return`

indigo_api/templatetags/indigo.py
longhotsummer: Would it be useful to have this on the work object itself, rather than as a template tag?

longhotsummer: That would mean you can't pass in the `friendly_date` flag

longhotsummer (approved):
nice, this simplifies a lot of the templates.

Issue #1786, opened 2023-07-11
As an editor / reviewer I don't want to compare an imported consolidation docx against the publication PDF


goose-life:
Editors and reviewers are very used to clicking the 'Show source' button to check the content of a document against. However, when the document is a consolidation they can waste time wondering why the content of the two documents differs. 

Some ideas:
- Show the imported docx as a PDF in the Source panel (on the platform only) when the point in time is a consolidation
    - Default to this while having the publication document available in the dropdown as usual
    - Similarly, if the point in time is an amendment, default to showing the amendment PDF at that date if available rather than the publication document
- Show an alert if viewing something other than the 'correct' source — i.e. the date of the PDF doesn't match the date of the document
    - This might be needed even if we default to showing the right thing, as the right thing won't always be available

Ref https://edit.laws.africa/documents/10941/ 

![Image](https://github.com/laws-africa/indigo/assets/32566441/3dd32684-eb17-491c-b2fe-8e459be939b6)

Ref https://edit.laws.africa/documents/9009/

![Image](https://github.com/laws-africa/indigo/assets/32566441/56d27abb-8860-4fdb-b73a-b6a64e61ddff)

PR #1815, opened 2023-08-18
Show current source document

actlikewill:
- adds a view to convert docx to pdf upon request
- shows docx attachment as source file by default

https://www.loom.com/share/d51ccc6882a742c5ae2bcb8f3b55334d?sid=26e9ecf0-da59-44c2-95fb-fd410f297719

Fixes https://github.com/laws-africa/indigo/issues/1786

### DIFF
--- a/indigo_api/views/attachments.py
+++ b/indigo_api/views/attachments.py
@@ -11,6 +11,16 @@ from ..serializers import AttachmentSerializer
 from ..authz import ModelPermissions, RelatedDocumentPermissions
 from .documents import DocumentResourceView
 from .misc import DEFAULT_PERMS
+from docpipe.soffice import soffice_convert
+import os
+
+DOC_MIMETYPES = [
+    "application/vnd.oasis.opendocument.text",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+    "application/msword",
+    "application/rtf",
+    "text/rtf",
+]
 
 
 def view_attachment(attachment):
@@ -28,6 +38,16 @@ def view_attachment_by_filename(doc_id, filename):
     document = get_object_or_404(qs, deleted__exact=False, id=doc_id)
     attachment = get_list_or_404(Attachment.objects, document=document, filename=filename)[0]
     return view_attachment(attachment)
+
+
+def view_attachment_as_pdf(attachment):
+    suffix = os.path.splitext(attachment.filename)[1].lstrip('.')
+    pdf = soffice_convert(attachment.file, suffix, 'pdf')[0]
+    file_bytes = pdf.read()
+    response = HttpResponse(file_bytes, content_type="application/pdf")
+    response['Content-Disposition'] = 'inline; filename=%s' % attachment.filename
+    response['Content-Length'] = str(len(file_bytes))
+    return response
 
 
 def download_attachment(attachment):
@@ -49,6 +69,8 @@ class AttachmentViewSet(DocumentResourceView, viewsets.ModelViewSet):
     @action(detail=True, methods=['GET'])
     def view(self, request, *args, **kwargs):
         attachment = self.get_object()
+        if attachment.mime_type in DOC_MIMETYPES:
+            return view_attachment_as_pdf(attachment)
         return view_attachment(attachment)
 
     def filter_queryset(self, queryset):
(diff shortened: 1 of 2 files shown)


actlikewill:
@longhotsummer @goose-life We could simplify this to handle everything in the `/view` view. and then in the js we could check for the first pdf attachment and show that by default. Let me know if we should do it like that instead.

goose-life (dismissed):
looks great to me as is, other than those javascript errors, which seem to be just a syntax issue?

goose-life:
but for context, I did see this working earlier — maybe a video would help as part of the PR?

indigo_app/static/javascript/indigo/views/document_source_att.js
```diff
@@ -61,11 +70,18 @@
         choices = this.attachments.filter(function(att) {
           return self.mime_types[att.get('mime_type')];
         }).map(function(att) {
+          let url = att.get('view_url');
+          let default_choice;
+           if (Object.keys(self.docx_mimetypes).includes(att.get('mime_type'))) {
```
longhotsummer, line 75:
```suggestion
          if (Object.keys(self.docx_mimetypes).includes(att.get('mime_type'))) {
```

indigo_api/views/attachments.py
```diff
@@ -51,6 +68,13 @@ def view(self, request, *args, **kwargs):
         attachment = self.get_object()
         return view_attachment(attachment)
```
longhotsummer, line 69: What about moving the PDF logic here? The idea being, if the document isn't already a PDF, convert it to a pdf. The UI only uses `../view` to show the file, so converting to a PDF is safe. The `../download` is used to actually download it.

It would mean the UI can always ask for `.../view` and know that it'll get back something the browser can show (ie. a pdf)

longhotsummer: It should mean that the bulk of the JS changes below fall away.

indigo_api/views/attachments.py
```diff
@@ -51,6 +68,13 @@ def view(self, request, *args, **kwargs):
...
 
+    @action(detail=True, methods=['GET'])
+    def as_pdf(self, request, *args, **kwargs):
+        attachment = self.get_object()
+        if attachment.mime_type in DOCX_MIME_TYPES:
```
longhotsummer, line 74: `soffice` is pretty good at converting almost anything to a PDF. So perhaps rather check if it's NOT a pdf, and then try to convert it.

indigo_app/static/javascript/indigo/views/document_source_att.js
```diff
@@ -23,7 +23,19 @@
...
         this.$toggle = this.$('.source-attachment-toggle');
         this.iframe = document.getElementById('source-attachment-iframe');
-        this.mime_types = {'application/pdf': true, 'text/html': true};
+        this.docx_mimetypes = {
+          'application/vnd.openxmlformats-officedocument.wordprocessingml.document': true,
+          'application/msword': true,
+          'application/vnd.oasis.opendocument.text': true,
+          'application/rtf' : true,
+          'text/rtf': true,
+        };
```
actlikewill, line 32: @longhotsummer I'm happy to have this check here because I think the logic is more clear what we are doing. i.e we want the first document that is of these mimetypes only and nothing else

longhotsummer: cool

indigo_api/views/attachments.py
```diff
@@ -49,6 +69,8 @@ def download(self, request, *args, **kwargs):
     @action(detail=True, methods=['GET'])
     def view(self, request, *args, **kwargs):
         attachment = self.get_object()
+        if attachment.mime_type in DOC_MIMETYPES:
+            return view_attachment_as_pdf(attachment)
```
actlikewill, line 73: @longhotsummer Similar to the reasoning in the js, I think this logic describes our goals a bit better i.e  we are interested in converting docx types only rather that putting everything through soffice.

indigo_api/views/attachments.py
```diff
@@ -11,6 +11,16 @@
...
+from docpipe.soffice import soffice_convert
+import os
+
+DOC_MIMETYPES = [
+    "application/vnd.oasis.opendocument.text",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+    "application/msword",
+    "application/rtf",
+    "text/rtf",
+]
```
actlikewill, line 23: @longhotsummer maybe this constant deserves a spot in docpipe itself?

actlikewill:
@goose-life Here is a loom for this: https://www.loom.com/share/d51ccc6882a742c5ae2bcb8f3b55334d?sid=26e9ecf0-da59-44c2-95fb-fd410f297719

longhotsummer (approved):
super

goose-life (approved):

PR #1821, opened 2023-08-29
Timeline

goose-life:


indigo_app/views/works.py
```diff
@@ -71,47 +71,11 @@ def get_context_data(self, **kwargs):
...
-                event['amendments'] = event_amendments
-            if event.get('consolidation'):
-                event['consolidations'] = work.arbitrary_expression_dates.filter(date=date).all()
-            event['expressions'] = work.expressions().filter(expression_date=date).all()
-
-        return dates
+        timeline = work.get_timeline()
+        # add expressions
+        for entry in timeline:
+            entry.expressions = work.expressions().filter(expression_date=entry.date).all()
```
longhotsummer, line 77: This is actually pretty inefficient. Better to load all the expressions once, index by date, and then filter them here.

```python
expressions = list(work.expressions().all())
...
entry.expressions = [e for in in expressions if e.expression_date == e.date]
```

indigo_app/templates/indigo_api/_work_points_in_time.html
```diff
@@ -30,157 +30,53 @@
...
-              <h6>{% trans 'Consolidation' %}</h6>
-              <div class="form-inline edit-wrapper mb-3 collapse" id="arbitrary-expression-date-form-{{ consolidation.id }}">
-                <input type="text" class="form-control amendment-date mr-2" data-provide="datepicker" placeholder="yyyy-mm-dd" pattern="\d{4}-\d\d-\d\d" required name="date" value="{{ consolidation.date|date:"Y-m-d" }}" autocomplete="off">
-                <button class="btn btn-success" type="submit">{% trans 'Save' %}</button>
-                <a href="#arbitrary-expression-date-form-{{ consolidation.id }}" class="btn btn-link" data-toggle="collapse">{% trans 'Cancel' %}</a>
-              </div>
-            </form>
+        {% for event in entry.events %}
+          {% if editable and event.related and event.type == 'amendment' %}
+            {% include 'indigo_api/_edit_amendment.html' %}
```
longhotsummer, line 39: Perhaps move these included templates under `indigo_api/timeline/` so that it's clear they're scoped to the timeline. We haven't done that much in indigo, and it means the `indigo_api` folder has a huge number of templates.

indigo_app/templates/indigo_api/_single_description.html
longhotsummer, line 1: `description` of what? perhaps `_timeline_event.html`?

indigo_api/models/works.py
```diff
@@ -400,97 +400,41 @@ def has_consolidation(self):
...
         return self.commencement_description(friendly_date=False)
 
     def commencement_description_external(self):
         return self.commencement_description()
 
+    def get_timeline(self):
+        from indigo_api.models.timeline import get_timeline
+        return get_timeline(self)
+
+    def get_timeline_as_dicts(self):
```
longhotsummer, line 434: Minor: since this is serializing them, normally this would be done by the place that cares about serializing, since this class doesn't really.

indigo_api/models/timeline.py
```diff
@@ -0,0 +1,174 @@
+from django.utils.formats import date_format
+from django.utils.translation import gettext as _
+
+
+class TimelineEntry:
```
longhotsummer, line 5: Python`s dataclasses are really useful for something like this. It will handle defaults and setup a constructor that takes the various fields as parameters.

See https://docs.python.org/3/library/dataclasses.html and https://github.com/laws-africa/indigo-lawsafrica/blob/master/citator/extraction/patterns.py#L18

indigo_api/models/timeline.py
```diff
@@ -0,0 +1,174 @@
...
+    events = None
+
+    def __init__(self, **kwargs):
+        self.initial = False
+        self.events = []
+
+        for k, v in kwargs.items():
+            setattr(self, k, v)
+
+    def as_dict(self):
```
longhotsummer, line 17: dataclasses's `asdict()` will do all this for you!

indigo_api/models/timeline.py
longhotsummer, line 1: if you put this into `indigo_api/timeline.py` then you may be able to avoid the circular imports stuff and having to do local imports. But it's not a huge issue.

indigo_api/models/timeline.py
```diff
@@ -0,0 +1,174 @@
...
+    # the initial date is the publication date, or the earliest of the consolidation and commencement dates, or None
+    initial = work.publication_date
+    if not initial and any(commencement_dates + consolidation_dates):
+        initial = min(d for d in commencement_dates + consolidation_dates if d)
+
+    for date in all_dates:
+        entry = TimelineEntry(date=date, initial=date == initial)
+        amendments = [a for a in all_amendments if a.date == date]
+        if len(amendments) > 1:
+            from indigo_api.models import Amendment
```
longhotsummer, line 122: if you're doing local imports, prefer to do them all at the start of the function

longhotsummer (approved):

Issue #1823, opened 2023-08-30
Indigo does not support all unicode


longhotsummer:
Indigo breaks in a few places when trying to work with XML that has unicode chars, like emojis. This is because we don't store our XML with an `<?xml encoding="utf-8">` heading, and we try to parse it as a string, rather than as bytes.

We need to do one or the other to fully support unicode.

longhotsummer:
See also https://github.com/laws-africa/cobalt/issues/76

longhotsummer:
Note #1829  and #1830 otherwise this is complete.

PR #1824, opened 2023-08-30
parse bytes, not strings

longhotsummer:
Requires https://github.com/laws-africa/cobalt/pull/77 and https://github.com/laws-africa/bluebell/pull/118


longhotsummer:
Fixes #1823

longhotsummer:
Ready for review, I'll check why tests are failing. It looks like slight formatting differences with XSL output which is curious.

goose-life (approved):

PR #1837, opened 2023-09-22
move hierarchicalize_stages out

goose-life:
This makes it a bit easier to manipulate the base stages, e.g. if we want to run an additional step before or after all the others

### DIFF
--- a/indigo/pipelines/hier.py
+++ b/indigo/pipelines/hier.py
@@ -1056,7 +1056,7 @@ class IdentifyContainerHeadings(Stage):
                     block.remove(first)
 
 
-hierarchicalize = Pipeline([
+hierarchicalize_stages = [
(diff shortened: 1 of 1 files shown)


indigo/pipelines/hier.py
```diff
@@ -1099,4 +1099,6 @@ def __call__(self, context):
 
     # identify container headings that were missed previously once the structure is set
     IdentifyContainerHeadings(),
-], name="Hierarchicalize")
+]
+
+hierarchicalize = Pipeline(hierarchicalize_stages, name="Hierarchicalize")
```
longhotsummer, line 1104: An alternative is to read `hierarchicalize.stages`

goose-life:
we're going with the suggestion above

PR #1844, opened 2023-10-05
Swahili

goose-life:
Include Swahili translations in HTML

image


indigo_api/tests/importer_fixtures/tz-cap6.txt
```diff
@@ -2002,9 +2000,7 @@ CHAPTER IV - PRODUCTION AND EFFECT OF EVIDENCE (ss 110-126)
 
 CHAPTER V - WITNESSES (ss 127-177)
 
-  PART I
-
-    COMPETENCY, COMPELLABILITY AND PRIVILEGE OF WITNESSES (ss 127-143)
+  PART I - COMPETENCY, COMPELLABILITY AND PRIVILEGE OF WITNESSES (ss 127-143)
```
longhotsummer, line 2003: if extending the length limit gets too much, another good signal for a heading is that the first 30 chars, say, are all uppercase

goose-life: this was actually just a by-product, I noticed that we use 80 in one place and 60 in another, so I went with the longer one for both

PR #1849, opened 2023-10-12
PDF translation support, and write XML using buffer

goose-life:


indigo_api/exporters.py
```diff
@@ -205,7 +210,11 @@ def insert_frontmatter(self, document):
         """ Generates the frontmatter and inserts it before the body (and coverpage if present) in the XML.
             The document's document_xml isn't updated.
         """
+        # explicitly set the document's language for e.g. numbered title to be translated
+        activate(document.django_language)
```
longhotsummer, line 214: rather user `override()` which ensures the deactivate is called if there is an error:

```python
with override(document.django_language):
  frontmatter = ...
```

indigo_api/exporters.py
```diff
@@ -172,10 +174,13 @@ def render(self, document, element=None):
...
             self.resize_tables(document.doc)
             # write the XML to file
-            xml = etree.tostring(document.doc.main, encoding='unicode')
+            xml_string = etree.tostring(document.doc.main, encoding='unicode')
+            xml_bytes = xml_string.encode('utf-8')
             xml_file = os.path.join(tmpdir, 'raw.xml')
-            with open(xml_file, "wb") as f:
-                f.write(xml.encode('utf-8'))
+            # use io.open, BytesIO, getbuffer to avoid XML parse error in production
+            with io.open(xml_file, "wb") as f:
```
longhotsummer, line 181: can you link to the issue this is trying to solve? I don't see how this change will fix anything.

longhotsummer: Ah https://github.com/laws-africa/indigo-lawsafrica/issues/1012

indigo_api/exporters.py
goose-life: thanks!

longhotsummer (approved):

PR #1853, opened 2023-11-01
Refs resolver supports text documents

longhotsummer:
Provision refs finders now handle text matches, too.

### DIFF
--- a/indigo/analysis/refs/provisions.py
+++ b/indigo/analysis/refs/provisions.py
@@ -442,7 +442,7 @@ class ProvisionRefsMatcher(CitationMatcher):
                             # prefix
                             match.string[max(offset + ref.start_pos - self.text_prefix_length, 0):offset + ref.start_pos],
                             # suffix
-                            match.string[offset + ref.end_pos:min(offset + ref.end_pos + self.text_suffix_length, len(match.string))],
+                            match.string[offset + last_ref.end_pos:min(offset + last_ref.end_pos + self.text_suffix_length, len(match.string))],
                         )
                     )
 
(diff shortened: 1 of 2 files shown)


longhotsummer:
This is required for the AfricanLII Citator.

indigo/analysis/refs/provisions.py
```diff
@@ -441,13 +506,46 @@ def candidates():
...
 
         return None, None
 
-    def resolve_target(self, node: Element, frbr_uri: str) -> Tuple[Union[str, None], Union[Element, None]]:
+    def get_target_from_text(self, match: re.Match, target: Union[str, None]) -> (Union[str, None], Union[Element, None]):
+        """Determine and resolve a remote target and return the target_frbr_uri and the appropriate root XML element of
+        the document to use to resolve the references. References to the local document are not supported because
+        the local document is assumed to be plain text.
+        """
+        if not target or target == "this":
```
goose-life, line 520: do these – 'this', 'thereof', 'of' – mean text ref extraction only works on English documents?

goose-life (approved):

indigo/analysis/refs/provisions.py
longhotsummer: No, they're magic constants that the grammar returns to indicate particular targets types. The grammar's alternative languages can use whatever they need to detect the various conditions, and then return these magic constants.

goose-life: oh amazing! I think I was thrown by them being strings, but that makes sense and is very cool!

PR #1857, opened 2023-11-02
PDF static content

goose-life:
This adds support for making running headers and footers translatable and easier to override.

### DIFF
--- a/indigo_api/exporters.py
+++ b/indigo_api/exporters.py
@@ -232,7 +232,7 @@ class PDFExporter(HTMLExporter, LocaleBasedMatcher):
         locality = document.work.locality.name if document.work.locality else None
         country = document.work.country.name if document.country not in self.dont_include_countries else None
         if locality and country:
-            return _('%(locality), %(country)s') % {'locality': _(locality), 'country': _(country)}
+            return f'{_(locality)}, {_(country)}'
         elif locality:
             return _(locality)
         elif country:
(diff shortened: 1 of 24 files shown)


indigo_api/templates/indigo_api/akn/export/pdf_static_content_akn.xml
longhotsummer, line 1: Does this need to be `.xml.html` for makemessages to work?

goose-life: just adding `-e xml` works — let me run that and add it to this PR

indigo_api/exporters.py
```diff
@@ -202,6 +203,41 @@ def html_to_xml(self, html_string):
...
+            'place_string': self.get_place_string(document),
+        }
+
+    def get_place_string(self, document):
+        """ Get the appropriate string based on the locality and country, in the right language.
+        """
+        locality = document.work.locality.name if document.work.locality else None
+        country = document.work.country.name if document.country not in self.dont_include_countries else None
+        if locality and country:
+            return _('%(locality), %(country)s') % {'locality': _(locality), 'country': _(country)}
```
longhotsummer, line 235: I'm not sure of the value of `(%s, %s)` being translateable, but it won't hurt.

goose-life: eh yeah this was silly, I got rid of it 😅

longhotsummer (approved):

PR #1862, opened 2023-11-06
Add topics to bulk creator

goose-life:
closes https://github.com/laws-africa/indigo-lawsafrica/issues/1036

This can all be merged at once, but we'll need to run the management command immediately upon deploying

image
image


indigo/bulk_creator.py
```diff
@@ -222,6 +222,7 @@ class BaseBulkCreator(LocaleBasedMatcher):
     consolidation_date = None
 
     workflow = None
+    project_topic = None
```
longhotsummer, line 225: The model being added is a taxonomy_topic -- let's be consistent internally with what we call things, even if the user-visible label is a bit different.

longhotsummer (changes requested):
Using a consistent internal name for the topics is important. We use them for projects, but they're broader than that.

indigo/bulk_creator.py
goose-life: I used this because we're only using a subset of taxonomy_topics here (the ones that fall under 'projects'); but I'm happy to change it 👍

longhotsummer (approved):

actlikewill (approved):
looks good

PR #1869, opened 2023-11-10
Work list v2

longhotsummer:
New work view

https://www.loom.com/share/3766cd19b4394e4b82c714468ee6f000

* [x] make sorting work
* [x] finalise all the details shown in the work card
* [x] finalise the work facets on the left
* [x] add taxonomy on the left
* [x] finalise expression facets on the left (eg. the stub) and perhaps move them under a "Point in time" section so that they're separate from the work facets
* [x] wire up all the tabs shown in the work cards
* [x] remove old /works view completely
* [x] ensure download as XLSX works
* [x] update advanced filter to only include ranges (or leave out for now and reinsert later)

Bulk editing: I'm tempted to remove the bulk editing support for this release and build on it later.


actlikewill:
@longhotsummer 
- about making sort work, could you elaborate, are we adding any fields ?, should it work just by selecting a value without clicking apply?
- what do we mean by adding a poin in time section in the facets?

longhotsummer:
> @longhotsummer
> 
> * about making sort work, could you elaborate, are we adding any fields ?, should it work just by selecting a value without clicking apply?

To make the sort submit on change, try this:

```diff
-  hx-trigger="submit, change from:#work-list target:.work-list-facet"
+  hx-trigger="submit, change from:#work-list target:.work-list-facet, change from:[name=sortby]"
   hx-indicator="#work-list"
```

You and Edith need to decide if new sorting fields are necessary, if some need to be removed, etc.

> * what do we mean by adding a point in time section in the facets?

All the current facets are work-based, except the "Expression stubs". To differentiate between the work vs document facets, I think the latter should be split out into a different `list-group` element.

goose-life:
@longhotsummer I think we can leave in the bulk editing of taxonomy topics? it'll be super useful just by itself.

indigo_app/templates/indigo_app/place/_works_actions.html
goose-life, line 1: can we add a checkbox back in for these?

indigo_app/forms.py
```diff
@@ -408,90 +403,115 @@ def is_set(a):
...
+                amendment_qs |= Q(amendments__date__isnull=True)
+
+            queryset = queryset.filter(amendment_qs)
+        
+        if self.cleaned_data.get('amendment_date_start') and self.cleaned_data.get('amendment_date_end'):
+            start_date = self.cleaned_data['amendment_date_start']
+            end_date = self.cleaned_data['amendment_date_end']
+            queryset = queryset.filter(amendments__date__range=[start_date, end_date]).order_by('-amendments__date')
 
         # filter by work completeness
```
goose-life, line 516: this can come out?

indigo_app/forms.py
```diff
@@ -500,27 +520,66 @@ def filter_queryset(self, queryset):
...
+                documents_qs |= Q(id__in=one_document_ids)
+            if 'multiple' in documents_filter:
+                multiple_document_ids = queryset.annotate(Count('document')).filter(document__count__gt=1).values_list('pk', flat=True)
+                documents_qs |= Q(id__in=multiple_document_ids)
+            if 'none' in documents_filter:
+                documents_qs |= Q(document__isnull=True)
+            if 'published' in documents_filter:
+                documents_qs |= Q(document__draft=False)
+            if 'draft' in documents_filter:
+                documents_qs |= Q(document__draft=True)
```
goose-life, line 567: published and draft are done separately under status, so these will never be in?

indigo_app/forms.py
goose-life, line 516: this block can come out?

goose-life:
- [ ] flip order of documents
- [ ] documents --> point in time
- [ ] multiple documents facet

longhotsummer:
Closes #1867

goose-life (approved):

actlikewill (approved):

PR #1870, opened 2023-11-13
Work list overview tab

goose-life:
- [x] Add overview
- [x] Add Repeals
- [x] Add Subleg / Primary work
- [x] Finalise above-the-fold information
- [x] Add tasks
- [x] Add consolidations
- [x] Include all commencements and their notes
- [ ] Add icons?
- [ ] Fix data-toggle to point down when not collapsed?
- [ ] Make the whole card collapse / clickable, rather than just the little arrow? or give it plenty of padding for easier clicking

https://www.loom.com/share/46226a141dbc41699e1dab8e043d3fee?sid=141f83d3-9aef-44ab-8eeb-96d061e5cdb7

PR #1874, opened 2023-11-16
Details and sorting

goose-life:
https://github.com/laws-africa/indigo/pull/1870 plus sorting and a fix for missing publication documents, and a couple other things


package.json
```diff
@@ -53,5 +53,8 @@
   "bugs": {
     "url": "https://github.com/laws-africa/indigo/issues"
   },
-  "homepage": "https://github.com/laws-africa/indigo#readme"
+  "homepage": "https://github.com/laws-africa/indigo#readme",
+  "dependencies": {
+    "htmx": "^0.0.2"
+  }
```
longhotsummer, line 59:
```suggestion
  "homepage": "https://github.com/laws-africa/indigo#readme"
```

longhotsummer: htmx.org is the real dependency and is above

goose-life: ah thanks, I reverted all this

longhotsummer (approved):

PR #1877, opened 2023-11-21
make subtypes multichoice, make 'Acts only' work

goose-life:
image

### DIFF
--- a/indigo_app/views/places.py
+++ b/indigo_app/views/places.py
@@ -833,13 +833,15 @@ class PlaceWorksFacetsView(PlaceViewBase, TemplateView):
                 st.name,
                 st.abbreviation,
                 counts.get(st.abbreviation, 0),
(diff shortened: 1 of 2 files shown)


indigo_app/views/places.py
```diff
@@ -833,13 +833,15 @@ def facet_subtype(self, facets, qs):
                 st.name,
                 st.abbreviation,
                 counts.get(st.abbreviation, 0),
-                self.form.cleaned_data.get("subtype") == st.abbreviation
+                st.abbreviation in self.form.cleaned_data.get("subtype")
```
actlikewill, line 836:
```suggestion
                st.abbreviation in self.form.cleaned_data.get("subtype", [])
```

to avoid errors in case of no value

actlikewill (approved):
looks good,

longhotsummer (approved):

PR #1879, opened 2023-11-22
include all doctypes in subtype filter

goose-life:
This takes EXTRA_DOCTYPES per country into account

image

image


actlikewill (approved):

longhotsummer (approved):
I want to change the model to always explicitly link to a FK doctype and a FK subtype, and to also store the textual frbr version of each. That's what we do in peachjam and it makes consistency and filtering far simpler, and would make this much simpler too.

PR #1892, opened 2023-11-29
Publication document with htmx

actlikewill:


indigo_app/templates/indigo_api/work_form.html
```diff
@@ -452,18 +431,20 @@ <h5 class="card-header">{% trans 'Country and locality (advanced)' %}</h5>
...
               {% endfor %}
             </select>
           </div>
 
-          <div class="form-group col-4">
-            <label for="{{ form.locality.id_for_label }}">{{ form.locality.label }}</label>
-            <select id="{{ form.locality.id_for_label }}" class="form-control" name="{{ form.locality.html_name }}">
-            </select>
-          </div>
+        {% include 'indigo_api/_work_locality_options.html' %}
```
longhotsummer, line 447: Minor: the property being changed is `locality`, not just the options. I suggest naming the template `_work_locality_form.html` -- that's the pattern used elsewhere for the form.

indigo_app/urls.py
```diff
@@ -92,6 +92,9 @@
     re_path(r'^works(?P<frbr_uri>/\S+?)/edit/$', works.EditWorkView.as_view(), name='work_edit'),
     re_path(r'^works(?P<frbr_uri>/\S+?)/edit/repeal$', works.EditWorkRepealView.as_view(), name='work_edit_repeal'),
     re_path(r'^works(?P<frbr_uri>/\S+?)/edit/parent$', works.EditWorkParentView.as_view(), name='work_edit_parent'),
+    re_path(r'^works(?P<frbr_uri>/\S+?)/edit/find-publication$', works.FindPublicationDocumentView.as_view(), name='find_publication_document'),
+    re_path(r'^works(?P<frbr_uri>/\S+?)/edit/attach-publication$', works.WorkFormPublicationDocumentView.as_view(), name='attach_publication_document'),
+    re_path(r'^works(?P<frbr_uri>/\S+?)/edit/localities-options$', works.WorkFormLocalityOptionsView.as_view(), name='locality_options'),
```
longhotsummer, line 97: The pattern used elsewhere is that the URL is the name of the field, such as `parent`, `repeal` -- in this case it would be `locality`. The options is neither here nor there.

indigo_app/views/works.py
```diff
@@ -1065,3 +1066,101 @@ def get_context_data(self, **kwargs):
...
+                return finder.find_publications(params)
+            except (ValueError, ConnectionError):
+                return []
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        form = self.Form(self.request.POST, prefix="work", instance=self.work)
+
+        if form.is_valid():
+            publication_date = form.cleaned_data['publication_date']
```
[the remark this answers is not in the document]
actlikewill: moved this to the form,

indigo_app/templates/indigo_api/work_form.html
actlikewill: changed :+1:

indigo_app/urls.py
actlikewill: @longhotsummer I've moved these views to be under the place view. This allows them to work while in whether adding a new work or editing an existing one. In places where we need to reference the work I've added a get parameter with the frbr uri,

actlikewill:
I'm opening another pr for to get the work chooser to work on new documents and possibly fixing the failing tests.

goose-life:
@actlikewill this all looks great! the only change I would want is for the possible documents to be loaded before that section of the form is edited, i.e. on load. so if the details were correct but the gazette was missing on creation and added later, if I edit the work I want to see the options without having to change the (correct) data.

indigo_app/templates/indigo_api/work_form.html
```diff
@@ -201,14 +201,25 @@ <h5 class="card-header">{% trans 'Publication' %}</h5>
           {% block publication-date %}
           <div class="form-group col">
             <label for="{{ form.publication_date.id_for_label }}" class="{% if not work.pk and not publication_date_optional %}required{% endif %}">{% trans 'Publication date' %}</label>
-            <input type="text" class="form-control" data-provide="datepicker" id="{{ form.publication_date.id_for_label }}" placeholder="yyyy-mm-dd" {% if not work.pk and not publication_date_optional %}required{% endif %} pattern="\d{4}-\d\d-\d\d" name="{{ form.publication_date.html_name }}" value="{{ form.publication_date.value|date:"Y-m-d"|default:'' }}">
+            <input
+                hx-target=".possible-documents-wrapper"
+                hx-swap="outerHTML"
+                onchange="htmx.trigger(this, 'change');"
+                hx-trigger="change, keyup delay:500ms changed"
```
goose-life, line 208: can we just add `load` to these?

actlikewill: let me see

longhotsummer:
We are very close!

* attaching a document doesn't update correctly when creating a new work
* repealed_date isn't filled in automatically for some reason
* the formset submission for attached documents can fall away and be simplified

actlikewill:
I see we moved the logic to the backend, but still the error occurs.
Setting the frbr_uri field to blank=True fixes the issue. It seems field level validations are run before the form clean method

indigo/settings.py
```diff
@@ -149,7 +149,7 @@
 
     # AKN document types the platform supports, as a list of (name, code) tuples
     # see http://docs.oasis-open.org/legaldocml/akn-core/v1.0/os/part1-vocabulary/akn-core-v1.0-os-part1-vocabulary.html#_Toc523925025
-    'DOCTYPES': [('Act', 'act')],
+    'DOCTYPES': [('Act', 'act'), ('Statement', 'statement')],
```
longhotsummer, line 152: otherwise the tests fail since we expect to be able to create statements. @goose-life I don't know if you want us to keep this in, or if we should adjust the tests for this test case.

longhotsummer (approved):

longhotsummer:
LGTM

PR #1893, opened 2023-11-30
Work actions

goose-life:
TODO:

- [ ] Add 'Select all' option

https://www.loom.com/share/21c7d1b2779d43af92534148ec20217b?sid=69848ebd-6c30-429e-a928-56e8e22ad894


longhotsummer:
I approached the top box as "add" -- so that we didn't have to take into account what was already a part of some/all of them, which simplifies it.

I think ultilmately the best experience is a tree view with checkboxes. Which we need for faceting, too. If we can get the selectpicker working then it's the same experience between the work edit form and this one. Then we can change both later.

Positioning of the checkboxes look great.

goose-life:
https://www.loom.com/share/50b111f4582644f5b134c8b2df4313c2?sid=0073a548-8357-4f74-a278-89d847955a4a

longhotsummer (approved):

PR #1904, opened 2023-12-06
Work in progress

goose-life:
https://www.loom.com/share/03479624a741430aa60298f8e7101626?sid=f9d7c502-4e07-4b20-a912-88a82b56f2b4

I probably need to update fixtures for the tests to pass

To do: add to work form once that's finalised


indigo_app/forms.py
```diff
@@ -767,12 +779,14 @@ class WorkBulkActionsForm(forms.Form):
     save = forms.BooleanField()
     all_work_pks = forms.CharField(required=False)
     works = forms.ModelMultipleChoiceField(queryset=Work.objects, required=True)
+    user = forms.ModelChoiceField(queryset=User.objects)
```
longhotsummer, line 782: huh? just use request.user? otherwise I can impersonate you by submitting this with a different pk

indigo_app/views/works.py
```diff
@@ -216,6 +216,7 @@ def get_context_data(self, **kwargs):
     def form_valid(self, form):
         form.instance.updated_by_user = self.request.user
         form.instance.created_by_user = self.request.user
+        form.instance.work_in_progress = True
```
longhotsummer, line 219: I wonder if we should default this to True (after backfilling with False), so that new works are WIP by default? That means we have to explicitly mark them as approved which I think is a bit safer?

indigo_app/forms.py
goose-life: I guess we can just pass the user when we call .save_form() — I'll do that instead

goose-life:
work form styling: 
https://www.loom.com/share/a3012c40f694469a99389476023b65f5?sid=a17b0b71-9503-4d37-8e3e-8fa4d1bec6a8

longhotsummer:
> work form styling: https://www.loom.com/share/a3012c40f694469a99389476023b65f5?sid=a17b0b71-9503-4d37-8e3e-8fa4d1bec6a8

Love it!

indigo_api/models/works.py
```diff
@@ -638,6 +648,29 @@ def versions(self):
...
         return self.as_at_date_override or self.place.settings.as_at_date
 
+    def approve(self, user, request=None):
+        try:
+            with transaction.atomic():
+                self.work_in_progress = False
+                self.approved_by_user = user
+                self.approved_at = datetime.now()
+                self.save_with_revision(user)
+                work_approved.send(sender=self, request=request)
```
longhotsummer, line 658: the sender is (weirdly) the class, and instance is self

```suggestion
                work_approved.send(sender=self.__class__, instance=self, request=request)
```

indigo_api/models/works.py
```diff
@@ -638,6 +648,29 @@ def versions(self):
...
+        except IntegrityError:
+            pass
+
+    def unapprove(self, user):
+        self.work_in_progress = True
+        self.approved_by_user = None
+        self.approved_at = None
+        self.save_with_revision(user)
+        work_unapproved.send(sender=self)
```
longhotsummer, line 667:
```suggestion
        work_unapproved.send(sender=self.__class__, instance=self)
```

indigo_api/models/works.py
longhotsummer: I don't think we should pass in the request since we won't always have it.

longhotsummer (approved):

longhotsummer:
This is super-solid, thanks Edith.

It looks like we can't approve a work that can't be signed off. I think we should be able to. So the "can't signoff because no expression date" should simply not sign-off that work, but allow the approval to succeed?

That simplifies the error-messaging too.

longhotsummer (approved):

PR #1907, opened 2023-12-09
Place works list improvements

longhotsummer:
* don't prefetch documents any more, they're no longer used - big speed improvement!
* show the basics of the expanded panel while it is loading, so the experience isn't quite as jerky

image


goose-life (approved):
I think filter_document_queryset on the WorkFilterForm can probably fall away too — last time I checked this was its only use

PR #1912, opened 2023-12-12
add FindPossibleDuplicatesView and friends

goose-life:
https://www.loom.com/share/38d9ff2505664a56a3397597a33b5e44?sid=9a804d99-257d-4cae-bc3b-ded92456d496


indigo_app/templates/indigo_api/work_form.html
```diff
@@ -80,28 +89,40 @@ <h5 class="card-header">{% trans 'Work identification' %}</h5>
...
 
         <div class="form-row">
           <div class="form-group col-md-3 mb-0">
             <label for="{{ form.frbr_date.id_for_label }}" class="required">{% trans 'Year (or date) of introduction' %}</label>
-            <input type="text" class="form-control" id="{{ form.frbr_date.id_for_label }}"
-                   required pattern="\d{4}(-\d{2}-\d{2})?" placeholder="yyyy(-mm-dd)"
-                   name="{{ form.frbr_date.html_name }}"
-                   value="{{ form.frbr_date.value|default:'' }}">
+            <input
+                hx-trigger="change, keyup delay:500ms changed"
```
longhotsummer, line 107: shouldn't all of these triggers be the same?

indigo_app/views/works.py
```diff
@@ -1090,6 +1090,49 @@ def get_context_data(self, **kwargs):
...
+        return self.get(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        form = self.Form(self.request.POST)
+
+        if form.is_valid():
+            actual_duplicate = form.find_actual_duplicate(self.country, self.locality)
+            if actual_duplicate:
+                context["actual_duplicate"] = actual_duplicate
```
longhotsummer, line 1127: why the if? just put it into the context directly?

indigo_app/templates/indigo_api/work_form.html
goose-life: some are input and some are select — each type should all be the same. I've only made the title one post on load though, then it just gets done once when the work form loads and each time one of the fields change.

longhotsummer (approved):
nice

PR #1913, opened 2023-12-12
Edit works modal

actlikewill:
- adds edit work modal


actlikewill:
We replaced this with sidebar off canvas edit form #1929

PR #1935, opened 2024-01-17
Htmx toasts

actlikewill:
- support for htmx toasts


indigo_app/middlware.py
```diff
@@ -0,0 +1,31 @@
...
+        # Ignore redirections because HTMX cannot read the headers
+        if 300 <= response.status_code < 400:
+            return response
+
+        # check for messages
+        messages = get_messages(request)
+        if not messages:
+            return response
+
+        response.write(
```
longhotsummer, line 28: what's this doing? appending to the end of the response? what if it's not an HTML response?

actlikewill: hmm, hadn't thought of that, let me change it up. There is another approach that uses HX-Trigger header that may be more suitable.

longhotsummer: another option is, for now, just to add this logic into the one template that needs it -- the work update form. then we make it generic when we need to do so. that will unblock here and let you move forward.

actlikewill: I've updated this to use the HX-Trigger header, I had tested this before changing it out for the previous approach. I thought injecting the toast into the response was simpler because you don't have any javascript.  I suppose I hadn't thought it through completely. This approach is more suitable for our situation I think.

https://www.loom.com/share/185f00a5635945b788d5a99ba7296c4b?sid=a43373f2-a01f-4de6-be16-17df30409dbc

longhotsummer (approved):

PR #1945, opened 2024-01-23
return queryset.distinct() when filtering

goose-life:
https://www.loom.com/share/93d64cd95e294e45833eed60577a96e7?sid=21e36b5c-151a-4ec0-80cc-9f6d0c64c183

### DIFF
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -679,7 +679,7 @@ class WorkFilterForm(forms.Form):
         if self.cleaned_data.get('sortby'):
             queryset = queryset.order_by(self.cleaned_data.get('sortby'))
 
-        return queryset
+        return queryset.distinct()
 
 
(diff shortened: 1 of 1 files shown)


indigo_app/forms.py
```diff
@@ -679,7 +679,7 @@ def filter_queryset(self, queryset, exclude=None):
         if self.cleaned_data.get('sortby'):
             queryset = queryset.order_by(self.cleaned_data.get('sortby'))
 
-        return queryset
+        return queryset.distinct()
```
longhotsummer, line 682: Try `.distinct("pk")` so that it only does the distnict on the primary key, and not on everything. I haven't tested.

goose-life: it didn't like that 😅

longhotsummer: never mind

longhotsummer (approved):

PR #1955, opened 2024-01-31
Amendments in work form

actlikewill:
- adds amendments to work form

https://www.loom.com/share/92f82049cb3f4f14a90ff286aa78269d?sid=8006318f-18fa-42b1-890a-7ee744f6d750

### DIFF
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -987,6 +987,11 @@ class RepealMadeForm(forms.Form):
             work.save()
 
 
+RepealMadeBaseFormSet = formset_factory(
+    RepealMadeForm,
+    extra=0,
+    can_delete=True,
+)
 
 
 class AmendmentForm(forms.Form):
@@ -994,7 +999,6 @@ class AmendmentForm(forms.Form):
     date = forms.DateField(widget=forms.DateInput(attrs={'type': 'date'}), required=True)
     updated_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
     created_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
-
 
     def __init__(self, work, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -108,6 +108,7 @@ class WorkForm(forms.ModelForm):
                 "initial": [{
                     "repealed_work": repealed_work,
                     "repealed_date": repealed_work.repealed_date,
+                    "updated_by_user": repealed_work.updated_by_user,
                 } for repealed_work in self.instance.repealed_works.all()]
             }
             if self.is_bound:
@@ -971,6 +972,7 @@ class FindPubDocForm(forms.Form):
 class RepealMadeForm(forms.Form):
     repealed_work = forms.ModelChoiceField(queryset=Work.objects)
     repealed_date = forms.DateField(required=False, widget=forms.DateInput(attrs={'type': 'date'}))
+    updated_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
 
     def __init__(self, work, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -986,13 +988,14 @@ class RepealMadeForm(forms.Form):
     def save(self):
         work = self.cleaned_data['repealed_work']
         if self.cleaned_data.get('DELETE'):
-            work.repealed_by = None
-            work.repealed_date = None
-            work.save()
+            if work.repealed_by == self.work:
+                work.repealed_by = None
+                work.repealed_date = None
+                work.save_with_revision(user=self.cleaned_data['updated_by_user'])
         else:
             work.repealed_by = self.work
             work.repealed_date = self.cleaned_data['repealed_date']
-            work.save()
+            work.save_with_revision(user=self.cleaned_data['updated_by_user'])
 
 
 RepealMadeBaseFormSet = formset_factory(
@@ -1003,8 +1006,8 @@ RepealMadeBaseFormSet = formset_factory(
 
 
 class AmendmentForm(forms.Form):
-    amending_work = forms.ModelChoiceField(queryset=Work.objects, required=True)
-    date = forms.DateField(widget=forms.DateInput(attrs={'type': 'date'}), required=True)
+    amending_work = forms.ModelChoiceField(queryset=Work.objects)
+    date = forms.DateField(widget=forms.DateInput(attrs={'type': 'date'}), required=False)
     updated_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
     created_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
 
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -1065,6 +1065,5 @@ class AmendmentsBaseFormSet(AmendmentsFormSet):
             amending_work = form.cleaned_data.get('amending_work')
             date = form.cleaned_data.get('date')
             if (amending_work, date) in seen:
-                print(seen)
                 raise ValidationError("Amending work and date must be unique together.")
             seen.add((amending_work, date))
--- a/indigo_app/forms.py
+++ b/indigo_app/forms.py
@@ -21,7 +21,7 @@ from cobalt import FrbrUri
 from indigo_app.models import Editor
 from indigo_api.models import Document, Country, Language, Work, PublicationDocument, Task, TaskLabel, User, Subtype, \
     Workflow, \
-    VocabularyTopic, Commencement, PlaceSettings, TaxonomyTopic, Locality
+    VocabularyTopic, Commencement, PlaceSettings, TaxonomyTopic, Locality, Amendment
 
 
 class WorkForm(forms.ModelForm):
@@ -115,6 +115,26 @@ class WorkForm(forms.ModelForm):
 
             self.repeals_made_formset = self.RepealMadeBaseFormSet(**repeals_made_formset_kwargs)
 
+            self.AmendmentsBaseFormSet = formset_factory(AmendmentForm, extra=0, can_delete=True)
+
+            amendments_formset_kwargs = {
+                "form_kwargs": {
+                    "work": self.instance,
+                },
+                "prefix": "amendments",
+                "initial": [{
+                    "amending_work": amendment.amending_work,
+                    "date": amendment.date,
+                    "created_by_user": amendment.created_by_user,
+                    "updated_by_user": amendment.updated_by_user,
+                }
+                   for amendment in Amendment.objects.filter(amended_work=self.instance)],
+            }
+            if self.is_bound:
+                amendments_formset_kwargs["data"] = self.data
+
+            self.amendments_formset = self.AmendmentsBaseFormSet(**amendments_formset_kwargs)
+
         self.fields['frbr_doctype'].choices = [
             (y, x)
             for (x, y) in settings.INDIGO['DOCTYPES'] + settings.INDIGO['EXTRA_DOCTYPES'].get(self.country.code, [])
@@ -172,6 +192,7 @@ class WorkForm(forms.ModelForm):
         self.save_publication_document()
         self.save_commencement()
         self.save_repeals()
+        self.save_amendments()
         return work
 
     def save_properties(self):
@@ -188,6 +209,11 @@ class WorkForm(forms.ModelForm):
         for repeal_form in self.repeals_made_formset:
             if repeal_form.is_valid() and repeal_form.has_changed():
                 repeal_form.save()
+
+    def save_amendments(self):
+        for amendment_form in self.amendments_formset:
+            if amendment_form.is_valid() and amendment_form.has_changed():
+                amendment_form.save()
 
     def save_publication_document(self):
         pub_doc_file = self.cleaned_data['publication_document_file']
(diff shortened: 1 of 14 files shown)


goose-life:
very cool! 
Purely based on the video, I haven't looked at the code yet:
- the default date should be the date of commencement of the amending work, if there is one – otherwise blank as you have it is right 
- the amendments should move up in the form, to be after commencements but before repeals (to reflect the lifecycle of a work)
- and while you're in there
  - the default date of repeal for repeals made should be the commencement date of the current work, if there is one
  - I made a comment on your closed PR yesterday about saving with revision rather than saving, when we're making changes to related works

indigo_app/templates/indigo_app/place/_work_chooser.html
```diff
@@ -79,8 +79,12 @@ <h5 class="modal-title">{% trans 'Choose a work' %}</h5>
...
+      <input type="submit" class="btn btn-primary" value="{% trans 'OK' %}" form="work-chooser-form-{{ work_field }}" formnovalidate>
     </div>
-
   </div>
 </div>
+<script>
+    document.querySelector('#work-chooser-form-{{ work_field }}').addEventListener('submit', () => {
+        $(document.querySelector('#work-chooser-modal-{{ work_field }}').parentElement).modal('hide');
+    });
+</script>
```
actlikewill, line 90: @longhotsummer I've made some changes to the work chooser. It was causing some strange behaviour because of conflicting ids, Each time a modal is created from a section of the form, the work chooser form is loaded with the same set of ids causing  the form to hang. I'm using the work field to differentiate them. I tried to use vanilla JS to auto close the modal after form submit but it is not working properly. The modal dissapears but the background doesn't :cry:  I've just replace the code here like you had it in workchooser.js

indigo_app/views/works.py
```diff
@@ -1319,37 +1319,42 @@ def get_context_data(self, **kwargs):
...
             if repeal_made:
                 repealed_work = Work.objects.filter(pk=repeal_made).first()
-                if repealed_work:
+                if repealed_work and repealed_work.repealed_by != self.work and not any(
+                        [True for i in initial if i["repealed_work"] == repealed_work]):
                     initial.append({
-                         'repealed_work': repealed_work,
-                         'repealed_date': repealed_work.repealed_date,
+                        'repealed_work': repealed_work,
+                        'repealed_date': repealed_work.repealed_date or self.work.commencement_date,
```
actlikewill, line 1356: @goose-life I've added the default dates for repealed work here

indigo_app/views/works.py
```diff
@@ -1359,3 +1364,48 @@ def get_context_data(self, **kwargs):
...
+                amending_work = Work.objects.filter(pk=amendment).first()
+                if amending_work:
+                    # add amendment only if it does not exist and is not in the initial data
+                    if not Amendment.objects.filter(amended_work=self.work, amending_work=amending_work).exists() and not any(
+                            [True for i in initial if i["amending_work"] == amending_work]):
+
+                        initial.append({
+                            "amended_work": self.work,
+                            "amending_work": amending_work,
+                            "date": amending_work.commencement_date,
```
actlikewill, line 1402: @goose-life default date for amending work

indigo_app/templates/indigo_app/place/_work_chooser.html
actlikewill: I think it may be possible to have only one work chooser modal for everything ...

indigo_app/forms.py
```diff
@@ -108,13 +108,34 @@ def __init__(self, country, locality, *args, **kwargs):
                 "initial": [{
                     "repealed_work": repealed_work,
                     "repealed_date": repealed_work.repealed_date,
+                    "updated_by_user": repealed_work.updated_by_user,
```
goose-life, line 111: I think this can come out

indigo_app/forms.py
```diff
@@ -108,13 +108,34 @@ def __init__(self, country, locality, *args, **kwargs):
...
+            amendments_formset_kwargs = {
+                "form_kwargs": {
+                    "work": self.instance,
+                },
+                "prefix": "amendments",
+                "initial": [{
+                    "amending_work": amendment.amending_work,
+                    "date": amendment.date,
+                    "created_by_user": amendment.created_by_user,
+                    "updated_by_user": amendment.updated_by_user,
```
goose-life, line 130: I think created_by_user and updated_by_user can come out? if you just use self.request.user later

indigo_app/forms.py
```diff
@@ -937,6 +972,7 @@ class FindPubDocForm(forms.Form):
 class RepealMadeForm(forms.Form):
     repealed_work = forms.ModelChoiceField(queryset=Work.objects)
     repealed_date = forms.DateField(required=False, widget=forms.DateInput(attrs={'type': 'date'}))
+    updated_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
```
goose-life, line 975: I think this can come out

indigo_app/forms.py
```diff
@@ -952,10 +988,54 @@ def is_repealed_work_saved(self):
     def save(self):
         work = self.cleaned_data['repealed_work']
         if self.cleaned_data.get('DELETE'):
-            work.repealed_by = None
-            work.repealed_date = None
-            work.save()
+            if work.repealed_by == self.work:
+                work.repealed_by = None
+                work.repealed_date = None
+                work.save_with_revision(user=self.cleaned_data['updated_by_user'])
```
goose-life, line 994: I think you can pass self.request.user to form.save() from the view and use that instead — you could rename it as save_changes if this is a builtin that you're overriding, see WorkBulkUpdateForm. see https://github.com/laws-africa/indigo/commit/17fd3996f7a84b20f1ee07b602db3111d63ed085 (that's gone now so you won't see it elsewhere, but it worked! :))

indigo_app/forms.py
```diff
@@ -952,10 +988,54 @@ def is_repealed_work_saved(self):
...
         else:
             work.repealed_by = self.work
             work.repealed_date = self.cleaned_data['repealed_date']
-            work.save()
+            work.save_with_revision(user=self.cleaned_data['updated_by_user'])
```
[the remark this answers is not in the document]
goose-life: same here

indigo_app/forms.py
```diff
@@ -952,10 +988,54 @@ def is_repealed_work_saved(self):
...
+    extra=0,
+    can_delete=True,
+)
+
+
+class AmendmentForm(forms.Form):
+    amending_work = forms.ModelChoiceField(queryset=Work.objects)
+    date = forms.DateField(widget=forms.DateInput(attrs={'type': 'date'}), required=False)
+    updated_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
+    created_by_user = forms.ModelChoiceField(queryset=User.objects, required=False)
```
goose-life, line 1012: I think updated_by_user and created_by_user can come out

indigo_app/forms.py
```diff
@@ -952,10 +988,54 @@ def is_repealed_work_saved(self):
...
+        if self.cleaned_data.get('DELETE'):
+            Amendment.objects.filter(amended_work=self.work, amending_work=self.cleaned_data['amending_work']).delete()
+        else:
+            Amendment.objects.update_or_create(
+                amended_work=self.work,
+                amending_work=self.cleaned_data['amending_work'],
+                defaults={
+                    "date": self.cleaned_data["date"],
+                    "created_by_user": self.cleaned_data["created_by_user"],
+                    "updated_by_user": self.cleaned_data["updated_by_user"],
```
goose-life, line 1032: same here

indigo_app/templates/indigo_api/_work_form_repeals_made_form.html
```diff
@@ -10,6 +10,7 @@
         <br />
         <span>{{ form.repealed_work_obj.frbr_uri }}</span>
         {{ form.repealed_work.as_hidden }}
+        {{ form.updated_by_user.as_hidden }}
```
goose-life, line 13: I think this can come out

indigo_app/views/works.py
```diff
@@ -1319,37 +1319,42 @@ def get_context_data(self, **kwargs):
...
+
+                # if repeal exists update the user only if the repealed date has changed
+                user = form.cleaned_data['updated_by_user']
+                if form.is_repealed_work_saved() and form.cleaned_data['repealed_date'] != form.repealed_work_obj.repealed_date:
+                    user = self.request.user
+
                 initial.append({
                     'repealed_work': form.cleaned_data['repealed_work'],
                     'repealed_date': form.cleaned_data['repealed_date'],
+                    'updated_by_user': user,
```
goose-life, line 1345: at least some of this extra logic can probably fall away too

indigo_app/views/works.py
```diff
@@ -1319,37 +1319,42 @@ def get_context_data(self, **kwargs):
...
                 repealed_work = Work.objects.filter(pk=repeal_made).first()
-                if repealed_work:
+                if repealed_work and repealed_work.repealed_by != self.work and not any(
+                        [True for i in initial if i["repealed_work"] == repealed_work]):
                     initial.append({
-                         'repealed_work': repealed_work,
-                         'repealed_date': repealed_work.repealed_date,
+                        'repealed_work': repealed_work,
+                        'repealed_date': repealed_work.repealed_date or self.work.commencement_date,
+                        'updated_by_user': self.request.user,
```
goose-life, line 1357: I think this can come out

indigo_app/views/works.py
```diff
@@ -1359,3 +1364,48 @@ def get_context_data(self, **kwargs):
...
+                delete = form.cleaned_data.get('DELETE')
+                if delete:
+                    if not Amendment.objects.filter(amended_work=self.work, amending_work=form.cleaned_data["amending_work"]).exists():
+                        continue
+                initial.append({
+                    "amended_work": self.work,
+                    "amending_work": form.cleaned_data["amending_work"],
+                    "date": form.cleaned_data["date"],
+                    "created_by_user": form.cleaned_data["created_by_user"] or self.request.user,
+                    "updated_by_user": form.cleaned_data["updated_by_user"] or self.request.user,
```
goose-life, line 1388: same here

goose-life (commented):
all versions of the same thing, the main thing is to see https://github.com/laws-africa/indigo/commit/17fd3996f7a84b20f1ee07b602db3111d63ed085

indigo_app/forms.py
actlikewill: @goose-life We are calling repeals_made_form.save() from within the work form save method, I think its better to let the inner forms (repeals and amendments) manage their own data,  they can be responsible for setting and updating the user if they change. Since we don't have request object in the form class, the easiest way to pass the user data is as a form field.

goose-life: got it! okay I thought I might be missing something 😅

indigo_app/forms.py
actlikewill: Thought they are null=True, they are used elsewhere in the code, I think to create the actions stream

goose-life: That's 100% right yes — I was still hoping you'd just be able to pass the user straight, but they're definitely needed if you can't

indigo_app/views/works.py
goose-life: this part I'm unclear on

indigo_app/templates/indigo_app/place/_work_chooser.html
longhotsummer: @actlikewill I think this is best solved by putting this only once into the page. So the top-level page that knows this form will be used should include the initial wrapper. All the components that then target the work chooser modal will target the same modal.

actlikewill: @longhotsummer I've moved the modal into the work form. It was a a matter of changing the htmx and bootstrap targets

indigo_app/forms.py
```diff
@@ -74,11 +74,12 @@ class Meta:
     commencing_work = forms.ModelChoiceField(queryset=Work.objects, required=False)
     commencement_note = forms.CharField(max_length=1024, required=False)
 
-    def __init__(self, country, locality, *args, **kwargs):
+    def __init__(self, country, locality, user, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.country = country
         self.locality = locality
         self.place = locality or country
+        self.user = user
```
actlikewill, line 82: @goose-life @longhotsummer I've added the user in the init method

indigo_app/templates/indigo_api/_work_form_content.html
```diff
@@ -359,6 +370,19 @@ <h4 class="text-muted work-frbr-uri">{{ work.frbr_uri }}</h4>
       </div>
     {% endblock work-form-commencement %}
 
+    {% block work-form-amendments %}
+      {% if work.pk %}
+        <div class="card mb-3" id="repeals-made">
```
goose-life, line 375:
```suggestion
        <div class="card mb-3" id="amendments">
```
or something — does this id actually get used? maybe this one (and repeals-made) can go away?

indigo_app/views/works.py
```diff
@@ -1346,16 +1345,69 @@ def get_context_data(self, **kwargs):
...
+                    "date": form.cleaned_data["date"],
+                    "DELETE": form.cleaned_data["DELETE"],
+                })
+            amendment = self.request.POST.get("amendment")
+            if amendment:
+                amending_work = Work.objects.filter(pk=amendment).first()
+                if amending_work:
+                    # add amendment only if it does not exist and is not in the initial data
+                    if not Amendment.objects.filter(amended_work=self.work, amending_work=amending_work).exists() and not any(
+                            [True for i in initial if i["amending_work"] == amending_work]):
```
goose-life, line 1397: hmmmmmm does this imply that we can't add more than one amendment by the same work? or am I misreading.

longhotsummer:
Nice, this is looking good. I'll let Edith sanity check the business logic of the amendment creation.

indigo_app/views/works.py
goose-life, line 1397:
```suggestion
                    in_initial_dates = [i["date"] for i in initial if i["amending_work"] == amending_work]
                    existing_amendments = Amendment.objects.filter(amended_work=self.work, amending_work=amending_work, date__in=in_initial_dates).all()
                    if not existing_amendments:
```
This is a toughie, damn! 😂 
I'm not sure if this does the right thing, but what we want is to be able to add multiple amendments by the same work, as long as they're at different dates.
So the only thing we don't want to do is to add another amendment by the same work at the same date, because then we'll get issues on an attempted save.
(Another thought: maybe if this is not the first amendment by this amending work, we actually don't give a default date?)

goose-life: oh, another thing we might do here is to only look in `initial` for another amendment by the same amending work _that doesn't yet have a date associated_ — and if there is one, don't add another one?

longhotsummer: An amendment happens through an associated object. Why can we store the pk of that object in the related form? When the form is saved, we know which Amendment objects must be update or created based on this pk.

goose-life: haha yes that sounds much better 😅

actlikewill: yeah, thats what I was doing. I take it (from this comment and the ones below) that we are allowed to have more than one amendment by the same work? just not at the same date?

indigo_app/forms.py
```diff
@@ -949,13 +983,88 @@ def repealed_work_obj(self):
...
+            return
+        # check if amending work and date are unique together
+        seen = set()
+        for form in self.forms:
+            if form.cleaned_data.get('DELETE'):
+                continue
+            amending_work = form.cleaned_data.get('amending_work')
+            date = form.cleaned_data.get('date')
+            if (amending_work, date) in seen:
+                print(seen)
```
longhotsummer, line 1068:
```suggestion
```

longhotsummer (dismissed):

longhotsummer (approved):

PR #1956, opened 2024-02-01
Pin boto3 version

goose-life:


longhotsummer (approved):
let's see if this works

Issue #1940, opened 2024-01-22
As a managing editor I want to select the label on a task more easily


goose-life:
![Image](https://github.com/laws-africa/indigo/assets/32566441/187d3ac1-6d00-4453-b74d-6356b6433a46)

'Projects' and 'Blocked by' are selectable on the work detail page; but to change the labels one needs to click into 'Edit' first.

PR #1960, opened 2024-02-05
edit task labels in detail view using HTMX, style using cards

goose-life:
https://www.loom.com/share/1a61bd5dc7e84b17b7e112fbc8f66aad?sid=3776fe1e-726a-43cb-8bc6-d725b040f8c4

not high priority but will make working with task labels easier, especially on tasks that take long to load because of resolving an anchor in the document


indigo_app/views/tasks.py
```diff
@@ -287,6 +290,21 @@ def get_context_data(self, **kwargs):
         return context
 
 
+class TaskEditLabelsView(SingleTaskViewBase, FormView, SingleObjectMixin):
```
longhotsummer, line 293: If you make this inherit from `UpdateView`, then it should Just Work because it knows how to apply changes to an existing object. You'd need to make the form class a ModelForm, but it only needs the labels field.

Probably not vastly different to what you have, but it does use django standards a bit more.

longhotsummer: also double check the right perms are required (change task, and the country perm)

indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -191,124 +191,168 @@ <h5>
...
+                hx-target="#task-labels-list"
+                name="labels"
+                data-style="btn-link ps-0 text-dark"
+                {% if not perms.indigo_api.change_task %}disabled{% endif %}
+                data-none-selected-text="{% trans 'Choose labels' %}"
+                data-selected-text-format="static"
+                data-size="7"
+            >
+              {% for label in possible_labels %}
+                <option value="{{ label.id }}" {% if label in task.labels.all %}selected{% endif %}>
```
longhotsummer, line 247: this re-requires the db for every option. you can either load the list once with a `with ...` or, pre-populate a TaskForm and then you can just iterate over `options` and it'll do the right thing, including knowing which are selected.

longhotsummer:
Such a great improvement! I love using cards, we should have done this from the start.

indigo_app/views/tasks.py
goose-life: thanks! I knew there was something fishy about it :)

indigo_app/templates/indigo_api/_task_detail.html
goose-life: I'd like to go with the second option but I'm actually unclear on how to do either of them 😅

goose-life: simply putting a `form` in the template uses the CommentForm by default, and I'm not sure how to change that?

goose-life:
@longhotsummer would you be happy to merge this now, and we tackle blocking tasks and potential assignees etc separately? I'm currently blocked on speeding up blocking tasks.

longhotsummer (approved):
merge it!

longhotsummer:
Fixes https://github.com/laws-africa/indigo/issues/1940

indigo_app/views/tasks.py
```diff
@@ -150,8 +148,8 @@ def get_context_data(self, **kwargs):
             context['work'] = task.work
             context['work_json'] = json.dumps(WorkSerializer(instance=task.work, context={'request': self.request}).data)
 
-        if task.code:
-            context['task_type'] = dict(Task.CODES)[task.code]
```
goose-life, line 154: this was added on master but not actually used in the end

longhotsummer (approved):

PR #1951, opened 2024-01-31
Tasks on approve

goose-life:
https://www.loom.com/share/bdfc22b254ef4b23800c6b474ef642c5?sid=44159d91-d79a-46b6-bbc6-353276b979bc

PR #1961, opened 2024-02-05
Add work tweaks

longhotsummer:
* new flow for creating an off-canvas work
* simplification of templates and forms
* support pre-fill of some details in new work form


indigo_app/views/works.py
```diff
@@ -1177,11 +1188,12 @@ def post(self, request, *args, **kwargs):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        form = self.Form(self.country, self.locality, self.request.POST)
-        form.full_clean()
 
-        context["actual_duplicate"] = form.find_actual_duplicate(self.request.GET.get('pk'))
-        context["possible_duplicates"] = form.find_possible_duplicates(self.request.GET.get('pk'))
+        if self.request.GET.get('pk'):
```
goose-life, line 1192: this looks like duplicates will only be looked for if the work already exists? which isn't what we want, if I'm reading this right — the duplicate finder should help people to realise that they're about to create a possible duplicate work as well.

longhotsummer: reverted this, it was breaking when the pk was null, but that seems to be fixed

longhotsummer:
@goose-life ready for re-review

indigo_app/views/works.py
```diff
@@ -176,7 +176,25 @@ def get_success_url(self):
         return reverse('work', kwargs={'frbr_uri': self.work.frbr_uri})
 
 
-class UnapproveWorkView(WorkViewBase, View):
```
goose-life, line 179: this isn't right; ApproveWorkView has fallen away and UnapproveWorkView is now its own thing: see https://github.com/laws-africa/indigo/pull/1951/commits/125dd5c4f893abcd83a04a3b021dc4cd4ef864d0

goose-life: I think it's fixed in https://github.com/laws-africa/indigo/pull/1965

indigo_app/templates/indigo_app/place/_work.html
```diff
@@ -25,7 +25,7 @@
           <div class="d-flex">
             <div class="col-9 ps-0">
               <div>
-                <a href="{% url 'work' frbr_uri=work.frbr_uri %}"><span>{{ work.title }}</span></a>
+                <a href="{% url 'work' frbr_uri=work.frbr_uri %}">{{ work.title }}</a>
```
goose-life, line 28:
```suggestion
                <a href="{% url 'work' frbr_uri=work.frbr_uri %}"><span>{{ work.title }}</span></a>
```

I know this sounds crazy, but that span is what's making the link actually clickable versus just expanding and collapsing the card (see https://www.loom.com/share/512c3599cf534efc8f4bc172bf864188)

longhotsummer: woah, it seems to be related to how data-bs-toggle works to cancel the event if it comes from `<a>`

PR #1966, opened 2024-02-07
don't double-create components

longhotsummer:
htmx:load fires on page load AND on htmx element loads (which is weird)

Ref https://github.com/bigskysoftware/htmx/issues/1500

This fixes a problem where our `data-component` logic is attached twice to the same element


goose-life:
oh nice!

PR #1970, opened 2024-02-09
commencement and repeal in work list overview

longhotsummer:
image

image

I still think we should combine (or re-use parts of) the logic for this snippet and the work detail page box.


longhotsummer:
Also more work list sorting options

PR #1979, opened 2024-02-15
offcanvas work form has button at the top

longhotsummer:
https://www.loom.com/share/8c61127caa524848b07ddc126a061ced

Moves the "Save" button on the off-canvas work edit form to the offcanvas header, so that it's always visible.

### DIFF
--- a/indigo_app/templates/indigo_api/_work_form_offcanvas_body.html
+++ b/indigo_app/templates/indigo_api/_work_form_offcanvas_body.html
@@ -1,0 +1,19 @@
+{% load i18n %}
+
+<div class="offcanvas-header border-bottom" id="offCanvasWorkForm">
+  {% if form %}
+    <button
+      class="btn btn-success save"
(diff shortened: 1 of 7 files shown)


indigo_app/templates/indigo_api/_work_form_offcanvas_body.html
```diff
@@ -0,0 +1,19 @@
+{% load i18n %}
+
+<div class="offcanvas-header border-bottom" id="offCanvasWorkForm">
+  {% if form %}
+    <button
+      class="btn btn-success save"
+      type="submit"
+      form="edit-work-form"
+    >
+      {% if work.pk %}{% trans 'Save' %}{% else %}{% trans 'Create' %}{% endif %}
```
goose-life, line 10: I think for now there's no option of creating a new work using the offcanvas, but this'll be handy if and when that changes

PR #1986, opened 2024-02-19
fix form-to-url conversion

longhotsummer:
we want raw data, support multiple values, exclude anything that the form doesn't explicitly care about

https://www.loom.com/share/1629d630f4454df5bcc51836935875c1


goose-life (approved):
nice!

actlikewill (approved):

PR #1990, opened 2024-02-19
Commencements made

actlikewill:
- This adds commencements to the work form
https://www.loom.com/share/aeb2c485ea164afcbef8b253776d3d27?sid=9618bf0d-59ee-43e0-b47e-124b764e9ee3


goose-life:
very nice! 
yes I totally agree — if there's already a main commencement on the work being commenced, we shouldn't (be able to) set either of ours. this goes for non-multiple commencements too.
if a commenced work also has already been marked as 'all provisions commenced' (which would be incorrect but possible), that could also create issues.
I'll review the code now and see if I can suggest something.

indigo_app/forms/works.py
```diff
@@ -480,6 +508,108 @@ def clean(self):
...
+            seen.add((amending_work, amended_work, date))
+
+    def save(self, user, work, *args, **kwargs):
+        if self.is_valid() and self.has_changed():
+            for form in self.forms:
+                form.save(user)
+
+            commenced_counts = Commencement.objects.filter(commencing_work=work).values("commenced_work").annotate(num_commencements=Count("id")).order_by()
+            multiple_commencements = commenced_counts.filter(num_commencements__gt=1).values_list("commenced_work", flat=True)
+            single_commencements = commenced_counts.filter(num_commencements=1).values_list("commenced_work", flat=True)
```
goose-life, line 591: so here we should also be taking into account any other existing commencements on all the commenced works, as you rightly pointed out in your video.
- if any of those other commencements is already `main`, don't set any of ours to `main`. if none are, we can still use the earliest (or only) one that we're adding
- if any have `all_provisions=True`, we should set that to False and save that commencement, and make ours False as well

you can probably treat all 'multiple' and 'single' commencements the same, as long as 
- the single and multiple lists are based on all commencements on the commenced work (Commencement.filter(commenced_work=commenced_work))
- the multiples logic is tweaked to check for existing `main` and `all_provisions`

goose-life:
I think some of your logic around 'if commencement made / else commenced by' can probably come out, as we likely won't be reusing too much of the commencements for the other side of the relationship — but it doesn't hurt to leave it in for now and see, maybe we do end up reusing some of it!

indigo_app/forms/works.py
actlikewill: @goose-life I've updated the PR with our changes and its working ok. See the demo https://www.loom.com/share/0a35e3c039d948deb5eb802708e8c4fd?sid=73bc0379-b580-4c2e-b4c3-2843677f8f71

actlikewill:
> I think some of your logic around 'if commencement made / else commenced by' can probably come out, as we likely won't be reusing too much of the commencements for the other side of the relationship — but it doesn't hurt to leave it in for now and see, maybe we do end up reusing some of it!

@goose-life I've removed references to the commence_by in the code. I don't think we will use it if we do commenced_by separately.

goose-life (approved):

PR #1991, opened 2024-02-21
add 'public' to TaxonomyTopic, filter to public for API and serializer

goose-life:
The additional test fixtures are to check that the test still passes (it only sees the public root).

I suspect Work.public_taxonomy_topics() can be finessed a bit, but it works as is.


indigo_api/models/works.py
```diff
@@ -149,6 +150,11 @@ def fix_up(item):
             }] + tree
         return tree
 
+    @classmethod
+    def get_public_root_nodes(cls):
+        root_nodes = cls.get_root_nodes()
+        return root_nodes.exclude(public=False)
```
longhotsummer, line 156: Minor: double negatives lead to confusion, prefer `.filter(public=True)`.

longhotsummer (dismissed):

longhotsummer (approved):

PR #1999, opened 2024-02-22
Fix task facet counts

goose-life:
https://www.loom.com/share/5d998687b7d64b5ba9a56004e4f3c3e8?sid=da1661a8-9ca2-43b1-8fbf-85e86e46f88b


longhotsummer (approved):
and it seems we get a cleaner url, which is great

PR #2002, opened 2024-02-27
Task file and conversion task

goose-life:
Next up
- output files
- automatically creating these tasks from source document

https://www.loom.com/share/d20f594a6dc24bb19f1796e4f8734945?sid=2e866217-365f-4997-862a-867735f742fd


goose-life:
I also haven't limited which tasks the files show up on — if we can assume that conversion tasks will all automatically be created, we can just check the task code and only include the subforms if it's `convert-document`, but if we want people to be able to manually create them too we'll have to do some more htmx magic to re-render those parts when the task type is chosen. So I'm leaving it alone for now, but we do have options.

indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -67,6 +67,28 @@ <h5>
         <div class="card">
           <div class="card-body">
             {{ task.description|default:''|urlize|linebreaksbr }}
+            {% if task.input_file %}
+              <div class="mt-2">
+                <h6>{% trans "Input" %}</h6>
```
longhotsummer, line 72:
```suggestion
                <h6>{% trans "Input file" %}</h6>
```

longhotsummer: since we're moving our forms to "key: value" with `<dl>`, maybe use that here too? It makes it a bit clearer that it's not part of the description text, too.

indigo_app/templates/indigo_api/_task_detail.html
```diff
@@ -67,6 +67,28 @@ <h5>
...
+                <a href="{% url 'task_input_file' place=place.place_code pk=task.pk %}"
+                   target="_blank" rel="noopener">
+                  {{ task.input_file.filename }}
+                </a>
+                <span class="text-muted">({{ task.input_file.size|filesizeformat }})</span>
+              </div>
+            {% endif %}
+            {% if task.output_file %}
+              <div class="mt-2">
+                <h6>{% trans "Output" %}</h6>
```
longhotsummer, line 83:
```suggestion
                <h6>{% trans "Output file" %}</h6>
```

indigo_app/templates/indigo_api/_task_input_file_form.html
```diff
@@ -0,0 +1,70 @@
+{% load i18n %}
+{% url 'task_form_input_file' place.place_code as edit_url %}
+<input type="hidden" name="{{ form.clear.html_name }}" value="{{ form.clear.value|default_if_none:'' }}">
+{% if task.input_file %}
+  <input type="hidden" name="{{ form.url.html_name }}" value="{{ task.input_file.url|default_if_none:'' }}">
+  <div class="mb-3 row">
+    <label class="col-2 col-form-label">{% trans 'Input' %}</label>
```
longhotsummer, line 7:
```suggestion
    <label class="col-2 col-form-label">{% trans 'Input file' %}</label>
```

indigo_app/templates/indigo_api/_task_input_file_form.html
```diff
@@ -0,0 +1,70 @@
...
+  </div>
+
+  <div class="mb-3 row">
+    <label for="{{ form.file.id_for_label }}" class="col-2 col-form-label">{% trans 'Input file' %}</label>
+    <div class="col-4">
+      <input
+          type="file"
+          id="{{ form.file.id_for_label }}"
+          name="{{ form.file.html_name }}"
+          accept=".pdf,.docx"
```
longhotsummer, line 59: I think we'll need .rtf and .doc as well.

longhotsummer:
This looks really great. I'm not sure how often the URL will actually be used, but let's find out.

I don't think we should special case the input and output fields for different task types, it's too clever. Keep it simple like you have done.

indigo_app/forms/tasks.py
```diff
@@ -52,6 +55,71 @@ def clean_title(self):
...
+            self.initial.get('file').delete()
+        if url:
+            resp = requests.get(url, timeout=10)
+            resp.raise_for_status()
+            filename = urlparse(url).path.split('/')[-1] or url
+            task_file.url = url
+            task_file.size = len(resp.content)
+            task_file.filename = filename
+            task_file.mime_type = resp.headers['Content-Type']
+            if self.prefix == 'input_file':
```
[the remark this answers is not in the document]
goose-life: it's needed for save() just because of where we store literal files — I'm moving the save() out of here and onto the task form, let me know what you think

indigo/tasks.py
```diff
@@ -0,0 +1,79 @@
...
+            'import-content': _('Import content'),
+            'link-gazette': _('Link Gazette'),
+            'apply-amendment': _('Apply amendment'),
+        }
+        task_title = task_titles[task_type]
+
+        task = Task.objects.filter(work=work, code=task_type, timeline_date=timeline_date).first()
+        if not task:
+            task = Task(country=work.country, locality=work.locality, work=work,
+                        code=task_type, timeline_date=timeline_date, created_by_user=user)
```
longhotsummer, line 56: `Task.objects.get_or_create()` will do this for you and guard against race conditions.

indigo/tasks.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def get_or_create_task(self, work, task_type, description, user, timeline_date=None):
+        task_titles = {
+            'import-content': _('Import content'),
+            'link-gazette': _('Link Gazette'),
+            'apply-amendment': _('Apply amendment'),
+        }
+        task_title = task_titles[task_type]
+
+        task = Task.objects.filter(work=work, code=task_type, timeline_date=timeline_date).first()
```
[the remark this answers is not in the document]
longhotsummer: Should we not filter out DONE tasks here? Otherwise, if the task is done, the logic below will update the description but never mark the task is available again. If the only existing tasks are DONE, then we should create a new one.

indigo/tasks.py
```diff
@@ -0,0 +1,79 @@
...
+        for amendment in self.amendments:
+            amendment_tasks.append(self.get_or_create_task(
+                work=amendment.amended_work, task_type='apply-amendment',
+                description=data[f'amendment_task_description_{amendment.pk}'],
+                user=user, timeline_date=amendment.date))
+        if data.get('update_amendment_tasks'):
+            self.block_or_cancel_tasks(amendment_tasks, data['update_amendment_tasks'], user)
+
+    def get_or_create_task(self, work, task_type, description, user, timeline_date=None):
+        task_titles = {
```
longhotsummer, line 46: This is useful to define on the class itself so new task types can easily be added by subclasses.

longhotsummer:
Nice, the Task Broker does seem to make thing simpler.

goose-life:
image

image

indigo/tasks.py
```diff
@@ -19,11 +20,12 @@ def __init__(self, works):
     def create_tasks(self, user, data):
         # import tasks
         # TODO: add the appropriate timeline date for Import tasks too?
-        import_tasks = [self.get_or_create_task(work=work, task_type='import-content',
-                                                description=data['import_task_description'],
-                                                user=user) for work in self.import_task_works]
+        # stash the import tasks in case we want to do something with them later
+        self.import_tasks = [self.get_or_create_task(work=work, task_type='import-content',
```
longhotsummer, line 24: save all the tasks like this

indigo/tasks.py
goose-life: I got rid of this altogether, we already have this on the model

longhotsummer (dismissed):
great

actlikewill (approved):

PR #2003, opened 2024-02-29
Tasks in timeline

goose-life:
image


goose-life:
cards
image

list
image

indigo_app/views/works.py
```diff
@@ -82,6 +82,28 @@ def get_work_timeline(self, work):
         # add expressions
         for entry in timeline:
             entry.expressions = [e for e in work_expressions if e.expression_date == entry.date]
+        # add tasks
+        timeline_tasks = work.tasks.filter(timeline_date__isnull=False)
```
longhotsummer, line 86: do you need to exclude done/cancelled tasks?

indigo_app/views/works.py
```diff
@@ -82,6 +82,28 @@ def get_work_timeline(self, work):
...
+        dates = [entry.date for entry in timeline]
+        # simple case: add tasks to existing corresponding entries
+        for entry in timeline:
+            entry.tasks = timeline_tasks.filter(timeline_date=entry.date)
```
longhotsummer, line 90: bear in mind that when the template iterates over `entry.tasks`, it will go to the db for each point in time. for something like the income tax act, that could be 50 times.

going to the db is expensive;  we know that we're going to use every task we fetch from the db, so it's better to load all the timeline-related tasks on line 86 and store them in a list, and then split them up in this loop.

indigo_app/views/works.py
```diff
@@ -82,6 +82,28 @@ def get_work_timeline(self, work):
...
+
+        # these will have their own entries as their dates aren't in the timeline yet
+        extra_tasks = timeline_tasks.exclude(timeline_date__in=dates).order_by('timeline_date')
```
longhotsummer, line 93: this only goes to the db once because you do the iteration here, so it's fine.

indigo_app/views/works.py
goose-life: I'm not sure if I've alleviated your concern here; when you say I should store them in a list and split them up, do you mean by some method other than filtering? If I turn timeline_tasks from a queryset into a list I guess I could iterate over the list and compare each one's date to the entry's? is that more what you were thinking?

indigo_app/views/works.py
```diff
@@ -82,6 +82,29 @@ def get_work_timeline(self, work):
         # add expressions
         for entry in timeline:
             entry.expressions = [e for e in work_expressions if e.expression_date == entry.date]
+        # add tasks
+        timeline_tasks = work.tasks.filter(timeline_date__isnull=False).exclude(state=Task.CANCELLED)
+        dates = [entry.date for entry in timeline]
+        # simple case: add tasks to existing corresponding entries
+        for entry in timeline:
+            entry_tasks = timeline_tasks.filter(timeline_date=entry.date)
+            entry.tasks = list(entry_tasks)
```
longhotsummer, line 91: only when `list()` is called will it go to the db; if there are 50 items in the timeline, then it's going go to 50 times.

Instead, pull it into the list on line 86, then filter on line 90 entirely in python, eg `[t for t in timeline_tasks if t.timeline_date == entry.date]`

goose-life: gotcha!

goose-life: I made a separate list for the simple tasks, as the extra tasks still need to filter timeline_tasks to get started.

longhotsummer (approved):

PR #2011, opened 2024-03-04
Conversion tasks

goose-life:
https://www.loom.com/share/5e9105ba705b4b399eca68fa9bbe69fe?sid=edb49ad1-653c-4724-97b2-b4a57b8b30f5


indigo_api/models/tasks.py
```diff
@@ -359,6 +384,10 @@ def update_blocked_tasks(self, task, user):
             for blocked_task in blocked_tasks:
                 action.send(user, verb='resolved a task previously blocking', action_object=blocked_task,
                             target=task, place_code=blocked_task.place.place_code)
+                action.send(user, verb='unblocked', action_object=task,
```
longhotsummer, line 387: why would we do this and then only conditionally actually call `blocked_task.unblock`?

indigo_app/views/tasks.py
```diff
@@ -388,8 +393,42 @@ def post(self, request, *args, **kwargs):
 
         task.save()
 
+        # if a conversion task has been finished, do the import
+        if task.code in ['convert-document'] and self.change in ['finish'] and task.work is not None:
+            self.create_and_import_document(task)
+
         return redirect(self.get_redirect_url())
 
+    def create_and_import_document(self, task):
```
longhotsummer, line 402: Would it make more sense to have this on the task object itself? Alternatively, all the logic for connecting tasks together could be on the task broker object, since it already has to do some of those things.

longhotsummer (approved):

PR #2015, opened 2024-03-07
disable the Save button in the work form on submit

goose-life:
I only got it working once locally, and then not again 🤷 

This will resolve https://lawsafrica.sentry.io/issues/5044647907/?alert_rule_id=1083292&alert_type=issue&notification_uuid=edb31ebd-36f6-4b24-9c91-75325c325e4c&project=5229057&referrer=slack and friends:
duplicate key value violates unique constraint "indigo_api_publicationdocument_work_id_key"
DETAIL:  Key (work_id)=(22322) already exists.

The only way I could find to trigger this error is to click the Save button more than once.


goose-life:
we also want to disable the 'Apply' button on the work filter view: https://github.com/laws-africa/indigo/blob/master/indigo_app/templates/indigo_api/_work_filter_form.html#L32
I thought this was what was causing https://lawsafrica.sentry.io/issues/5045589689/?referrer=alert_email&alert_type=email&alert_timestamp=1709835445558&alert_rule_id=1083292&notification_uuid=6e0f1b88-d3e8-4837-bce8-d476ea75d697&environment=production 
but it's actually something else — either way, it should be disabled :)

longhotsummer:
@goose-life fixed this

longhotsummer:
webpack will be run in master automatically

longhotsummer (approved):

actlikewill:
I'm not able to repro the original error though :grimacing: but I'm ok to merge

goose-life:
The work filter view is a little more complicated, as the filter form itself isn't reloaded when 'Apply' is clicked, so the button would be permanently disabled after the first click. So let's do this one separately?

PR #2026, opened 2024-03-08
filter works by FRBR URI

longhotsummer:
* add FRBR URIs as an advanced filter
* filters the works on the parsed Work FRBR URI; bad FRBR URIs are ignored
* move advanced filters into columns

image


goose-life:
nice!

PR #2029, opened 2024-03-08
Disables submit button on offcanvas

actlikewill:
- adds the disable to the off canvas form


goose-life:
thanks, forgot that's a different button! 🤦

PR #2032, opened 2024-03-11
allow filtering on updated_at and created_at fields

longhotsummer:
this allows callers to ask for data only updated/created since they lasted refreshed their content


goose-life (approved):
nice!

PR #2042, opened 2024-03-18
helper for parsing unicode html safely

longhotsummer:


indigo/xmlutils.py
```diff
@@ -42,3 +42,10 @@ def closest(element, predicate):
         return next(e for e in chain([element], element.iterancestors()) if predicate(e))
     except StopIteration:
         return None
+
+
+html_parser = lxml.html.HTMLParser(encoding='utf-8')
```
goose-life, line 47: so we instantiate this parser literally once (when the app starts) and it gets used whenever parse_html_str is called? cool!

longhotsummer: yup

PR #2048, opened 2024-03-20
I18n works

longhotsummer:
The major idea here is that all the field labels and titles are driven by the form, rather than repeating it everywhere.

### DIFF
--- a/indigo_app/forms/works.py
+++ b/indigo_app/forms/works.py
@@ -1340,7 +1340,8 @@ class WorkBulkUpdateForm(forms.Form):
 
 
 class WorkBulkApproveForm(forms.Form):
-    TASK_CHOICES = [('', 'Create tasks'), ('block', 'Create and block tasks'), ('cancel', 'Create and cancel tasks')]
+    TASK_CHOICES = [('', 'Create tasks'), ('block', _('Create and block tasks')), ('cancel', _('Create and cancel tasks'))]
+
     works_in_progress = forms.ModelMultipleChoiceField(queryset=Work.objects, required=False)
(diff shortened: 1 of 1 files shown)


actlikewill:
@longhotsummer I think I see a few missed errors like this one https://github.com/laws-africa/indigo/blob/f08ff7920a26f1c99d38f0b1459a01b919000f17/indigo_app/forms/works.py#L551

indigo_app/forms/works.py
```diff
@@ -1025,6 +1043,17 @@ def filter_queryset(self, queryset, exclude=None):
 
         return queryset.distinct()
 
+    def facet_item(self, field, value, count):
+        try:
+            label = next(lab for (val, lab) in self.fields[field].choices if val == value)
+        except StopIteration:
+            raise ValueError(f"Unknown choice {value} for field {field}")
```
actlikewill, line 1050: does this error need translation?

actlikewill:
these two:
https://github.com/laws-africa/indigo/blob/f08ff7920a26f1c99d38f0b1459a01b919000f17/indigo_app/forms/works.py#L551
https://github.com/laws-africa/indigo/blob/f08ff7920a26f1c99d38f0b1459a01b919000f17/indigo_app/forms/works.py#L470

indigo_app/forms/works.py
longhotsummer: no, it's not user facing, it will raise a 500-level error. It's to guard against typos lower down.

longhotsummer:
@actlikewill updated, thanks!

PR #2051, opened 2024-03-20
form.py translations

goose-life:


indigo_app/forms/tasks.py
```diff
@@ -146,6 +149,10 @@ def __init__(self, country, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['assigned_to'].queryset = User.objects.filter(editor__permitted_countries=self.country).order_by('first_name', 'last_name').all()
         self.fields['submitted_by'].queryset = self.fields['assigned_to'].queryset
+        # insert 'assigned' in task states after 'open' so that they're in the right order in the form
+        states = list(Task.STATES)
+        states.insert(states.index('open') + 1, 'assigned')
+        self.fields['state'].choices = [(x, _(x.replace('_', ' ').capitalize())) for x in states]
```
longhotsummer, line 155: The translation system extracts strings to translate by scanning the code, not by running it. So it doesn't know what's going to be included in this string. So, unless `Unassigned`, `Assigned`, `Open` etc. are marked as translateable somewhere else, they won't be extracted here and given to the translators to be translated.

In other words, we have to be explicit. 

I suggest adding `Task.STATE_CHOICES` with `((OPEN, _('Unassigned')), ...)`  etc., and when you add `assigned` to the list you add `(ASSIGNED, _('Assigned'))`. Don't rely on transforming the code into a human-friendly form.

longhotsummer (approved):

PR #2055, opened 2024-03-22
All places

longhotsummer:
This adds an "works for all places" view.

* adding a new work is disabled
* countries and works that are visible are based on user permissions
* batch operations sanity check against user permissions
* move help to end of navigation bar

https://www.loom.com/share/7812bd3f3bbf449a9f4968357559fc68


goose-life (approved):
neat-o!

PR #2057, opened 2024-03-24
remove old taxonomy system (replaced with taxonomy topics)

longhotsummer:


goose-life (approved):
should we add something in the changelog?

longhotsummer:
> should we add something in the changelog?

Yes, I was thinking the same thing. Downstream users will need to run the taxonomies -> taxonomy topics migration command before upgrading to this point. So I think we need an official version before this point, and then a major version bump for this release.

PR #2059, opened 2024-03-25
Remove workflow

### LINKED ISSUE #2039: Get rid of old projects (Workflow model)
* [ ] remove from UI
* [ ] remove models

 **Sanity checks**
- [ ] add/update/remove tasks
- [ ] add/update/remove works 
- [ ] bulk add works
- [ ] bulk update works
- [ ] update tests

actlikewill:
Removes workflow model

closes #2039


indigo_api/management/commands/migrate_project_taxonomies.py
longhotsummer, line 1: we need to update docs/changelog.rst to call out that workflows are going away and installations must also run this command when they run the `migrate_taxonomies` command at version 18.1.0.

longhotsummer (approved):

goose-life (approved):
nice :)

PR #2068, opened 2024-03-31
Task facets

actlikewill:
- This adds task facets
https://www.loom.com/share/e18de5a8d7c540ceaeeea23e2e0d3f35?sid=fb18a105-0cf0-4c21-b134-3d5f9aadd070

part of laws-africa/indigo-lawsafrica#1177


indigo_app/forms/tasks.py
```diff
@@ -134,64 +136,137 @@ class Meta:
...
+class TaskFilterForm(WorkFilterForm):
+    labels = forms.ModelMultipleChoiceField(label=_("Labels"), queryset=TaskLabel.objects, to_field_name='slug')
     state = forms.MultipleChoiceField(label=_('State'), choices=Task.STATE_CHOICES)
     format = forms.ChoiceField(choices=[('columns', _('columns')), ('list', _('list'))])
     assigned_to = forms.ModelMultipleChoiceField(label=_('Assigned to'), queryset=User.objects)
     submitted_by = forms.ModelMultipleChoiceField(label=_('Submitted by'), queryset=User.objects)
     type = forms.MultipleChoiceField(label=_('Task type'), choices=Task.CODES)
     country = forms.ModelMultipleChoiceField(queryset=Country.objects.select_related('country'))
-    taxonomy_topic = forms.CharField()
+    taxonomy_topic = forms.ModelMultipleChoiceField(queryset=TaxonomyTopic.objects, to_field_name='slug', required=False)
```
longhotsummer, line 147: this should not be necessary because the superclass has it, I think?

indigo_app/forms/tasks.py
```diff
@@ -134,64 +136,137 @@ class Meta:
     labels = forms.ModelMultipleChoiceField(queryset=TaskLabel.objects, required=False)
 
 
-class TaskFilterForm(forms.Form, FormAsUrlMixin):
-    labels = forms.ModelMultipleChoiceField(queryset=TaskLabel.objects, to_field_name='slug')
+class TaskFilterForm(WorkFilterForm):
+    labels = forms.ModelMultipleChoiceField(label=_("Labels"), queryset=TaskLabel.objects, to_field_name='slug')
     state = forms.MultipleChoiceField(label=_('State'), choices=Task.STATE_CHOICES)
     format = forms.ChoiceField(choices=[('columns', _('columns')), ('list', _('list'))])
```
longhotsummer, line 142: this falls away, it's always a list now

indigo_app/forms/tasks.py
```diff
@@ -134,64 +136,137 @@ class Meta:
...
     country = forms.ModelMultipleChoiceField(queryset=Country.objects.select_related('country'))
-    taxonomy_topic = forms.CharField()
+    taxonomy_topic = forms.ModelMultipleChoiceField(queryset=TaxonomyTopic.objects, to_field_name='slug', required=False)
     sortby = forms.ChoiceField(choices=[
         ('-created_at', _('Created at (newest first)')), ('created_at', _('Created at (oldest first)')),
         ('-updated_at', _('Updated at (newest first)')), ('updated_at', _('Updated at (oldest first)')),
     ])
 
     def __init__(self, country, *args, **kwargs):
         self.country = country
```
longhotsummer, line 154:
```suggestion
```

longhotsummer: superclass does this

indigo_app/urls.py
```diff
@@ -48,6 +48,7 @@
     path('places/<str:place>/localities', places.PlaceLocalitiesView.as_view(), name='place_localities'),
 
     path('places/<str:place>/tasks', tasks.TaskListView.as_view(), name='tasks'),
+    path('places/<str:place>/tasks/facets', tasks.TaskFacetsView.as_view(), name='tasks_facets'),
```
longhotsummer, line 51: Pls apply 30 seconds of caching, like we do for `places/<str:place>/works/facets`

indigo_app/views/tasks.py
```diff
@@ -64,7 +65,8 @@ def get(self, request, *args, **kwargs):
 
         # initial state
         if not params.get('state'):
-            params.setlist('state', ['open', 'assigned', 'pending_review', 'blocked'])
+            pass
```
longhotsummer, line 68: The idea behind this is we don't want to show completed tasks by default, otherwise it would spam the listing.

indigo_app/views/tasks.py
```diff
@@ -90,6 +105,8 @@ def get_context_data(self, **kwargs):
         context['total_tasks'] = self.get_base_queryset().count()
 
         context["taxonomy_toc"] = TaxonomyTopic.get_toc_tree(self.request.GET)
+        context["work_facets"] = self.form.work_facets(self.get_base_queryset(), context['taxonomy_toc'], [])
```
longhotsummer, line 108: In the place work view, we don't calculate facets on the main page of the page. Instead, we render a placeholder for the facet list and then load the facets via htmx. This means we keep that page fast to load.

actlikewill:
@longhotsummer @goose-life this bit here throws off the counts for the filters, I'm not sure if we should change it.
https://github.com/laws-africa/indigo/blob/bafcbd2b9904cdb55b38b0f2e5b47952813fd8fc/indigo_app/forms/tasks.py#L174-L184

indigo_app/views/tasks.py
```diff
@@ -90,6 +105,8 @@ def get_context_data(self, **kwargs):
         context['total_tasks'] = self.get_base_queryset().count()
 
         context["taxonomy_toc"] = TaxonomyTopic.get_toc_tree(self.request.GET)
```
longhotsummer, line 107: We no longer need the `all_topics` support, which was only used for the task view, so we can drop that parameter completely and remove the support, rather than defaulting it to True.

longhotsummer:
The key outstanding task here is to fix the "status" facet so that counts are correct, which I'll do today.

longhotsummer:
We still use columns in the work task view (and in the place work list), but we can fix that later.

actlikewill:
@longhotsummer I'm happy to merge this PR as is without calculating the facet counts separately. We can observe how it behaves in production and if the loading is too slow we can pull out the facet calculations.

longhotsummer (approved):

PR #2070, opened 2024-04-05
Task filters changes

longhotsummer:
* remove old global "priority tasks" and taxonomy-based task listings, as those are supported by this new listing now 
* fixes facet counts for state by splitting out two concepts we previously combined: state and assignment
* include "assigned to" in task table
* country faceting for all tasks view
* some icons

https://www.loom.com/share/e97e734818ee4e97a0f827359a6fed3c


actlikewill (approved):
great, I like the state icons on the facets

PR #2073, opened 2024-04-09
Edit commencements

goose-life:
https://www.loom.com/share/8656bc632bfd4effa0b85c2e9d8138b1?sid=964cfb3b-58b9-4702-922e-0deea5b87649

To do:

- [x] figure out double / triple modal-backdrop
- [x] ~disable buttons when any form is open on the page (using `resp.headers['HX-Trigger']` on WorkCommencementEditView?)~


indigo_app/templates/indigo_api/commencement_form.html
```diff
@@ -0,0 +1,145 @@
+{% load i18n %}
+
+<form
+    hx-post="{% url 'work_commencement_edit' frbr_uri=work.frbr_uri pk=commencement.id %}"
+    hx-target="body"
```
longhotsummer, line 5: As suggested, you can safely re-render the whole page, so these would change to `method="post" action="(url)"`

longhotsummer: and the success handler for the UpdateView would redirect to the commencements detail page for the work.

longhotsummer (dismissed):

goose-life:
image

PR #2074, opened 2024-04-10
Work form commencements

goose-life:
https://www.loom.com/share/83a9d562a740495ab6a2e41b119f11db?sid=668d8fc4-ccd3-4ae4-ac7c-e1bbec90e4fe


goose-life:
I've reverted requiring a date, because a commencement without a date could be added in the advanced view and then the work form can never be saved again after that

longhotsummer (approved):

PR #2080, opened 2024-04-18
make it less work to include extra urls in the content api

longhotsummer:
Previously it has been difficult to add new routes into the content API. This change makes it much simpler:

1. the content api patterns can be adjusted by other apps during `app_ready` by adjusting the content of the `router.router_patterns` variable
2. other apps don't need to re-implement the root content API views, and the new schema views, at all now
3. other apps have a mechanism to specify to the schema where the API lives and what URL patterns to use, if they need it, through `settings.py`

### DIFF
--- a/indigo/settings.py
+++ b/indigo/settings.py
@@ -151,7 +151,17 @@ INDIGO = {
     'EXTRA_DOCTYPES': {},
 
     # disable entire site except for superusers and the content API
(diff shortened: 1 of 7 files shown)


goose-life (approved):
cool!

goose-life (approved):

PR #2081, opened 2024-04-18
French translation fixes

goose-life:


longhotsummer (approved):
assuming the .mo changes are because they have been re-compiled (necessary)

PR #2086, opened 2024-04-23
remove build dir, it should not be committed

longhotsummer:


goose-life (approved):

longhotsummer:
These tests pass locally, and I can't see any reason why this should not work, so I'm going to be bold and merge.

PR #2089, opened 2024-04-25
fix django-allauth dependency

longhotsummer:
see https://github.com/pennersr/django-allauth/pull/3758 and https://github.com/pennersr/django-allauth/discussions/3757


goose-life (approved):

longhotsummer:
Oooh, this is why the tests failed in the translations PR. Allauth have changed their urls

Issue #2104, opened 2024-05-22
Citator is being confused by mixed citations


longhotsummer:
eg. this text has provisions and judgments cited. With the last 150 chars removed, the refs to section 47(1)(e) are matched. With that tail text, though, those provisions aren't matched.

```
'THERON J\n12\ncustodial sentence has been remitted by the President. Further, it raises the question\nof the role of the Commission in deciding eligibility for membership of the\nNational Assembly.\n[29] Is it in the interests of justice to grant leave to appeal? The consequences of\nthe contempt judgment, and in particular the effect of a conviction and sentence on\neligibility for the National Assembly, is of public interest. The general public needs\nto know if candidates on a party list are eligible to be members of the\nNational Assembly. The issues raised in this appeal “transcend the narrow interests\nof the litigants and implicate the interest of a significant part of the general public”.25\nThere needs to be finality on Mr Zuma’s eligibility to stand in the election to be held\non 29 May 2024.\n[30] An appeal to the Supreme Court of Appeal and, thereafter, possibly to this\nCourt is not feasible before 29 May 2024. This Court also has the benefit of the\njudgment of the Electoral Court, a specialist court. There are no material disputes of\nfact and the appeal turns on narrow questions of law. As will soon become plain in\nthis judgment, the appeal has reasonable prospects of success. In these\ncircumstances, a direct appeal is in the interests of justice and leave to appeal must\nbe granted.\nInterpretation of section 47(1)(e) of the Constitution\n[31] Section 47 of the Constitution deals with eligibility to be a member of the\nNational Assembly. This section states:\n“47. Membership\n(1) Every citizen who is qualified to vote for the National Assembly is eligible\nto be a member of the Assembly, except—\n. . .\n25 Paulsen v Slip Knot Investments 777 (Pty) Limited [2015] ZACC 5; 2015 (3) SA 479 (CC); 2015 (5) BCLR\n509 (CC) at para 26.'
```

Without the tail:

```
>>> m.call_citator({"frbr_uri": doc.expression_frbr_uri, "format": "text", "body": pages[11][:-120], "citations": []})['citations']
[...]

PR #2105, opened 2024-05-22
provision ref citation fixes

longhotsummer:
* ensure citations are sorted by position, so that we find the right one when looking for the target of a provision ref
* don't go too far forward or backwards, or cross sentence boundaries when looking for the target of a citation
* don't generate duplicate citations when extracting from text

Fixes https://github.com/laws-africa/indigo/issues/2104

Fixes https://github.com/laws-africa/indigo/issues/2106


goose-life (approved):
from the test outputs it looks like we're matching fewer references — but I assume this is better than the alternative, if we were mis-matching things?

longhotsummer:
> from the test outputs it looks like we're matching fewer references — but I assume this is better than the alternative, if we were mis-matching things?

we were generating duplicates in some cases, and matching incorrectly in others. So yes, slightly fewer.

PR #2112, opened 2024-06-05
Approved relationships

goose-life:


indigo_api/models/works.py
```diff
@@ -34,12 +34,14 @@ class WorkManager(models.Manager):
 
     def get_queryset(self):
         # defer expensive or unnecessary fields
-        return super(WorkManager, self) \
-            .get_queryset() \
+        return super().get_queryset() \
             .select_related('updated_by_user', 'created_by_user', 'country',
                             'country__country', 'locality', 'publication_document') \
             .prefetch_related('commencements')
```
longhotsummer, line 40: commencements are potentially huge, I'm not sure we ALWAYS want to get them whenever we load a work. Can this prefetch be moved to where it's needed?

indigo_api/models/works.py
```diff
@@ -34,12 +34,14 @@ class WorkManager(models.Manager):
...
 
+    def approved(self):
```
longhotsummer, line 42: this should be WorkQuerySet above, so you don't need to get another queryset.

indigo_api/models/works.py
goose-life: I've opened https://github.com/laws-africa/indigo/pull/2114 for this

longhotsummer: sorry, I don't know how I misread this as a new addition 🤦

longhotsummer (approved):

PR #2113, opened 2024-06-05
nuke the work_unapproved signal (optional)

goose-life:


longhotsummer:
i think we can just keep this in place

PR #2114, opened 2024-06-05
Dont prefetch commencements

goose-life:
> commencements are potentially huge, I'm not sure we ALWAYS want to get them whenever we load a work. Can this prefetch be moved to where it's needed?

I assume I'm missing at least some places where it is needed, but this is a start — feedback welcome!

### DIFF
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -793,16 +793,9 @@ def post_save_commencement(sender, instance, **kwargs):
                     place_code=instance.commenced_work.place.place_code)
 
 
-class AmendmentManager(models.Manager):
-    def approved(self):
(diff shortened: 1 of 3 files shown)


longhotsummer:
The only time this is strictly necessary is when we're dealing with a lot of work objects AND their commencements objects.

Here's the commit to add this prefetch: https://github.com/laws-africa/indigo/commit/64047413e0409e79b68275439b064fa0342a7122

I suspect we can safely just remove it - our listing views don't try to use the commencement data (AFAIK). The HTML views for a document will use them, but only for one work.

I propose just removing it, and then watching for n+1 alerts in sentry.

goose-life:
that's definitely the most commits I've made to remove a single line 😂

longhotsummer (approved):

PR #2120, opened 2024-06-18
don't Unhyphenate text when cleaning up (too many false positives)

goose-life:


longhotsummer (approved):
What are some of the false positives we're seeing? This is a useful cleanup (when it works)

goose-life:
> What are some of the false positives we're seeing? This is a useful cleanup (when it works)

It is, but we have a separate check that highlights `- `s in the editor, and it turns out more people use dashes and friends more improperly than I had dreamed possible 😅 — there's even examples of false positives in the tests (self- incriminating, public- private partnership).

PR #2142, opened 2024-07-18
Fix dodgy popups by using tippy; turn off user popups

longhotsummer:
https://www.loom.com/share/784fec3ed54f4a3a88813871d600e80c


goose-life (approved):
nice!

PR #2143, opened 2024-07-19
Provision references citator handles French

longhotsummer:
* make provision refs grammar choose different parse trees based on the current language; this means that only one language is used when parsing, so languages can have terms that clash
* add basics of French provision lookups

### DIFF
--- a/indigo/tests/test_provision_refs.py
+++ b/indigo/tests/test_provision_refs.py
@@ -3,7 +3,6 @@ from unittest.mock import patch
 
 from django.test import TestCase
 from lxml import etree
-import lxml.html
 
 from cobalt import AkomaNtosoDocument, FrbrUri
 from docpipe.matchers import ExtractedCitation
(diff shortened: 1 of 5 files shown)


indigo/tests/test_provision_refs.py
```diff
@@ -1770,15 +1825,32 @@ def test_target_truncated(self):
...
 
-        result = parse_provision_refs("Afdeling 2 daarvan")
+        result = parse_provision_refs("Afdeling 2 daarvan", "afr")
         self.assertEqual("thereof", result.target)
         self.assertEqual(result.end, 18)
 
-        result = parse_provision_refs("Afdeling 2 van die Wet met kak")
-        self.assertEqual("of", result.target)
-        self.assertEqual(result.end, 15)
+        result = parse_provision_refs("Afdeling 2 van die Wet met kak", "afr")
```
goose-life, line 1837: 😂

goose-life (approved):

indigo/tests/test_provision_refs.py
longhotsummer: translation courtesy of copilot 🤣

longhotsummer:
i think this is good enough to ship and start testing with actual data; FRA will have more edge cases, but this will handle basics

goose-life (approved):

PR #2144, opened 2024-07-19
Fix publication document bug

goose-life:


indigo/tasks.py
```diff
@@ -83,16 +84,22 @@ def make_input_task_file(self, task, use_publication_document=False):
...
+                self.save_input_file_using_publication_document_info(input_file, task, publication_document)
             elif publication_document.file:
-                input_file.file = publication_document.file
-            input_file.filename = publication_document.filename
-            input_file.mime_type = publication_document.mime_type
-            input_file.size = publication_document.size
-            input_file.task_as_input = task
-            input_file.save()
+                # make a copy of the publication document's file, since the task form can be edited and the file deleted
+                with open(publication_document.file.path, 'rb') as f:
```
longhotsummer, line 90: This doesn't work with cloud storage. Instead of opening it try just giving File(publication_documemt.file, ...)

longhotsummer (dismissed):

PR #2154, opened 2024-07-26
Commencement bug

goose-life:
Shows an error rather than breaking when we're about to create a duplicate commencement

https://www.loom.com/share/5d1dac5091284b83956b38bbf27ed6fd?sid=6edf4be3-835a-4775-965e-16541b0afd3d

### DIFF
--- a/indigo_app/forms/works.py
+++ b/indigo_app/forms/works.py
@@ -581,12 +581,10 @@ class CommencementForm(forms.ModelForm):
             raise ValidationError(_("Cannot specify all provisions, and a list of provisions."))
 
         # don't try to save a duplicate commencement, we'll get an IntegrityError
-        duplicate_commencement = self.work.commencements.exclude(
-            pk=self.instance.pk
-        ).filter(
+        if self.work.commencements.exclude(pk=self.instance.pk).filter(
             commencing_work=cleaned_data.get('commencing_work'),
-            date=cleaned_data.get('date')).first()
-        if duplicate_commencement:
+            date=cleaned_data.get('date')
+        ).exists():
             raise ValidationError(_("A commencement at this date and with this commencing work already exists."))
 
         return cleaned_data
--- a/indigo_app/forms/works.py
+++ b/indigo_app/forms/works.py
@@ -1,6 +1,5 @@
 import re
 from dataclasses import dataclass, field
-from datetime import date
 from functools import cached_property
 from typing import List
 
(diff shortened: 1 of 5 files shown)


indigo_app/forms/works.py
```diff
@@ -576,11 +575,22 @@ def clean_all_provisions(self):
...
-        if self.cleaned_data.get('all_provisions') and self.cleaned_data['provisions']:
+        if cleaned_data.get('all_provisions') and cleaned_data['provisions']:
             raise ValidationError(_("Cannot specify all provisions, and a list of provisions."))
 
+        # don't try to save a duplicate commencement, we'll get an IntegrityError
+        duplicate_commencement = self.work.commencements.exclude(
+            pk=self.instance.pk
+        ).filter(
+            commencing_work=cleaned_data.get('commencing_work'),
+            date=cleaned_data.get('date')).first()
```
longhotsummer, line 588: minor: use `.exists()` when you want to check if it exists without actually fetching any data.

indigo_app/views/works.py
```diff
@@ -405,18 +406,34 @@ def get_permission_required(self):
             return ('indigo_api.delete_commencement',)
         return ('indigo_api.view_commencement',)
 
+    def get(self, request, *args, **kwargs):
+        if request.htmx.target == 'commencement-new':
+            # delete the new commencement and reload the page
+            work_commencements_url = self.get_work_commencements_url()
+            self.do_delete()
```
longhotsummer, line 413: it's dangerous to do a deletion operation on a GET call

longhotsummer: My understanding is we do this when we're "creating" a new commencement and want to cancel. Since "creating a new commencement" is actually "create and edit", we need to delete if we cancel. Perhaps we change the experience of "create" to "create and edit" explicitly, so then "cancel" becomes "delete".

Linked to this: what's the experience if I click "new commencement" and then refresh the page -- presumably an empty commencement? What's the experience if I then click "edit"?

goose-life:
https://www.loom.com/share/031821fd625645f49ac6501fac1c2bca

longhotsummer:
Interesting edge cases. I suspect we're trying to be too clever. Here are some thoughts to simplify the logical model we implement and we want the user to have in their mind.

1. if you can't add an amendment because a blank one already exists, show that as a warning and don't show the add button at all. So we don't let the user do something they can't do.
2. when clicking "Add" and it fails because our page was out of date and a blank one already exists, the page should reload with an error message at the top (django messages error). Effectively: you tried to do something, you can't, here's the page again with an error message at the top. This is an edge case of 1.
3. if I click "Add" and then cancel, that shouldn't be different to "Edit" and cancel: that entry's card (form) is re-loaded with htmx. it will mean that the Add button no longer appears, but that's by design now: a blank item was created so you can't create another one.

All along, the mental model is consistent:

1. adding creates a new, blank model and immediately makes it editable.
2. only one blank model can exist at a time.
3. cancel refreshes just that element being edited
4. saving refreshes the whole page because the timeline may change
5. deleting refreshes the whole page for the same reason

goose-life:
https://www.loom.com/share/6f651f9878ce47bb9a686a140cf615ee

goose-life:
data-confirm removal:
https://www.loom.com/share/ac4d7483360a43d0b678704cb2433636

longhotsummer (approved):

PR #2163, opened 2024-08-12
django 4.2 and updated dependencies

longhotsummer:
* django 4.2
* update other dependencies
* include test reports for pull requests
* remove pinax badges and replace with local models
* remove legacy javascript packages and bundle dynamically with webpack -- this moves modern dependencies (eg. bootstrap) into a normal npm + webpack flow, rather than including them in the repo
* remove bootstrap-datepicker and use standard date input instead


longhotsummer:
@actlikewill and @goose-life this is a good PR to review, because it has substantive changes. It merges into the python 3.10 upgrade branch, which also has the migration to pyproject.toml. So this PR is the best for checking what actual changes are required for Django 4.2.

indigo/tests/bulk_creator/errors.csv
```diff
@@ -1,7 +1,7 @@
 country,locality,title,cap,subtype (blank for Acts),number,year,publication_name,publication_number,assent_date,publication_date,commencement_date,principal (✔),stub (✔),taxonomy,primary_work,commenced_by,amended_by,repealed_by,subleg,commences,commences_on_date,amends,repeals,Ignore (x) or in (✔),frbr_uri,frbr_uri_title,comments,LINKS ETC (add columns as needed)
 ,xyz,Baddie,,,,Typo,,,2020-13-08,Not a date,,x,,,,,,,,,,,,,,,,
 zA,jhb,Testy 2,,,2,2020,,,,2020-06-01,,,x,,,,,,,,,,,,,,,
-ZA,,Bad commencement date,,,45,2013,Government Gazette,36949,2013-10-03,2013-10-21,21 Oct 2013,,x,,,,,,,,,,,,,,,
+ZA,,Bad commencement date,,,45,2013,Government Gazette,36949,2013-10-03,2013-10-21,2013 21 Oct,,x,,,,,,,,,,,,,,,
```
longhotsummer, line 4: newer python versions are smarter about dates, so it needs to be more crazy to break the validator

indigo_api/models/works.py
```diff
@@ -246,10 +246,10 @@ def all_provisions(self):
 
     @cached_property
     def main_commencement(self):
-        return self.commencements.filter(main=True).first()
+        return self.commencements.filter(main=True).first() if self.pk else None
```
longhotsummer, line 249: Django > 4 doesn't allow relationship lookups on a model that is not saved; this tends to happen mostly in testing, so perhaps changing the test is better than this, but the tests now pass

indigo_social/migrations/0008_badgeaward.py
```diff
@@ -0,0 +1,33 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('indigo_social', '0007_models_i18n'),
+    ]
+
+    replaces = [
+        ('pinax_badges', '0001_initial'),
+    ]
```
goose-life, line 17: do we have this here so that the db entries for the old badges get copied over to the new tables, and then in future we would delete both this and the `options` below?

longhotsummer: this tells django to ignore this migration if the pinax_badges one is already done, but in new installations to use this migration.

The actual data still stays in the old table (and our model is configured to use that table), I'm not changing that for now.

indigo_social/models.py
```diff
@@ -78,3 +78,36 @@ def retrieve_profile_photo_on_signup(sender, **kwargs):
...
         pass
+
+
+class BadgeAward(models.Model):
+    user = models.ForeignKey(User, related_name="badges_earned", on_delete=models.CASCADE)
+    awarded_at = models.DateTimeField(auto_now_add=True)
+    slug = models.CharField(max_length=255)
+
+    class Meta:
+        db_table = "pinax_badges_badgeaward"
```
goose-life, line 89: would this also fall away at some point? or is it (and the `options` above) safe to leave in forever?

longhotsummer: we can leave this as is; the only argument for moving the table is to have a table with a different name

indigo_social/models.py
```diff
@@ -78,3 +78,36 @@ def retrieve_profile_photo_on_signup(sender, **kwargs):
...
+        return self
+
+    @property
+    def _badge(self):
+        from indigo_social.badges import badges
+        return badges.registry[self.slug]
+
+    @property
+    def name(self):
+        return self._badge.levels[self.level].name
```
goose-life, line 105: will `self.level` ever be anything? can't this concept now fall away entirely?

indigo_api/models/works.py
goose-life: good to know! we might need to make this change elsewhere too

indigo_social/migrations/0008_badgeaward.py
goose-life: got it, thanks!

goose-life (approved):
I'm not sure that badge levels are still relevant; looks great otherwise!

goose-life (approved):

goose-life (approved):

Issue #2196, opened 2024-08-15
Table pasting functionality in monaco editor is broken


longhotsummer:
It looks related to this: https://github.com/microsoft/monaco-editor/issues/4030

longhotsummer:
The work-around is to disable the pasteAs functionality. It would be great to tie into the editor's actual paste functionality so we could adjust how the paste happens, but I can't see how to do that.

longhotsummer:
The correct solution here is to change indigo-akn to update monacoOptions() to include this option.

PR #2198, opened 2024-08-16
latest indigo-akn to solve pasting bug properly

longhotsummer:


longhotsummer:
Closes https://github.com/laws-africa/indigo/issues/2196

goose-life (approved):

PR #2206, opened 2024-08-23
update reverse_content_api to always return an absolute URL

goose-life:


longhotsummer (approved):
You could be even blunter and just check for `url.startswith("http:") or url.startswith("https:")` and then just use `return settings.INDIGO_URL + url`

PR #2215, opened 2024-08-30
add xml to makemessages extensions; re-ran translation extraction

goose-life:


longhotsummer (approved):

longhotsummer:
you can safely add `accounts/login/code/confirm/\Z` to the list of urls that don't need auth

PR #2228, opened 2024-09-06
saved searches for task and work views

longhotsummer:
This makes it possible to created saved searches for work and task listing views.

Effectively this is a mechanism for a content manager to manage a set of saved searches to share with the editorial team. They apply to the work or task listing views, both the per-country and All views.

The Available Tasks view falls away because it can be replicated and managed more easily (eg. for reviewers, for editors, etc.) using this feature. Our editorial team shows they don't regularly use the Available Tasks view, and rather use the All Tasks view.

https://www.loom.com/share/ff70ed8ad87446f8a7db948670f8d874


goose-life (approved):
nice!

PR #2254, opened 2024-10-03
citator mustn't match subtypes mid-word, or lowercase

longhotsummer:
otherwise application 2 of 2022 is match as "on 2 of 2022"

fixes https://github.com/laws-africa/indigo/issues/2253

### DIFF
--- a/indigo/analysis/refs/base.py
+++ b/indigo/analysis/refs/base.py
@@ -142,7 +142,7 @@ class SubtypeNumberCitationMatcherENG(DocumentPatternMatcherMixin, CitationMatch
     def setup_subtypes(self):
         self.subtypes = [s for s in Subtype.objects.all()]
         subtype_names = [s.name for s in self.subtypes]
-        subtype_abbreviations = [s.abbreviation for s in self.subtypes]
+        subtype_abbreviations = [s.abbreviation.upper() for s in self.subtypes]
 
         # sort, longest first
         subtypes = sorted(subtype_names + subtype_abbreviations, key=len, reverse=True)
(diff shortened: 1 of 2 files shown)


indigo/analysis/refs/base.py
```diff
@@ -160,13 +160,13 @@ def setup_subtypes(self):
         self.pattern_re = re.compile(
             fr'''
                 (?P<ref>
-                    (?P<subtype>{subtypes_string})\s*
-                    (No\.?\s*)?
+                    \b(?P<subtype>{subtypes_string})\s*
+                    ([nN]o\.?\s*)?
                     (?P<num>[a-z0-9-]+)
                     (\s+of\s+|/)
```
goose-life, line 166: we might find that we actually want to support NO X OF Y (and other variations), but let's see? (if that happens we'll need a bit more tweaking because ACT etc also won't match now)

indigo/tests/test_refs.py
```diff
@@ -33,6 +33,8 @@ def test_find_simple(self):
             <content>
               <p>Something to do with GN no 102 of 2012.</p>
               <p>And another thing about SI 4a of 1998.</p>
+              <p>don't match lowercase gn no 102 of 2012.</p>
+              <p>don't match where it's at the end of somethign no 102 of 2012.</p>
```
goose-life, line 37: shouldn't GN be uppercased here to test that it doesn't match mid-word? (minor)

goose-life (approved):

indigo/tests/test_refs.py
longhotsummer: Good point, additional test added.

indigo/analysis/refs/base.py
longhotsummer: Yes, you're right. Let's see how often that happens. In my sampling, "OF" isn't common.

This only applies to sub-types, "ACT" is handled by a different matcher.

PR #2277, opened 2024-11-12
Provision editing

goose-life:


indigo_api/views/documents.py
```diff
@@ -122,6 +122,7 @@ def content(self, request, *args, **kwargs):
 
         if request.method == 'PUT':
             try:
+                # TODO: don't just reset here, integrate the content IFF in provision mode (or always?)
```
goose-life, line 125: how do I hit this breakpoint?

goose-life:
https://www.loom.com/share/0e2085f141914f78b0baa96d0ea3c1a0

longhotsummer:
Nice! The chooser looks great

goose-life:
https://www.loom.com/share/d62443b5c76a4f96bf5a9c27e9fbb47f

goose-life:
https://www.loom.com/share/cda7a6f665cf4e769152f72adb70d04c

indigo_api/models/documents.py
```diff
@@ -544,6 +547,20 @@ def change_date(self, new_date, user, comment=None):
             self.expression_date = new_date
             self.save_with_revision(user, comment=comment)
 
+    def update_provision_xml(self, provision_eid, new_provision_eid, provision_xml):
+        generator = XmlGenerator(self.frbr_uri)
+        xml = etree.fromstring(provision_xml)
+        akn_provision = generator.wrap_akn(xml)
+        portion = Portion(etree.tostring(akn_provision, encoding='unicode'))
+        updated_provision = portion.get_portion_element(new_provision_eid)
```
longhotsummer, line 555: seems odd to have to have the new_provision_id, can't we just use the portionBody first child element?

indigo_api/models/documents.py
```diff
@@ -544,6 +547,20 @@ def change_date(self, new_date, user, comment=None):
...
+        old_provision = self.doc.get_portion_element(provision_eid)
+        elem_parent = old_provision.getparent()
+        elem_parent.replace(old_provision, updated_provision)
+        # fix up the XML: rewrite eids, correct tags etc.
+        updated_xml = generator.post_process(self.doc.main)
+        with_akn_tag = generator.wrap_akn(updated_xml)
+        return etree.tostring(with_akn_tag, encoding='unicode')
```
longhotsummer, line 562: call `reset_xml` here to save the updates into the string, so that the string and self.doc are still aligned.

indigo_api/serializers.py
```diff
@@ -326,6 +332,19 @@ def validate(self, attrs):
             if doc.namespace != AKN_NAMESPACES['3.0']:
                 raise ValidationError(f"Document must have namespace {AKN_NAMESPACES['3.0']}, but it has {doc.namespace} instead.")
 
+        if attrs.get('content') and attrs.get('provision_eid'):
+            # validate the portion content
+            try:
+                xml = attrs['content']
+                ns = AKN_NAMESPACES[DEFAULT_VERSION]
+                xml = f'<akomaNtoso xmlns="{ns}">{xml}</akomaNtoso>'
```
longhotsummer, line 340: rather get the client to provide this as fully valid AKN, so the server doesn't have to make guesses

indigo_api/views/documents.py
longhotsummer: I don't think this is called at all any more; it can be removed (and the `PUT` above) and we'll see when it breaks.

indigo_api/views/documents.py
```diff
@@ -354,7 +356,11 @@ def post(self, request, document_id):
             doc.frbr_uri = frbr_uri
             xml = doc.to_xml(encoding='unicode')
 
-        return Response({'output': xml})
```
longhotsummer, line 357: ideally the client should be extracting the eId and doing with it whatever it needs to, so these changes can fall away

indigo_app/static/javascript/indigo/models.js
```diff
@@ -61,10 +61,12 @@
       // originally triggered by a content change
       if (options && options.fromContent) return;
 
-      // rewrite all eIds before setting the content
-      new indigoAkn.EidRewriter().rewriteAllEids(this.xmlDocument.documentElement);
-      // rewrite all attachment FRBR URI work components too
-      new indigoAkn.WorkComponentRewriter().rewriteAllAttachmentWorkComponents(this.xmlDocument.documentElement);
+      if (Indigo.Preloads.provisionEid === "") {
```
longhotsummer, line 64:
```suggestion
      if (!Indigo.Preloads.provisionEid) {
```

longhotsummer, line 64: :( always needs to be done, in case you quick-edit something and give it a duplicate eid

indigo_app/static/javascript/indigo/models.js
```diff
@@ -185,6 +187,8 @@
       // changes in a single revision on the server.
       // We do this by delegating to the document object.
       this.document.attributes.content = this.get('content');
```
longhotsummer, line 189: when in provision mode, wrap this in appropriate tags to make it a fully valid akn portion

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -576,6 +578,11 @@
           .fail(fail);
       }
 
+      // TODO: a better way of reloading the page (will redirect to provision chooser for now)
+      if (Indigo.Preloads.provisionEid !== Indigo.Preloads.newProvisionEid) {
```
longhotsummer, line 582: this will change to check the expected against the actual in the XML, rather than using preloads.newProvisionEid

longhotsummer:
Nice! This is now much simpler, actually.

longhotsummer (approved):

longhotsummer (dismissed):

longhotsummer (approved):

PR #2281, opened 2024-11-15
use commencement date if publication date is missing

goose-life:


indigo_api/models/works.py
```diff
@@ -703,7 +703,7 @@ def public_taxonomy_topics(self):
 
     def get_import_date(self):
         """Return the date at which content should be imported for this work; may be None."""
-        import_timeline_dates = [self.publication_date] if self.publication_date else []
+        import_timeline_dates = [self.publication_date] if self.publication_date else [self.commencement_date] if self.commencement_date else []
```
longhotsummer, line 706: The duplicated `if` makes this a bit awkward, I think this is simpler:

```python
import_timeline_dates = [x for x in [self.publication_date, self.commencement_date] if x]
```

goose-life: hmmm. we only want to use the commencement date if there isn't a publication date, because it returns the latest of the two. So:
```
import_timeline_dates = [self.publication_date] if self.publication_date else []
if not import_timeline_dates:
    import_timeline_dates = [self. commencement_date] if self. commencement_date else []
```

longhotsummer: that's clearer

longhotsummer (approved):

PR #2297, opened 2024-11-27
Draft provision editing

goose-life:


indigo_api/models/documents.py
```diff
@@ -544,6 +548,33 @@ def change_date(self, new_date, user, comment=None):
...
+        provision = etree.fromstring(provision_xml)
+        # TODO: a better way to discard the wrapping akn tag (using xpath / getchildren()?)
+        actual_provision = list(provision)[0]
+
+        # TODO: don't need this, just for debugging for now
+        actual_provision_str = etree.tostring(actual_provision, encoding='unicode', pretty_print=True)
+
+        main_tree = self.doc.main
+        # TODO: a better way to grab the (only) element with the given provision_eid
+        elems = main_tree.xpath(f".//a:*[@eId='{provision_eid}']", namespaces={'a': self.doc.namespace})
```
longhotsummer, line 561: you can use `self.doc.get_portion_element()` which is a method on `cobalt.StructuredDocument`

longhotsummer: This is what we use when fetching portions for popups etc.

indigo_api/serializers.py
```diff
@@ -389,8 +394,11 @@ def update_document(self, document, validated_data=None):
         # Document content must always come first so it can be overridden
         # by the other properties.
         content = validated_data.pop('content', None)
+        # TODO: include provision_eid
+        provision_eid = validated_data.pop('provision_eid', None)
         if content is not None:
-            # TODO: this is where we want to intercept the XML update (as well?)
+            if provision_eid:
+                content = document.update_provision_xml(provision_eid=provision_eid, provision_xml=content)
```
longhotsummer, line 401: I think `update_provision_xml` should call `reset_xml` itself -- so it updates the full internal representation of the document.

indigo_api/views/documents.py
```diff
@@ -356,7 +356,7 @@ def post(self, request, document_id):
             doc.frbr_uri = frbr_uri
             xml = doc.to_xml(encoding='unicode')
 
-        return Response({'output': xml})
+        return Response({'output': xml, 'provisionEid': serializer.validated_data.get('provision_eid')})
```
longhotsummer, line 359: JSON responses usually use the `provision_eid` style member names. It's good to standardise on that rather than having the request and the response use different names.

goose-life:
closing in favour of https://github.com/laws-africa/indigo/pull/2277

PR #2304, opened 2024-12-10
Client-side bluebell parsing with pyodide

longhotsummer:
Optionally use pyodide to run bluebell in the client. This is about 3 times faster (especially for large documents) than going to the server and parsing.

Note that pyodide has lxml >= 5 built in, but currently we use lxml < 5 on the server. This doesn't impact bluebell but we should align them just to be safe.


longhotsummer:
We now use lxml 5 and this is ready for review.

goose-life (approved):
wow, nice! so simple 😍

PR #2308, opened 2024-12-11
prevent long lines of non-breakable content overflowing

longhotsummer:
the document area

see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#break-word

before:

image

after:

image


longhotsummer:
I don't know if this should be applied on la-akoma-ntoso in general in our standard stylesheets. I suspect it could safely be done.

PR #2320, opened 2025-01-08
Live editing

### LINKED ISSUE #2313: Clicking on a ToC item should trigger a scroll to the top of the pane
This is new since the changes to the editor

longhotsummer:
* allows live editing
* replaces explicit communication of "the current element changed" to using MutationObservers, which allows arbitrary XML tree changes and all components react independently and correctly
* don't try to keep text version and XML dom in sync; only serialise to text when saving
* removes some use of jquery
* removes DocumentEditorView in favour of functionality pushed into SourceEditorView and pulled up into DocumentDetailView
* simplifies document saving to use async which is easier to understand 
* change from jshint to eslint, which is better supported by pycharm
* explicitly re-write eIds after major changes to the DOM, rather than always doing it even for minor DOM changes
* highlight element being quick-edited and scroll it into view 
* run linters asynchronously so that slow ones don't block the UI; also allow popup enrichments for linter issues to be rendered asynchronously
* dynamically adjust live updates for large documents (anything over 100k takes about 1s to parse, which is too slow)
* move "update" and "cancel" buttons to be closer to the text editing box, since that's what they impact
* fix #2313 

image

image


longhotsummer:
Fewer bugs, linters and refs: https://www.loom.com/share/c83a064dfb6a475bb8849b161afeaf3a

longhotsummer:
https://www.loom.com/share/51d5a6c3a176430b9eae18a2c6588354

longhotsummer:
I'm not yet sure how well this performs for very long texts

longhotsummer:
Move update/cancel buttons, hide text editor when table editing starts, hide main save button when editing is under way https://www.loom.com/share/0d5ff9f1b09440ce8b373d576def3c05

longhotsummer:
Speed improvements to linters, and disabling live updates for large documents https://www.loom.com/share/612fc0256c0846648de6948fceb36941

indigo_app/static/javascript/indigo/models.js
```diff
@@ -40,27 +45,86 @@
...
+      }
+
+      // we don't care about attribute or character data changes elsewhere in the document
+    },
+
     contentChanged: function(model, newValue, options) {
-      // don't bother updating the DOM if the source of this event
-      // is already a change to the DOM
-      if (options && options.fromXmlDocument) return;
+      let root = null;
```
goose-life, line 98: initially setting the content triggers this, so the Save button is enabled from the start

goose-life: But changing `this.documentContent.trigger('sync');` to `this.document.trigger('sync');` at 
https://github.com/laws-africa/indigo/pull/2320/files#diff-f541713a19b39bec6d0c84799a3285e34a13cf02ec2e2a471ba9d7973ddd80a7L220 
fixes it (I can't comment there 🤷)

indigo_app/static/javascript/indigo/views/document_xml_editor.js
```diff
@@ -57,13 +66,39 @@ class AknTextEditor {
...
   setXmlElement (element) {
     this.xmlElement = element;
 
+    if (!this.editing) {
+      this.editing = true;
+      this.xmlElementOriginal = element;
+      this.dirty = false;
+    }
+
     if (!this.updating) {
```
longhotsummer, line 95: won't this always be updating = false?

indigo_app/static/javascript/indigo/views/document_xml_editor.js
```diff
@@ -125,33 +179,128 @@ class AknTextEditor {
...
+
+    if (!this.liveUpdates) {
+      let elements;
+      // use a nonce to check if we're still the current save when the parse completes
+      const nonce = this.nonce = Math.random();
+      const btn = this.root.querySelector('.btn.save');
+
+      try {
+        btn.setAttribute('disabled', 'true');
+        elements = await this.parse();
```
longhotsummer, line 221: can we use the same stale check, rather than both nonce and checking the text?

indigo_app/static/javascript/indigo/views/document_editor.js
```diff
@@ -388,32 +379,35 @@
...
-
-      if (!this.tableEditor.canEditTable(table)) {
-        alert($t('This table contains content that cannot be edited in this mode. Edit the table in text mode instead.'));
-      } else {
-        this.tableEditor.editTable(table);
-        // disable other table edit buttons
-        this.$('.edit-table').prop('disabled', true);
+      const tableId = e.currentTarget.dataset.tableId;
+
+      if (this.confirmAndDiscardChanges()) {
```
longhotsummer, line 384: where else is this called that could potentially change the DOM, but we're not waiting for it to settle?

indigo_app/static/javascript/indigo/views/document_xml_editor.js
longhotsummer: one mechanism added

indigo_app/static/javascript/indigo/views/document_xml_editor.js
longhotsummer: yes; removed

indigo_app/static/javascript/indigo/views/document_editor.js
longhotsummer: it only applies in this case because we need the updated XML to be re-rendered to HTML, because table editing happens directly the html.

indigo_app/static/javascript/indigo/models.js
longhotsummer: fixed

indigo_app/static/javascript/indigo/models.js
```diff
@@ -40,27 +45,86 @@
...
+      if (mutation.type === 'childList') {
+        if (mutation.removedNodes[0] === element) {
+          // the element has been replaced
+          return 'replaced';
+        }
+
+        const ownerDocument = target.nodeType === Node.DOCUMENT_NODE ? target : target.ownerDocument;
+        if (!ownerDocument.contains(element)) {
+          // the change removed xmlElement from the tree
+          return 'removed';
```
goose-life, line 85: This is returned when the editor is open, a change has been made but not saved, and some document analysis is run:
- The editor is then closed and we scroll to the top of the selected ToC element or quick-edited element
- The italics terms or whatever are applied
- AND the change made in the editor is kept, but this isn't easy to see or obvious if the quick-edited provision is out of sight.

Probably we should discard changes in the editor first (with confirmation), or disable the analysis menu while the editor is open. In the current world, the editor is closed and the unsaved changes are lost, but the user isn't warned.

longhotsummer: Hmm, good point. Analysis replaces the whole document. Any analyser view that does that should first call:

```
if (Indigo.view.sourceEditorView.confirmAndDiscardChanges()) {
...
}
```

longhotsummer: I'll make this change

longhotsummer: done

indigo_app/static/javascript/indigo/views/document_defined_terms.js
```diff
@@ -28,18 +28,21 @@
 
     /** Find defined terms in this document */
     findTerms: function() {
-      var terms = {};
-
-      return _.map(this.model.xmlDocument.querySelectorAll('def'), function(elem) {
-        return elem.textContent;
-      }).sort();
+      const terms = [];
```
goose-life, line 31: same question here about `const` vs `let`

indigo_app/static/javascript/indigo/views/document_xml_editor.js
```diff
@@ -3,18 +3,25 @@
  * text into XML, and handling text-based editor actions (like bolding, etc.).
  */
 class AknTextEditor {
-  constructor (root, document, onElementParsed) {
+  constructor (root, document, onSave, onDiscard) {
     this.root = root;
     this.document = document;
+    this.onSave = onSave;
+    this.onDiscard = onDiscard
```
goose-life, line 10:
```suggestion
    this.onDiscard = onDiscard;
```

indigo_app/static/javascript/indigo/views/document_xml_editor.js
```diff
@@ -125,33 +165,137 @@ class AknTextEditor {
...
+    if (this.xmlElement) {
+      if (elements && elements.length) {
+        // regular update
+        this.document.content.replaceNode(this.xmlElement, elements);
+      } else if (this.xmlElement.parentNode.tagName === 'portionBody') {
+        // we can't delete the whole provision in portion mode
+        alert($t('You cannot delete the whole provision in provision editing mode.'));
+      } else if (confirm($t('Go ahead and delete this provision from the document?'))) {
+        // remove element
+        this.document.content.replaceNode(this.xmlElement, null);
```
goose-life, line 266: would removing it rather than replacing it work better? this also can't be undone by clicking 'Cancel' after confirming deleting the provision.

longhotsummer: i've pushed a fix so that `getMutationChanged()` returning `removed` in a case like this. Under the hood `replaceNode(..., null)` lands up calling `.remove()`.

goose-life (dismissed):

PR #2330, opened 2025-01-22
Provision-level PiT comparison

goose-life:


indigo_api/views/documents.py
```diff
@@ -433,10 +435,23 @@ def post(self, request, document_id):
...
+        provision_eid = serializer.validated_data.get('provision_eid')
+        if provision_eid:
+            portion = remote_doc.get_portion(provision_eid)
+            # the same structure as the 'xml' we're getting from the browser: akn/portion/portionBody/element
+            remote_xml = etree.tostring(portion.root, encoding='unicode')
+            remote_doc.work.work_uri.doctype = 'portion'
+            remote_doc.content = differ.preprocess_xml_str(remote_xml)
+
+        else:
+            # full document mode
```
longhotsummer, line 453: Why does this call decode but the call a few lines up doesn't?

longhotsummer: It looks like at the end of the day that cobalt's `AkomaNtosoDocument` class will encode a `str` back into `bytes`, and `preprocess_xml_str` returns `bytes`. So actually you don't need any of the `decode()`s when setting `.content = `.

PR #2331, opened 2025-01-23
Provision editing permissions, links

goose-life:
image

To do:

- [x] Link from document

### DIFF
--- a/indigo_app/templates/indigo_api/document/show.html
+++ b/indigo_app/templates/indigo_api/document/show.html
@@ -175,8 +175,8 @@
       </div>
     </li>
     {% endverbatim %}
-    {% if provision_eid %}
-      <li class="breadcrumb-item"><a href="{% url 'choose_document_provision' document.pk %}">{{ provision_eid }}</a></li>
+    {% if provision_eid or perms.indigo_api.publish_document %}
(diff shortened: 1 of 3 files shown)


longhotsummer:
I think it would make more sense to have the links in the same line as the current document links, like the reader mode links. Perhaps before the reader link?

goose-life:
> I think it would make more sense to have the links in the same line as the current document links, like the reader mode links. Perhaps before the reader link?

I thought you might say that 😅 it'll involve a change to the Reader at https://github.com/laws-africa/indigo-la-tradition/blob/master/indigo_la_tradition/js/components/DocumentReader/index.ts#L172 but I think you're right :)

goose-life:
> Why not add a named block to `_work_points_in_time.html` that is inside the `div` on line 72, then indigo and other packages can override that in templates (rather than in JS) to add additional links.

oops too late I already tweaked the query selector, PR incoming 🙈

goose-life:
Provision mode:
image

Whole document:
image

goose-life:
Without permission:
image

image

image

longhotsummer (approved):

goose-life:
> Why not add a named block to `_work_points_in_time.html` that is inside the `div` on line 72, then indigo and other packages can override that in templates (rather than in JS) to add additional links.

I assumed there was a reason we did this differently than all our other overrides and I didn't want to rush into getting rid of the whole ReaderWorkOverviewShim, but I do now think this would be safe to do! I'll update the two PRs quick

longhotsummer (approved):

Issue #2329, opened 2025-01-22
As an editor I want to quick-edit an element from the Entire document as quickly as from its direct parent


goose-life:
With live editing, when the 'Entire document' is selected in the ToC and an element is quick-edited, it's (almost?) as slow as editing the entire document while previewing changes.

Similarly, a Chapter is a bit slower than a section.

https://www.loom.com/share/e23e6c50e04c4b919547ab9a6dabb776

The speed of the preview should ideally be affected by the size of the element being edited, rather than the size of the element being rendered. But if that's impossible, we should rather select whether the preview is on by default based on the size of the element being rendered.

goose-life:
@longhotsummer this is a new one based on editor feedback since we deployed live editing last week.

Should we pull it into this sprint?

longhotsummer:
We'll need to know to re-render only the quick-edited portion, which isn't too tricky. Pull it in if you think you'll get to it and the priority warrants it.

longhotsummer:
This is done.

PR #2332, opened 2025-01-24
Better re-rendering on quick-edit

longhotsummer:
* This saves us from re-rendering an entire document when quick editing just a portion.
* renderCoverpage uses async

In the common case of a quick-edit without an eid change, the quick-edited element is replaced.

In the other case where the eid changes or for some reason another portion of the tree changes, then the target of the mutation (If chp_1__sec_1 is quick edited, then the mutation target will be chp_1) is re-rendered.

This also fixes a bug where changing the eid and then deleting the portion would nuke the whole document.

https://www.loom.com/share/a64501a5e30d4118a3f12b3bf90d80d9

### DIFF
--- a/indigo_app/static/javascript/indigo/models.js
+++ b/indigo_app/static/javascript/indigo/models.js
@@ -84,8 +84,7 @@
           return 'replaced';
         }
 
-        const ownerDocument = target.nodeType === Node.DOCUMENT_NODE ? target : target.ownerDocument;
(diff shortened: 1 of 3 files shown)


longhotsummer:
I've made this smarter to handle the common case of editing just the quick edited portion.

longhotsummer:
Fixes #2329

goose-life (approved):
nice!

PR #2340, opened 2025-01-31
Facets

### LINKED ISSUE #2325: As an editor I want the filters ordered a bit more meaningfully
Specifically, 
- the stub, principal, and primary fields should be next to each other (cf. commencement, amendment, consolidation, and repeal lower down)
- the date field should probably move down (or up?) — it's currently at the best position for accidentally scrolling through the years instead of seeing more filters 😅 

Image

goose-life:
closes #2325 

https://www.loom.com/share/ecab4d806dd942fc928b559508599c82

### DIFF
--- a/indigo_app/forms/works.py
+++ b/indigo_app/forms/works.py
@@ -1201,17 +1201,17 @@ class WorkFilterForm(forms.Form, FormAsUrlMixin):
     def work_facets(self, queryset, taxonomy_toc, places_toc):
         work_facets = []
         self.facet_subtype(work_facets, queryset)
-        self.facet_work_in_progress(work_facets, queryset)
         self.facet_principal(work_facets, queryset)
-        self.facet_frbr_date(work_facets, queryset)
         self.facet_stub(work_facets, queryset)
-        self.facet_tasks(work_facets, queryset)
-        self.facet_publication_document(work_facets, queryset)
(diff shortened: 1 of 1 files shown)


indigo_app/forms/works.py
```diff
@@ -1414,8 +1412,8 @@ def facet_taxonomy(self, taxonomy_tree, qs):
         qs = self.filter_queryset(qs, exclude="taxonomy_topic")
         # count works per taxonomy topic
         counts = {
-            x["taxonomy_topics__slug"]: x["count"]
-            for x in qs.values("taxonomy_topics__slug").annotate(count=Count("taxonomy_topics__slug")).order_by()
+            x: qs.filter(taxonomy_topics__slug=x).count()
```
longhotsummer, line 1415: this will do a new query for every taxonomy slug, which is too slow.

In what cases is the previous mechanism incorrect?

goose-life: I think all cases, e.g. https://edit.laws.africa/places/za/works?taxonomy_topic=projects-priority-1-highest-recent-updates
The count in the facet is 250 but the works listed are 181
image

goose-life:
https://www.loom.com/share/0b76c2edd96240229413c0dd008fa9d3

longhotsummer (approved):

Issue #2343, opened 2025-02-03
Act citation refs finder includes trailing `)` and doesn't markup constitution


longhotsummer:
`39 of 2004)` in `"Air Quality Act" means the National Environment Management: Air Quality Act, 2004 (Act No. 39 of 2004)`

longhotsummer:
Fixed by https://github.com/laws-africa/docpipe/pull/20

longhotsummer:
The Constitution issue is because a CitationAlias with country = `za` doesn't apply to places `za-*`.

longhotsummer:
The Constitution doesn't match because that matcher wasn't being imported :(

Fixed by https://github.com/laws-africa/indigo/pull/2344

PR #2344, opened 2025-02-03
latest docpipe, use alias and title refs matchers

longhotsummer:
* latest docpipe (which fixes `)` at end of refs)
* correctly import plugins for common work titles


longhotsummer:
Ref https://github.com/laws-africa/indigo/issues/2343

goose-life (approved):

PR #2348, opened 2025-02-06
Counters

goose-life:
this should help reduce the very many mutations especially in schedules, but also in e.g. part_II_2
closes #2342 

https://www.loom.com/share/4791d445b9b743d2973f4abae5c8cb95

PR #2349, opened 2025-02-07
process mutations in bulk

longhotsummer:
When eIds are re-written, we can have hundreds of attribute modification mutations. These are slow when processed individually. Instead, we process them as a group. After we find first mutation we find in the list that impacts a component, we can stop processing the remainder.

In the common case, bulk attribute modifications happen after the tree is modified by a node change; so the first change in the list is a node change. In many cases this causes the editors to update and so the attribute modifications are ignored completely.


longhotsummer:
I think this means we don't have to do the hard work in https://github.com/laws-africa/indigo/pull/2348

goose-life (approved):

PR #2350, opened 2025-02-07
if pyodide fails, then parse with the server

longhotsummer:
This will handle an unexpected failure in python (eg. a memory error) by falling back to server-side parsing.

If the parse fails because the grammar is not matched, it's still shown as an error.


goose-life:
neat!

PR #2352, opened 2025-02-10
support "sub section" in provision refs, strip punct around nums

longhotsummer:
and sub regulation
and sub paragraph

also strip `()` and `º` from nums when looking them up.


goose-life (approved):
nice!

PR #2354, opened 2025-02-10
Handle refs to schedules

longhotsummer:
Handles:

* Schedule 1
* Schedule 1 to 4
* Schedule 1 and 4 (etc.)
* the Schedule

Does not handle:

* the First Schedule (etc.)

It works by looking for an `<attachment>` with an exact match on `<heading>`.

### DIFF
--- a/indigo/analysis/refs/provisions.py
+++ b/indigo/analysis/refs/provisions.py
@@ -113,7 +113,15 @@ def parse_provision_refs(text, lang_code='eng'):
             target = elements[2] if not hasattr(elements[2], 'elements') else None
             return ParseResult(refs, target, elements[3].offset)
 
-        def references(self, input, start, end, elements):
+        def attachment_num_ref(self, input, start, end, elements):
+            ref = ProvisionRef(input[start:end], start, end)
+            return [MainProvisionRef("attachment", ref)]
+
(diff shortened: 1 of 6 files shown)


indigo/analysis/refs/provisions.py
```diff
@@ -361,7 +397,8 @@ class ProvisionRefsMatcher(CitationMatcher):
 
     # this just finds the start of a potential match, the grammar looks for the rest
     pattern_names = '|'.join(ProvisionRefsResolver.element_names.keys())
-    pattern_re = re.compile(fr'\b({pattern_names})\s+(\d|\([a-z0-9])', re.IGNORECASE)
+    the_schedule_patterns = '|'.join(['the schedule', 'die bylaag'])
```
goose-life, line 400: is the French missing here, or how does this part fit in?

goose-life (dismissed):
looks good but I'm not 100% clear on how it all fits together, perhaps we could talk through it if there's time this week?

indigo/analysis/refs/provisions.py
longhotsummer: we don't currently support french at all when doing lookups - only the grammar supports it. I don't want to change that with this PR.

goose-life (approved):

PR #2360, opened 2025-02-13
disable htmx's history support, ensure caching takes htmx into account

longhotsummer:
this fixes related issues:

1. htmx was using its own caching engine to refresh the full page contents when we provided hx-push-url in a response (eg. the work listing page). this caused our scripts to run twice, since htmx stores the full <body> in the cache and re-loads it, without calling the server.
2. turning htmx history off with hx-history="false" means that it won't try to store anything in the cache, instead doing an AJAX request.
3. BUT, our server code thinks that's a partial page load, and so doesn't return a full response. So we turn off ajax htmx history off too.
4. finally, ensure we vary caches on the hx-request and hx-target attributes.

closes https://github.com/laws-africa/indigo/issues/2359

### DIFF
--- a/indigo/settings.py
+++ b/indigo/settings.py
@@ -99,6 +99,7 @@ INSTALLED_APPS = (
 MIDDLEWARE = (
     'django.middleware.security.SecurityMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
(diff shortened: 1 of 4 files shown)


longhotsummer:
See https://htmx.org/docs/#history for some details.

goose-life (approved):

PR #2379, opened 2025-03-04
taxonomy topics form should use pks not slugs

### LINKED ISSUE #2371: as a user I want to easily facet on taxonomy topic without it silently failing
for example, one should be able to select https://edit.laws.africa/places/all/works?taxonomy_topic=subject-areas-per-country-cote-divoire-codes as well as all its children, but currently we send back a too-large HTMX header. 

> the solution is to use numeric PKs for the topics and not slugs.

longhotsummer:
otherwise the querystrings get huge

fixes #2371

### DIFF
--- a/indigo_api/models/works.py
+++ b/indigo_api/models/works.py
@@ -111,7 +111,7 @@ class TaxonomyTopic(MP_Node):
         def fix_up(item):
             item["title"] = item["data"]["name"]
             if new_query:
-                new_query.update({'taxonomy_topic': item['data']['slug']})
+                new_query.update({'taxonomy_topic': item['id']})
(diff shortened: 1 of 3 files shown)


goose-life (dismissed):
nice! 

I can't see where and how the href generated here is used, but should it update to use item['id'] rather than the slug as well?
https://github.com/laws-africa/indigo/blob/master/indigo_api/models/works.py#L114-L115

longhotsummer:
> nice!
> 
> I can't see where and how the href generated here is used, but should it update to use item['id'] rather than the slug as well? https://github.com/laws-africa/indigo/blob/master/indigo_api/models/works.py#L114-L115

Updated - I'm not sure this is even used any more.

PR #2381, opened 2025-03-05
Document task pane

### LINKED ISSUE #2322: Show tasks in the document view
Add a new tab to the document view sidebar that includes tasks for the document (or tasks at the document date). This would bring task context to the document view and allow users to make comments on tasks without leaving the document view.

goose-life:
closes #2322 

https://www.loom.com/share/44bcaefe33bc4aa2a9a50687adf27623 

TODO:

- [x] add submission buttons
- [x] fix scrolling
- [ ] deal with empty comment (disable button if there's no comment? do nothing?)
- [ ] deal with logged-out user submitting task within a document (don't load login page, don't redirect to template partial)


indigo_app/static/stylesheets/_core.scss
```diff
@@ -232,3 +232,8 @@ body.is-staff {
   max-height: 50vh;
   overflow-y: auto;
 }
+
+.scrollable {
+  max-height: 100vh;
```
goose-life, line 237: this is not correcto

indigo_app/templates/indigo_api/document/_tasks.html
```diff
@@ -0,0 +1,24 @@
...
+{% if tasks %}
+  <div class="mt-1 me-1 scrollable">
+    {% for task in tasks %}
+      {% include "indigo_api/document/_task.html" %}
+    {% endfor %}
+  </div>
+{% else %}
+  <div class="card-body mt-1 me-1">
+    <p class="alert alert-info">
+      {% blocktrans with expression_date=document.expression_date|date:'Y-m-d' %}
```
longhotsummer, line 19:
```suggestion
      {% blocktrans trimmed with expression_date=document.expression_date|date:'Y-m-d' %}
```

indigo_app/static/stylesheets/_core.scss
```diff
@@ -232,3 +232,8 @@ body.is-staff {
   max-height: 50vh;
   overflow-y: auto;
 }
+
+.scrollable {
```
longhotsummer, line 236: this is no longer necessary

longhotsummer:
I've pushed a change to fix scrolling. I find it pretty confusing - you have to prevent one of the ancestor containers from trying to grow, to prevent allow an inner one to realise it needs to scroll.

longhotsummer:
First prize is if the normal task view uses the same timeline and button templates. I'm not sure how complex that is -- so that we don't have duplicate ways of showing almost identical information.

goose-life:
100%, working on that

longhotsummer:
An option for the logged-out thing is to change the behaviour on the base authenticated view mixin: if the request is HTMX, then rather than redirecting to the login view simply return a very simple "Permission denie" 403 response.

longhotsummer:
This is looking great - https://github.com/laws-africa/indigo/pull/2383 opened to resolve I think all the outstanding issue. HTMX is awesome.

Please also remove the old "Tasks" link in the menu bar, it's now redundant.

image

goose-life:
Thank you! I've taken it for a spin and it seems all good now 🙌

longhotsummer (approved):

PR #2401, opened 2025-04-09
Repealed verb, note

goose-life:
https://www.loom.com/share/db3a3e6e6028407284e295fd318ab4cb

We may want to 
- serialize the verb and note — changes will be required in cobalt for this
- add the verb and note to the 'repeals made' section of the work form

But I think it's good to go as is.


indigo_api/exporters.py
```diff
@@ -281,18 +281,28 @@ def get_notices(self, document, short=False):
         if short:
             # we only care if it's uncommenced and/or repealed for the running header
             if not work.commenced and work.repealed_date:
-                return _('Not commenced; Repealed')
+                return _('Not commenced') + '; ' + _(work.repealed_verb).capitalize()
```
longhotsummer, line 284: I hope that `work.repealed_verb` is always non-null if `repealed_date` is non-null. I know we can't enforce it in the db.

longhotsummer (approved):

PR #2407, opened 2025-04-21
Chapter number details

### LINKED ISSUE #2372: As a researcher / editor I want to record rich 'Chapter'-related information on a work
Chapter numbers, while not suitable for uniquely identifying works, are in fact used and it is therefore worthwhile to record them.

However, they can change over time in certain jurisdictions, since they are a hangover of treating legislation corpuses as print products.

### Proposal

Instead of a single optional 'Chapter' field on a work under its extra properties, which we have now, model ChapterNumber separately:
- FK to `work`, so a work can have more than one — required, cascade on deletion
- `validity_start_date` — optional
- `validity_end_date` — optional
- `revision_name`, a CharField for e.g. 'Revised Edition, 2000' — optional

Don't do any fancy validation, and don't make any assumptions.

In the work form, in places where Chapter numbers are being used (managed in settings), include a formset for Chapter numbers.

In the timeline, add the ChapterNumber at its `validity_start_date` if it has one (otherwise leave it out).

For the numbered title on a work, 
- if there's only one associated ChapterNumber use that (common case)
- if there are multiple, 
  - use the one with the most recent `validity_start_date` if all have
  - if it's at all non-obvious, don't use any and default to Act NN of YYYY

goose-life:
closes #2372 

https://www.loom.com/share/90a1b68a224c4216b906c4e4abd7ae91


indigo/analysis/work_detail/base.py
```diff
@@ -19,15 +19,21 @@ class BaseWorkDetail(LocaleBasedMatcher):
...
         """ Return a formatted title using the number for this work, such as "Act 5 of 2009".
         This usually differs from the short title. May return None.
         """
         # check chapter first
-        if work.place.settings.uses_chapter and work.properties.get('cap'):
-            # eg. Chapter 2
-            return _('Chapter %(cap)s') % {'cap': work.properties['cap']}
+        chapter_number = self.work_chapter_number(work)
+        if chapter_number:
+            return _('%(name)s %(number)s') % {
```
longhotsummer, line 34: This doesn't really make sense as a translation string -- the translator just sees two variables. Instead, since the name is already translated, just use a normal f-string.

indigo_api/timeline.py
```diff
@@ -219,6 +226,13 @@ def get_timeline(work, only_approved_events=False):
         # even though the timeline is given in reverse chronological order,
         # each date on the timeline is described in regular order: assent first, repeal last
 
+        # chapter number (start of validity)
+        for chapter_number in all_chapter_numbers.filter(validity_start_date=date):
```
longhotsummer, line 230: this goes to the DB each time through the outer for loop. Instead, load all the chapter numbers once and filter them in python.

longhotsummer (changes requested):
looks great - minor changes pls.

As much as I like using htmx and django forms, these formsets are the most awkward part :(

longhotsummer (approved):

PR #2426, opened 2025-07-07
don't support lxml 6 just yet

goose-life:


longhotsummer (commented):
We have utilities here which do strictly require lxml 5 or greater so I think this should stay as is, because it's explicit.

PR #2429, opened 2025-07-24
Basic async diffs

longhotsummer:
A starting point for doing diffs asynchronously by running them in a separate process.


indigo_lib/differ.py
longhotsummer, line 1: this is moved here so that it can be used without having to setup django -- if it lives in indigo.analysis.differ, then the other code in indigo.analysis requires that django is set up first

longhotsummer:
For this to work, we also need to change how gunicorn runs its processes. Once we've deployed and refined that for us, I'll bring the changes in here.

PR #2445, opened 2025-08-26
Defs in toc

goose-life:
video to follow

### DIFF
--- a/indigo/analysis/toc/base.py
+++ b/indigo/analysis/toc/base.py
@@ -155,7 +155,7 @@ class TOCBuilderBase(LocaleBasedMatcher):
 
     toc_elements = [
         # top-level
-        'coverpage', 'preface', 'preamble', 'conclusions', 'attachment', 'component',
+        'coverpage', 'preface', 'preamble', 'conclusions', 'attachment', 'component', 'blockContainer',
     ] + AkomaNtoso30.hier_elements
     """ Elements we include in the table of contents, without their XML namespace.
         Base includes the following from the from the AKN schema:
@@ -194,6 +194,10 @@ class TOCBuilderBase(LocaleBasedMatcher):
 
     non_commenceable_toplevel_elements = set(['coverpage', 'preface', 'preamble', 'conclusions', 'attachment', 'component'])
 
+    definition_types = ['blockContainer']
+    """ List of elements that may contain a definition (and should only be included if they do).
+    """
+
     def table_of_contents_for_document(self, document):
         """ Build the table of contents for a document.
         """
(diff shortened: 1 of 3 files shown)


indigo/analysis/toc/base.py
```diff
@@ -377,6 +388,21 @@ def title_without_type(self, item):
 
         return title
 
+    def get_def_element(self, element):
+        """ Uses the refersTo of an element containing a definition to find the matching <def> inside that element.
+            Returns the result of the xpath, which may be an empty list.
+        """
+        term = element.get('refersTo')
```
longhotsummer, line 395: be defensive and assume that this `blockContainer` could appear anywhere in the document, and may not always be related to a term definition. So guard against this being None.

indigo/analysis/toc/base.py
```diff
@@ -377,6 +388,21 @@ def title_without_type(self, item):
...
+        return element.xpath(f'.//a:def[@refersTo="{term}"]', namespaces={'a': self.act.namespace})
```
longhotsummer, line 396: To be pedantic, this returns a list, not a single element, so you may want to iterate over it here and return the first element. That will naturally return `None` to the caller if there is no element.

indigo/analysis/toc/base.py
```diff
@@ -377,6 +388,21 @@ def title_without_type(self, item):
...
+
+    def definition_title(self, item):
+        """ Generates a title for an element containing a definition.
+            Use the text content of the descendant <def> that matches the refersTo of the containing element.
+        """
+        def_elem = self.get_def_element(item.element)
+        if def_elem:
+            return ''.join(def_elem[0].xpath('.//text()')).strip()
```
longhotsummer, line 404: Use `self.heading_text_path`?

indigo_app/js/components/DocumentTOCView.vue
```diff
@@ -113,6 +113,14 @@ export default {
         return headingText;
       };
 
+      const getDefinitionTitle = (node) => {
+        const term = node.getAttribute('refersTo');
+        const result = this.model.xpath(`.//a:def[@refersTo="${term}"]`, node);
+        if (result.snapshotLength > 0) {
+          return result.snapshotItem(0).textContent;
+        }
```
longhotsummer, line 121: I agree that have it there is useful, so perhaps use something like `(term missing)` if there is no text just to handle that edge case and ensure the user has something clickable?

indigo/analysis/toc/base.py
```diff
@@ -194,6 +194,10 @@ class TOCBuilderBase(LocaleBasedMatcher):
 
     non_commenceable_toplevel_elements = set(['coverpage', 'preface', 'preamble', 'conclusions', 'attachment', 'component'])
```
longhotsummer, line 195: should `blockContainer` be added to this? Otherwise definitions will show up as commenceable things.

longhotsummer (approved):

longhotsummer:
nice!

PR #2446, opened 2025-08-29
Defs into blocks update

goose-life:
https://www.loom.com/share/570e5bce5e49482793243b2c9009ddce

### DIFF
--- a/indigo_za/terms.py
+++ b/indigo_za/terms.py
@@ -30,6 +30,18 @@ class TermsFinderENG(BaseTermsFinder):
         self.term_re = re.compile('|'.join(in_quotes + self.others))
         super().setup(doc)
 
+    def contains_definition_block_container(self, element):
+        # returns True for the first blockContainer that has 'definition' as one of its classes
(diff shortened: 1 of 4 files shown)


indigo_za/terms.py
```diff
@@ -30,6 +30,18 @@ def setup(self, doc):
...
+        # returns True for the first blockContainer that has 'definition' as one of its classes
+        for block_container in element.xpath('.//a:blockContainer[@class]', namespaces=self.nsmap):
+            if 'definition' in block_container.get('class', '').split():
+                return True
+
+    def may_contain_definitions(self, element):
+        base_may_contain = super().may_contain_definitions(element)
+        if not base_may_contain:
+            return self.contains_definition_block_container(element)
+        return base_may_contain
```
longhotsummer, line 43: it's cleaner to do this:

```python
return super().may_contain_definitions(element) or self.contains_definition_block_container(element)
```

Python's short-circuit boolean logic will mean that if the first value in `a or b` is True, then `b` is not evaluated at all since it doesn't matter what the value of `b` is.

PR #2475, opened 2025-10-30
make work.main_commencement a first-class property

longhotsummer:
this is much faster than loading it through querysets.

Fixes INDIGO-LAWSAFRICA-249 and many other N+1 issues, since we use this information in any work listing view that includes badges (which is most of them)


indigo_api/migrations/0056_work_main_commencement.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 4.2.15 on 2025-10-30 12:21
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def backfill_main_commencement(apps, schema_editor):
+    Work = apps.get_model('indigo_api', 'Work')
+    Commencement = apps.get_model('indigo_api', 'Commencement')
+    for work in Work.objects.all():
```
goose-life: shouldn't we be using a db_alias here?

PR #2497, opened 2026-01-16
button to open provision editor on current provision

longhotsummer:
image

this makes it easy to jump to editing just the current provision in the provision editor.


indigo_app/static/javascript/indigo/views/document.js
```diff
@@ -353,6 +354,23 @@
       if (e.originalEvent.detail.pane === 'document-secondary-pane') {
         this.secondaryPaneToggle.classList.toggle('active', e.originalEvent.detail.visible);
       }
+    },
+
+    onTocSelectionChanged: function (selection) {
+      // update provision editor link
+      const a = this.el.querySelector('.document-toolbar-wrapper .open-provision-editor');
```
Jonathan-Ngondi: This DOM query can be cached in the `initialize()` method to avoid multiple DOM queries every time a selection changes.

PR #2563, opened 2026-07-15
fixes plain-text provision target resolution

actlikewill:
## what changed

- skips provision-level citations when resolving plain-text `of ...` targets
- normalizes absolute target URIs before appending nested provision fragments
- adds a self-contained regression test in `indigo/tests/test_provision_refs.py`

## why it changed

A plain-text reference like `paragraph 4(1) of the First Schedule; sections 57(1) and 60(2)` could incorrectly bind the earlier `of ...` reference to the later `57(1)` citation. That produced a bogus third citation for `4` and could generate malformed nested hrefs.

## user impact

Citator/provision matching no longer fabricates an extra citation when an earlier plain-text `of ...` reference is followed by later provision citations in the same fragment.

## root cause

The plain-text forward target lookup considered later citations indiscriminately, including provision-level hrefs. Once a provision href was selected as the target root, the matcher appended another `~/eId` segment and produced an invalid nested provision URI.

## validation

- `python3 manage.py test indigo.tests.test_provision_refs.ProvisionRefsMatcherTestCase.test_markup_text_does_not_bind_earlier_of_reference_to_later_provision_citation`
- `python3 manage.py test indigo.tests.test_provision_refs.ProvisionRefsMatcherTestCase.test_markup_text indigo.tests.test_provision_refs.ProvisionRefsMatcherTestCase.test_markup_text_thereof indigo.tests.test_provision_refs.ProvisionRefsMatcherTestCase.test_markup_text_does_not_bind_earlier_of_reference_to_later_provision_citation`

fixes https://github.com/laws-africa/indigo-lawsafrica/issues/1784

### DIFF
--- a/indigo/analysis/refs/provisions.py
+++ b/indigo/analysis/refs/provisions.py
@@ -743,6 +743,13 @@ class ProvisionRefsMatcher(CitationMatcher):
             # find the first citation after the end of this match, but don't cross the end of a sentence or look too far.
             for c in citations:
                 if c.start > match.end():
+                    # Plain-text "of X" lookups should bind to a cited document/work, not to a later provision-level
+                    # citation such as "section 57(1)", which would incorrectly become the target root.
+                    try:
+                        if c.href and FrbrUri.parse(c.href).portion:
+                            continue
+                    except ValueError:
+                        continue
                     # don't cross the end of a sentence or look too far ahead
                     if not ('. ' in match.string[match.end():c.start] or c.start - match.end() > self.max_ref_to_target):
                         frbr_uri = c.href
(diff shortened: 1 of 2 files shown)


indigo/analysis/refs/provisions.py
```diff
@@ -787,9 +794,11 @@ def resolve_target_uri(self, node: Element, frbr_uri: str) -> Tuple[Union[str, N
                             self.target_root_cache[frbr_uri] = (None, elements[0])
             else:
                 try:
-                    # we normalise the FRBR URI to the work level
+                    # Normalise to the base expression/work URI before using it as the root for nested references.
                     uri = FrbrUri.parse(frbr_uri)
-                    self.target_root_cache[frbr_uri] = (frbr_uri, self.find_document_root(uri))
+                    uri.portion = None
+                    base_uri = uri.expression_uri() if uri.language else uri.work_uri(False)
```
longhotsummer: Check for an expression date rather than a language. For historical reasons sometimes we get a default language when parsing a uri even if one wasn't given. A decision I regret!

Pls also add a test with just a work uri. The bulk of citations (if not all in the current setup) will be to a work and not an expression.

longhotsummer (changes requested):
Nice!

indigo/analysis/refs/provisions.py
actlikewill: interesting, done

longhotsummer (approved):

